PR #1, opened 2020-03-10
Make <=> work as expected with strings

watzon:
Currently the spaceship operator with strings just returns the result of running `strcmp`. In reality the spaceship operator is supposed to return -1 if the left side is less than the right side, 1 if the left side is greater than the right side, and 0 if the two sides are equal.

### DIFF
--- a/lib/natalie/repl.rb
+++ b/lib/natalie/repl.rb
@@ -14,32 +14,52 @@ end
 
 module Natalie
   class Repl
+    def initialize
+      @to_clean_up = []
+      @var_decls = {}
+      # Readline.completion_proc = self.method(:completion)
+    end
+
     def go
-      to_clean_up = []
       env = nil
-      vars = {}
       loop do
-        break unless (line = get_line)
-        ast = Natalie::Parser.new(line, '(repl)').ast
-        last_node = ast.pop
-        ast << last_node.new(:call, nil, 'puts', s(:call, s(:lasgn, :_, last_node), 'inspect'))
-        out = Tempfile.create('natalie.so')
-        to_clean_up << out
-        compiler = Compiler.new(ast, '(repl)')
-        compiler.repl = true
-        compiler.vars = vars
-        compiler.out_path = out.path
-        compiler.compile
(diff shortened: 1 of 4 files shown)


test/natalie/string_test.nat
```diff
@@ -3,6 +3,20 @@
 require_relative '../spec_helper'
 
 describe 'string' do
+  describe "#<=>" do
+    it "should return -1 if lhs is less than rhs" do
+      ('a' <=> 'b').should == -1
```
seven1m: If you make this `('a' <=> 'z').should == -1`, then the test fails on master.

test/natalie/string_test.nat
```diff
@@ -3,6 +3,20 @@
...
+    end
+
+    it "should return 1 if lhs is greater than rhs" do
+      ('b' <=> 'a').should == 1
```
seven1m: Same here: `('z' <=> 'a')`

watzon:
Changes made :+1: should fail on master now I'd expect

lib/natalie/repl.rb
```diff
@@ -14,33 +14,53 @@ def self.readline(prompt, _)
...
+      compiler.out_path = outfile.path
+      compiler.compile
+      
+      @var_decls = compiler.context[:vars]
+    end
+
+    def completion(str)
+      # TODO
+    end
+
```
seven1m: I think I got a sneak peak of something else you're working on. 😉

watzon: Oops didn't realize the repl stuff got pushed :joy:
Yeah I at least made it not panic on `Ctrl+C` now. I definitely need a little more time with the compiler before I figure out how to make completion work.

seven1m:
Thanks for fixing this!

lib/natalie/repl.rb
seven1m: I pulled out the REPL changes and merged this via the terminal. If you think the REPL stuff is good to go, feel free to make a new PR for that! Can't wait!

PR #3, opened 2020-07-16
Add missing methods to Float

watzon:
Just wanted to get involved here again, decided since we have Float now I'd go add some missing stuff. I'll just keep this open until we have covered the majority of the Float API.

- [x] `#magnitude` (alias for `#abs`)
- [x] `#modulo` (alias for `#%`)
- [x] `#negative?`
- [x] `#positive?`
- [x] `#next_float`
- [ ] `#prev_float`
- [ ] `#round`
- [ ] `#truncate`


seven1m:
@watzon Welcome back! This is is excellent, and I'm really glad to have your help here.

I have done some work recently to automatically generate bindings between Ruby and C++. The binding generator is here: https://github.com/seven1m/natalie/blob/master/lib/natalie/compiler/binding_gen.rb

The individual bindings are specified here: https://github.com/seven1m/natalie/blob/9573566ad24d101c473bcbf4efc449836fb753f6/lib/natalie/compiler/binding_gen.rb#L47-L52

Here is an example commit where I am moving stuff out of `float.cpp` and into `float_value.cpp`: https://github.com/seven1m/natalie/commit/a7c359c30c0b886709b17ed7d60163053f75f677 (The `bindings.cpp` file used to be called `bridge.cpp`, until I renamed it later.)

My end goal is to git rid of the `float.cpp` file and the other ones like it, such as `integer.cpp`, etc.

I hope that makes sense!

watzon:
Ooooh I like the idea of that, let me check it out!

seven1m:
@watzon I've got an incoming commit that will change the bindings to use pointers. I had tried to make use of C++ references, but it's proving to be a bit harder of a transition for me all at once. I'll stick with Value* (rather than moving to Value&). Just didn't want you to have to rewrite your stuff more than once, sorry!

I'll update you here when that commit lands!

seven1m:
@watzon I pushed ba159a776c051efc189b7e7b983617846a3a02e6 which moves the remaining Float methods over to `float_value.cpp` with auto-generated bindings. This is the approach I'd like to take for all bindings going forward, and I think this is stable enough to build on.

Sorry for the churn! Hopefully I didn't scare you away. Looking forward to your help on this!

watzon:
I'm starting to think that for methods like [`Float#round`](https://ruby-doc.org/core-2.5.3/Float.html#method-i-round) we may need a `min_argc` and `max_argc`. to cover optional parameters. Also is there a way currently to require that parameters after a certain point be named?

seven1m:
I haven’t done much work with named params yet. Is there a Float method that has a named arg?

watzon:
Yeah `#round` has a named argument called `mode`. I'm not bothering to implement it at the moment anyway because the logic is a little outside my realm of expertise, but it's definitely a thing to think about.

seven1m:
@watzon I merged this in 08d78fd973fb401d32f7b4aad8c5b6a0f9608fa5. I don't want the PR to get stale, and you can open another PR if you make any more progress! 😄 

I thought the PR would show as merged, but I guess I did it wrong. Closing manually...

Issue #7, opened 2020-07-19
Build with CMake


seven1m:
I'd like to use CMake rather than our custom bespoke Makefile.

PR #14, opened 2020-08-12
Build: General CMake-y improvements

alimpfard:
Heya!
I'm not much of a CMake expert, but I figured I might as well help towards #7.
As it stands, my additions are based off the `Makefile` in the master branch, so the intermediate external library builds spew out a lot of logs, this is not immediately fixable without cmake hacks, so I left that out.

Here's an overview of what this PR changes:
- Move all (needed) built external libraries and includes to build dir
- Make root CMakeFiles less verbose ('fancy')
- 'Fix' `compiler_command` to be explicit about library dependencies
- Swap out LIB_PATHS to LIB_NAMES for the compiler invocation

Also, might I interest you in `cmake .. -GNinja` for faster and more reliably incremental builds?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,45 +2,30 @@ cmake_minimum_required(VERSION 3.10)
 
 project(Natalie)
 
+file(GLOB LIB_SOURCES "src/*.cpp")
+file(GLOB GENERATED_NAMES "src/*.rb")
+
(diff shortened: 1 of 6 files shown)


seven1m:
Woah!! This is awesome! ❤️ 

I'm at work right now, but I'll check this out later this evening!!!

PR #15, opened 2020-08-14
Kernel: Add 'loop'

### LINKED ISSUE #4: Implement Kernel#loop method
I had to use `while true` just now because I guess I forgot to implement `loop`. Whoops!

It appears that `loop` should be implemented as a method on the `Kernel` module:

```ruby
irb(main):001:0> method(:loop)
=> #<Method: Object(Kernel)#loop()>
irb(main):002:0> method(:loop).owner
=> Kernel
```

alimpfard:
This closes #4. "good first issue" indeed, made me look around the codebase :D 
Not sure whether the tests are sufficient, but that's all I could think of.

irrelevant idea: could this stuff be generated too? I know some stuff is implemented in ruby (`Array#permutations` for instance), so I'd think something as straightforward as `Kernel#loop` can too be generated?

### DIFF
--- a/include/natalie/kernel_module.hpp
+++ b/include/natalie/kernel_module.hpp
@@ -6,6 +6,8 @@
 
 #include "natalie/forward.hpp"
 #include "natalie/value.hpp"
+#include "natalie/integer_value.hpp"
+#include "natalie/nil_value.hpp"
 
 namespace Natalie {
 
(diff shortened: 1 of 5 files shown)


include/natalie/kernel_module.hpp
```diff
@@ -6,6 +6,8 @@
 
 #include "natalie/forward.hpp"
 #include "natalie/value.hpp"
+#include "natalie/integer_value.hpp"
+#include "natalie/nil_value.hpp"
```
alimpfard, line 10: These two were missing (probably included in the other headers/source files), but I added them here anyway

seven1m:
This is great! Thanks!

As for your idea: Yes, I suppose we could pick some low level control flow expressions, implement those in C++, then implement some higher level ones in Ruby. I had considered it. But I think it might make the code harder to understand and debug, esp. for newcomers, if it's unclear _where_ something is supposed to be implemented. That being said, it might be worth revisiting down the road!

PR #27, opened 2020-10-03
Array: #sample

MiguelBel:
I saw a tweet about this project and I thought it was cool.

I'm not versed in cpp but I wanted to add the sample method for arrays.

My concerns are regarding how accurate is the randomness, I guess is good enough (I wouldn't use it for an online casino though). I did a script to check the distribution:

```
ary = [1,2,3,4,5,6,7,8,9,10]
ocs = []

99999.times { ocs << ary.sample }

ary.each { |e| puts e; puts ocs.select { |s| s == e }.length }
```

Output:

```$ bin/natalie /tmp/distribution.natalie
1
10037
2
9904
3
9934
4
9949
5
9979
6
10034
7
10079
8
10104
9
9975
10
10004
```

Looking for feedback.

### DIFF
--- a/include/natalie/array_value.hpp
+++ b/include/natalie/array_value.hpp
@@ -106,6 +106,7 @@ struct ArrayValue : Value {
     Value *join(Env *, Value *);
     Value *cmp(Env *, Value *);
     Value *select(Env *, Block *);
+    Value *sample(Env *);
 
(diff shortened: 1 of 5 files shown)


seven1m:
This is great! Thank you!!!

PR #33, opened 2020-12-22
Build on OpenBSD

jcs:
Requires `gmake` to be installed, along with the `automake` and `autoconf` versions specified in the environment.

`-ldl` is not needed on OpenBSD, and I think is only used on Linux.  `RUBY_PLATFORM.match(/linux/)` would probably be nicer than shelling out to `uname` but I don't think that exists in Natalie yet. 

```
$ env AUTOCONF_VERSION=2.69 AUTOMAKE_VERSION=1.16 gmake
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -S . -B build -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_MAKE_PROGRAM="gmake"
-- The C compiler identification is Clang 10.0.1
-- The CXX compiler identification is Clang 10.0.1
[...]
[ 96%] Building CXX object CMakeFiles/natalie.dir/generated/struct.cpp.o
[ 98%] Linking CXX static library libnatalie.a
[100%] Built target natalie
cp build/compile_commands.json .
$ uname -srp
OpenBSD 6.8 amd64
$ bin/natalie examples/hello.rb
hello world
$
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,12 @@ cmake_minimum_required(VERSION 3.10)
 project(Natalie)
 
 set(default_build_type "Debug")
-set(CMAKE_CXX_FLAGS_DEBUG "-pthread -ldl -g -Wall -Wextra -Werror -Wno-unused-parameter -Wno-unused-variable -Wno-unused-but-set-variable -Wno-unknown-warning-option")
-set(CMAKE_CXX_FLAGS_RELEASE "-pthread -ldl -O2")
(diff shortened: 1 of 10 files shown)


seven1m:
Thanks Joshua! This is fantastic.

seven1m:
Sourcehut supports OpenBSD build containers, so I'm going to work on getting Natalie to build over at https://git.sr.ht/~tim/natalie...

PR #34, opened 2020-12-24
More OpenBSD fixes

jcs:
The bdwgc patch is kind of broad, but it fixes the crashing binary during tests.

The other two get all non-skipped tests passing without `dmesg` chatter, except for the first REPL test which seems to be a completely separate issue.


seven1m:
Wow thanks!! ❤️

PR #35, opened 2020-12-25
Repl: If we don't have a TTY, don't use Readline

jcs:
The failure of this test on OpenBSD was due to some differences in Readline/TTY handling which caused Readline to not echo the input back to the output when not on a TTY.  I spent some time trying to figure out why and how to force the echoing by using a PTY in `ReplWrapper`, but I couldn't get it to work properly.

In the end, I'm not sure it's that useful to test Readline's behavior anyway, so maybe the REPL should just print its output when STDOUT isn't a TTY, and the test can just check for the expected output.

🎄 🎁 🎅 

```
Finished in 496.087667s, 0.4334 runs/s, 0.5826 assertions/s.

215 runs, 289 assertions, 0 failures, 0 errors, 54 skips
```

### DIFF
--- a/lib/natalie/repl.rb
+++ b/lib/natalie/repl.rb
@@ -3,12 +3,18 @@ require 'fileutils'
 require 'tempfile'
 
 begin
-  require 'readline'
+  if STDOUT.tty?
+    require 'readline'
+  else
+    raise LoadError
+  end
 rescue LoadError
   class Readline
     def self.readline(prompt, _)
-      print prompt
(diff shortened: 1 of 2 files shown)


seven1m:
OMG I didn't see this PR until just now. I actually tackled this in f6a94701f8ef60f2c2ad7871cf45136bbb4429ce, but took a slightly different approach. I think I like your work better, so I'll be merging this shortly. Thanks!

PR #37, opened 2021-06-24
Deduplicate looping logic in ranges

davidot:
Was inspired by your video to check if I could remove the duplicated code.
Could not really find much about project style so hope this is correct.

Let me know if you do not want `if constexpr` since it is not used anywhere else in the project and might thus require a higher version of c++.

(Also I have barely used ruby so hope switching the ordering slightly does not break the spec or some other assumptions)

### DIFF
--- a/include/natalie/range_value.hpp
+++ b/include/natalie/range_value.hpp
@@ -51,6 +51,9 @@ private:
     ValuePtr m_begin { nullptr };
     ValuePtr m_end { nullptr };
     bool m_exclude_end { false };
+
+    template <typename Function>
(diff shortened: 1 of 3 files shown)


seven1m:
Thanks so much; this is sick! ...And definitely beyond my current C++ coding abilities. 😆 

I'm fascinated to learn about that `std::is_void_v` trick for handling the lack of return value. Using a template to accept the lambda is cool too. I'd only used the `<functional>` header before, but I like this approach better I think.

Cool stuff, thanks for taking the time to code it up!

PR #38, opened 2021-06-25
Implement Array#last and #first with argument n

davidot:
Saw a couple of TODO's so tried implementing this based on https://ruby-doc.org/core-3.0.1/Array.html#method-i-last 
and observed behavior from ruby.
Again I am at all familiar with ruby so I don't know if I missed some edge cases in tests.
But all that is implemented and tested now behaves like ruby.

I tried to copy the style as much as possible but do have a couple of questions which I will try to post below.

(And somehow I always end up working on arrays :thinking:)


lib/natalie/compiler/binding_gen.rb
```diff
@@ -554,7 +554,7 @@ def generate_name
 
 gen.binding('Parser::Sexp', 'new', 'SexpValue', 'new_method', argc: :any, pass_env: true, pass_block: false, return_type: :Value)
 gen.binding('Parser::Sexp', 'inspect', 'SexpValue', 'inspect', argc: 0, pass_env: true, pass_block: false, return_type: :Value)
-gen.binding('Parser::Sexp', 'sexp_type', 'SexpValue', 'first', argc: 0, pass_env: true, pass_block: false, return_type: :Value)
+gen.binding('Parser::Sexp', 'sexp_type', 'SexpValue', 'first', argc: 0..1, pass_env: true, pass_block: false, return_type: :Value)
```
davidot, line 557: Should this binding be changed? I could not get it to compile without this or adding some specialization to sexpvalue.
But I could not find any spec or tests to run so not sure if this broke anything

src/array_value.cpp
```diff
@@ -228,31 +228,69 @@ ValuePtr ArrayValue::map(Env *env, Block *block) {
...
-ValuePtr ArrayValue::first(Env *env) {
-    if (size() > 0) {
+ValuePtr ArrayValue::first(Env *env, ValuePtr n) {
+    auto has_count = n != nullptr;
+
+    if (!has_count) {
+        if (is_empty())
+            return NilValue::the();
+
         return (*this)[0];
```
davidot, line 238: I copied the usage of `(*this)[0]` which only seems to add an assert but since we have checked it is non empty is that necessary? 
Same for below we (should) know the index is correct and can thus go directly to `m_vector`.

src/array_value.cpp
```diff
@@ -228,31 +228,69 @@ ValuePtr ArrayValue::map(Env *env, Block *block) {
...
-    } else {
-        return NilValue::the();
     }
+
+    ArrayValue* array = new ArrayValue();
+
+    n->assert_type(env, Value::Type::Integer, "Integer");
+    nat_int_t n_value = n->as_integer()->to_nat_int_t();
```
davidot, line 244: When we have an argument which needs to be an int something is this the way to convert / check it?
I copied this code from `Array#shift` but was not sure about if this is correct enough.

test/natalie/array_test.rb
```diff
@@ -495,6 +495,62 @@
...
+     [1, 2, 3].first(0).should == []
+     [].first(0).should == []
+   end
+
+    it 'returns empty array if the array is empty' do
+      [].first(2).should == []
+    end
+
+    it 'Should raise ArgumentError on negative n' do
+       r = begin
```
davidot, line 516: Is this how we should test for error and check their type?
Could not find anything else that did this.

davidot (commented):
Some questions I had

src/array_value.cpp
```diff
@@ -228,13 +228,60 @@ ValuePtr ArrayValue::map(Env *env, Block *block) {
...
+
+    if (n_value < 0) {
+        env->raise("ArgumentError", "negative array size");
+        return nullptr;
```
davidot, line 248: Should we return anything here?
I saw raise is `[[noreturn]]` but to me it look weird to have the if not "stop" the code but maybe I'm just not familar enough with the codebase.

davidot:
I also have a couple more general questions about the project but not sure where I can ask those.
Most importantly should this have been implemented in c++ at all?
Or should it have been implemented in ruby in `array.rb`?
What should be done in which language? Based on speed or ease of implementation? Or just what you feel like at that time?

lib/natalie/compiler/binding_gen.rb
seven1m: This is correct. Since it's actually calling through to the `ArrayValue::first` underneath, the argc needs to match. :smile:

test/natalie/array_test.rb
```diff
@@ -495,6 +495,62 @@
     end
   end
 
+  describe '#first(n)' do
+    it 'should return a new array of the last n elements' do
```
seven1m, line 499:
```suggestion
    it 'should return a new array of the first n elements' do
```

test/natalie/array_test.rb
seven1m: This is fine. I'm not super picky. There is a shortcut for this though:

```ruby
-> { [].first(-1) }.should raise_error(ArgumentError, 'negative array size')
```

src/array_value.cpp
seven1m: Yeah it looks weird to not have a return I guess, but it's a common pattern throughout the codebase to just `raise` and have no return, so I'm used to it. No return needed.

src/array_value.cpp
seven1m: This is correct. This method only accepts an int (or nil), so if the user passes, say, a string, then they'll get the error:

```
TypeError (no implicit conversion of String into Integer)
```

If we have need to convert from String to Integer, we'd need to `send("to_i")`, but that's not what Ruby does here. So we're good! :smile:

src/array_value.cpp
seven1m: You're correct, it's unneeded, and I've probably been inconsistent in using `[]` vs going directly to `m_vector`. Your call.

seven1m (approved):
This is fantastic! Thank you so much for seeing a need and jumping in there to fix it. Your tests are great, so I feel good about merging this as-is.

I did try to answer your questions best I could; if you want to make any changes, I'll wait a bit before merging.

<3

seven1m:
>Most importantly should this have been implemented in c++ at all?
>Or should it have been implemented in ruby in array.rb?
>What should be done in which language? Based on speed or ease of implementation? Or just what you feel like at that time?

Uh, yeah, that's a hard one. I've gone back and forth on that, and I don't have a good system for deciding which code should go where.

Naturally, there are some optimizations we can do in the C++ that we cannot do in the Ruby, so I think for methods that are used a lot like `first` and `last`, having the code in C++ "feels right" to me.

On the other hand, I implemented `Array#permutation` in [array.rb](https://github.com/seven1m/natalie/blob/master/src/array.rb) (as you probably noticed) because it's not typically a method that a Ruby program would call often (or in a tight loop), and it was just soooo much funner to implement in Ruby itself. :wink:

Thus I think we just weight the tradeoffs of maintainability with performance, and there is no perfect answer. :man_shrugging:

davidot:
Just changed the error test as I think they make the intent of the test more clear.
Lets leave the rest for optimization or cleaning later.

test/natalie/array_test.rb
```diff
@@ -541,13 +535,7 @@
...
 
     it 'Should raise ArgumentError on negative n' do
-       r = begin
-             [].last(-1)
-             raise TestFailed 'Should raise'
-           rescue ArgumentError => e
-             e
-           end
-       r.message.should == 'negative array size'
+       { [].first(-1) }.should raise_error(ArgumentError, 'negative array size')
```
seven1m, line 538:
```suggestion
       { [].last(-1) }.should raise_error(ArgumentError, 'negative array size')
```

davidot: Ah too quick on the copy and paste

PR #40, opened 2021-06-26
Add basic array methods

davidot:
Most of these were easy pickings
But I did run into a couple of interesting cases so I have another couple of questions


include/natalie/array_value.hpp
```diff
@@ -110,7 +113,8 @@ class ArrayValue : public Value {
     ValuePtr max(Env *);
     ValuePtr min(Env *);
     ValuePtr push(Env *, size_t, ValuePtr *);
-    ValuePtr ref(Env *, ValuePtr, ValuePtr);
+    ValuePtr rassoc(Env *, ValuePtr);
+    ValuePtr ref(Env *, ValuePtr, ValuePtr = nullptr);
```
davidot, line 117: Giving the second argument a default value make it easy to use as a binding
But is that what we want? Or would a method `at` which just calls `ref` be better?

src/array_value.cpp
```diff
@@ -257,6 +257,25 @@ ValuePtr ArrayValue::first(Env *env, ValuePtr n) {
     return array;
 }
 
+
+ValuePtr ArrayValue::drop(Env *env, ValuePtr n) {
```
davidot, line 261: A lot of duplicated logic but let first implement everything before we refactor and optimize

src/array_value.cpp
```diff
@@ -507,4 +526,43 @@ ValuePtr ArrayValue::uniq(Env *env) {
     return hash->keys(env);
 }
 
+ValuePtr ArrayValue::clear(Env *env) {
+    this->assert_not_frozen(env);
```
davidot, line 530: Noticed `assert_not_frozen` in pop when should this be called?
Anytime when changing the underlying value?

test/natalie/array_test.rb
```diff
@@ -475,6 +475,15 @@
...
 
+  describe '#collect' do
+      it 'returns a new array of the result of evaluating the block for each item in the array' do
+        result = [1, 2, 3].collect do |i|
+          i * 2
+        end
+        result.should == [2, 4, 6]
+      end
+    end
+
```
davidot, line 486: This is just a copy and paste of the `#map` test is there some easier way to test aliases?
Since now you would have to double test writing as well (or you might just forget something has an alias

test/natalie/array_test.rb
```diff
@@ -658,6 +700,96 @@
     end
   end
 
+  describe '#clear' do
+    it 'should leave an empty array empty' do
+      a = []
+      (a.clear).should === a
```
davidot, line 706: Is this how you test the value of an expression?
And do we want to actually test / specify values of expressions?

include/natalie/array_value.hpp
seven1m: I think I prefer the latter (method `at` which calls `ref`), just for clarity of purpose. :smile:

src/array_value.cpp
seven1m: Yes, any time we are mutating the underlying object rather than returning a copy, we should add `assert_not_frozen`.

test/natalie/array_test.rb
seven1m: I supposed you could do something like this:

```ruby
['map', 'collect'].each do |method_name|
  describe "##{method_name}" do
    it 'whatever' do
      result = [1, 2, 3].send(method_name) do |i|
        i * 2
      end
      result.should == [2, 4, 6]
    end
  end
end
```

BUT I'm not sure that's better. I think I'd rather just have the duplication. :man_shrugging:

test/natalie/array_test.rb
seven1m: I don't think this would test what you want. The left side would be evaluated before the right side is passed into `should ===` Also, `===` in Ruby is kinda special for use by `case/when`, so probably not appropriate here.

I think you just want:

```
a = []
a.clear.should == []
a.should == []
```

seven1m: [Note: I deleted an earlier comment here that wasn't right.]

This is close, but `===` isn't what you want here. You should use `a.clear.should equal(a)` since that tests for object identity (same pointer).

If `===` is doing that currently, then it's a bug :wink:

include/natalie/array_value.hpp
davidot: Made it an explicit method

test/natalie/array_test.rb
davidot: Alright, kept it like this

test/natalie/array_test.rb
davidot: Changed it to `equal(a)` but this did work in both ruby and natalie so not sure what that means?

davidot:
I think I addressed all the comments.
Maybe another question, should array_test.rb be split up at some point? It is already quite large but then testing does become more difficult of course

test/natalie/array_test.rb
seven1m: Ahh, you're right. I'm so used to seeing `===` used for case/when that it looked wrong to me. Carry on :smile:

seven1m:
Yeah, if you think we should, we can split out tests for each method into their own files inside an `array` directory:

```
array/
  at_test.rb
  assoc_test.rb
  clear_test.rb
  ...
```

I'm super happy with this work, so I'm going to go ahead and merge. <3

PR #41, opened 2021-06-27
Array #intersection, #reverse and friends

davidot:
Saw a todo for #reverse and was already working on #& and intersection.
(I did not actually use reverse at the TODO because I'm not entirely sure how it should be used)


src/array_value.cpp
```diff
@@ -569,5 +571,53 @@ ValuePtr ArrayValue::rassoc(Env *env, ValuePtr needle) {
...
+    ArrayValue* copy = new ArrayValue(env, *this);
+    copy->reverse_in_place(env);
+    return copy;
+}
+
+ValuePtr ArrayValue::reverse_in_place(Env *env) {
+    assert_not_frozen(env);
+
+    for (size_t i = 0; i < size() / 2; i++) {
+        std::swap(m_vector[i], m_vector[size() - 1 - i]);
```
seven1m, line 618: This is cool! Is this safe for an array with an odd number of elements? :thinking: 

```
a = [1, 2, 3]
size() / 2 = 1
swap(0, 2)
```

Yeah, I guess that works! :sweat_smile:

seven1m:
Really cool, thanks!! <3

PR #43, opened 2021-06-30
More array methods

davidot:
Since `#any?` was implemented via enumerable I tried to do the same
But I cannot get one of the spec tests to pass
Also this is probably the most ruby I have written in a long time (maybe ever) so feel free to rip it to shreds


src/array_value.cpp
```diff
@@ -620,4 +620,61 @@ ValuePtr ArrayValue::reverse_in_place(Env *env) {
...
+    auto Enumerable = GlobalEnv::the()->Object()->const_fetch(SymbolValue::intern("Enumerable"))->as_module();
+    auto any_method = Enumerable->find_method(env, SymbolValue::intern("none?"));
+    return any_method->call(env, this, argc, args, block);
+}
+
+ValuePtr ArrayValue::one(Env *env, size_t argc, ValuePtr *args, Block *block) {
+    auto Enumerable = GlobalEnv::the()->Object()->const_fetch(SymbolValue::intern("Enumerable"))->as_module();
+    auto any_method = Enumerable->find_method(env, SymbolValue::intern("one?"));
+    return any_method->call(env, this, argc, args, block);
+}
```
seven1m, line 677: Calling through to the version of `one?` and `none?` on Enumerable is totally fine for now, but we should add a comment here saying it's not exactly correct. Because that's not what Ruby does.

```ruby
class Thing
  include Enumerable

  def each
    yield 1
    yield 2
    yield 3
  end
end

t = Thing.new
a = [1, 2, 3]

p t.method(:one?).owner # => Enumerable
p a.method(:one?).owner # => Array


# works
p a.one? { |i| i == 2 } # => true

# works
p t.one? { |i| i == 2 } # => true

module Enumerable
  def one?
    false
  end
end

# works
p a.one? { |i| i == 2 } # => true

# broken
p t.one? { |i| i == 2 } # => false
```

seven1m: Again, don't worry about fixing this -- what you have here is still a huge improvement, so let's ship it!

davidot: Added the TODOs
(Still had a couple more changes that's why it was still a draft)
So let's ship it :boat:

seven1m:
Awesome!

PR #47, opened 2021-07-08
Fix and add Array methods

davidot:
I also started implementing Array#[] with sequences but that is ruby 3.
Which version do we want to emulate?

### DIFF
--- a/test/natalie/array_test.rb
+++ b/test/natalie/array_test.rb
@@ -1550,4 +1550,77 @@ describe 'array' do
     end
   end
 
+  describe '#slice' do
+    it 'returns the item at the given index' do
+      a = [1, 2, 3, 4, 5, 6]
+      a.slice(0).should == 1
+      a.slice(1).should == 2
+    end
+
+    it 'returns the item offset from the end when given a negative index' do
+      a = [1, 2, 3, 4, 5, 6]
+      a.slice(-1).should == 6
+      a.slice(-2).should == 5
+    end
+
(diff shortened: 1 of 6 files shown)


test/natalie/array_test.rb
```diff
@@ -454,19 +454,244 @@
       a = []
```
davidot, line 454: This file is getting very large.
What would be a good way to split it up?

seven1m: Just start putting new stuff in a separate file as you go. `test/natalie/array/whatever_test.rb`

seven1m: Alternatively, if you want, you can pull spec files over from https://github.com/ruby/spec like I did with `spec/core/enumerable/*`. But I know writing the tests yourself is a bit more fun in some cases, so whichever you prefer!

seven1m:
I didn't even know about the Ruby 3 taking an Enumerator (docs say "sequence"?) as an argument to `Array#[]`. That's cool!

We can implement that, for sure. Now that Enumerators are kinda working. :wink:

test/natalie/array_test.rb
davidot: I probably will at least also pull in their tests since we are missing frozen tests and other edge cases
Is there some way to run all the tests in a directory?

seven1m: > Is there some way to run all the tests in a directory?

No. :disappointed: I think our spec runner would need to be taught how to run more than one file.

PR #50, opened 2021-07-20
Implement Enumerable#tally

richardboehme:
This PR implements Enumerable#tally. To make the specs work I also needed to implement Object#itself and Hash#transform_values.

I'm kind of unsure about the implementation of Object#itself in CPP and about Hash#transform_values being in hash.rb instead of the HashValue class. Feel free to correct me if there are any mistakes 😅 

I also skipped Hash#transform_values! for now because it was not relevant to make the tally spec working.


seven1m:
This is great, thanks!!! <3

Issue #19, opened 2020-08-28
Implement the Enumerable module


seven1m:
I started work in 2508f404717945e5fc8f6eac2649b3126bfd8528 on the Enumerable module. It's a biggie! Anyone is welcome to jump in and lend a hand!

Here are the specs that need to be enabled and passing:

- [x] all_spec.rb
- [x] any_spec.rb
- [x] chain_spec.rb
- [x] chunk_spec.rb
- [x] chunk_while_spec.rb
- [x] collect_concat_spec.rb
- [x] collect_spec.rb
- [x] count_spec.rb
- [x] cycle_spec.rb
- [x] detect_spec.rb
- [x] drop_spec.rb
- [x] drop_while_spec.rb
- [x] each_cons_spec.rb
- [x] each_entry_spec.rb
- [x] each_slice_spec.rb
- [x] each_with_index_spec.rb
- [x] each_with_object_spec.rb
- [x] entries_spec.rb
- [x] filter_map_spec.rb
- [x] filter_spec.rb
- [x] find_all_spec.rb
- [x] find_index_spec.rb
- [x] find_spec.rb
- [x] first_spec.rb
- [x] flat_map_spec.rb
- [x] grep_spec.rb
- [x] grep_v_spec.rb
- [x] group_by_spec.rb
- [x] include_spec.rb
- [x] inject_spec.rb
- [x] lazy_spec.rb
- [x] map_spec.rb
- [x] max_by_spec.rb
- [x] max_spec.rb
- [x] member_spec.rb
- [x] min_by_spec.rb
- [x] minmax_by_spec.rb
- [x] minmax_spec.rb
- [x] min_spec.rb
- [x] none_spec.rb
- [x] one_spec.rb
- [x] partition_spec.rb
- [x] reduce_spec.rb
- [x] reject_spec.rb
- [x] reverse_each_spec.rb
- [x] select_spec.rb
- [x] slice_after_spec.rb
- [x] slice_before_spec.rb
- [x] slice_when_spec.rb
- [x] sort_by_spec.rb
- [x] sort_spec.rb
- [x] sum_spec.rb
- [x] take_spec.rb
- [x] take_while_spec.rb
- [x] tally_spec.rb
- [x] to_a_spec.rb
- [x] to_h_spec.rb
- [x] uniq_spec.rb
- [x] zip_spec.rb

seven1m:
[...]

PR #51, opened 2021-07-22
Initial stab at trying to implement min_by

philberty:
This function ideally should call the already defined min function to get
the existing behaviour with the 'n' parameter but with my non existant ruby
knowledge I am unable to call this function.

Addresses #19

I am getting:

```
Failed specs:                                                                                                              
Enumerable#min_by                                                                                                          
  returns the object for whom the value returned by block is the smallest                  
    wrong number of arguments (given 1, expected 0) (ArgumentError)          
      /home/philbert/workspace/natalie/src/enumerable.rb:519:in `min'             
      /home/philbert/workspace/natalie/src/enumerable.rb:519:in `min_by'               
      ../spec/core/enumerable/min_by_spec.rb:15:in `block in block in block'
      /home/philbert/workspace/natalie/test/support/spec.rb:663:in `call'
      /home/philbert/workspace/natalie/test/support/spec.rb:663:in `block in run_specs'
      /home/philbert/workspace/natalie/test/support/spec.rb:644:in `each'
      /home/philbert/workspace/natalie/test/support/spec.rb:644:in `run_specs'
      /home/philbert/workspace/natalie/test/support/spec.rb:734:in `block in block'
      ../spec/core/enumerable/min_by_spec.rb:5:in `<main>'

12 spec(s) passed.
0 spec(s) failed.                                          
8 spec(s) errored.                                                                                                         
2 spec(s) skipped.
```

### DIFF
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -505,6 +505,19 @@ module Enumerable
     end
   end
 
+  def min_by(n = nil)
+    return enum_for(:min_by) unless block_given?
+
+    ary = []
+
+    gather = ->(obj) { obj.size <= 1 ? obj.first : obj }
+    each do |*item|
+      ary << gather.(item)
+    end
+
+    ary.size == 0 ? nil : ary.min(n)
+  end
+
   def reject
     return enum_for(:reject) unless block_given?
 


seven1m:
Unfortunately, the Array#min method doesn't accept an `n` parameter yet. `Enumerable#min` does, but Array has its own min function and we haven't done much work on it yet.

For now, you can duplicate the code from `Enumerable#min` and then we can see if there is a way the logic could be shared.

philberty:
Thanks for the guidance I think it's working now Let me close this PR and reopen a new one.

PR #52, opened 2021-07-22
Fix "spec(s) passed" count if some specs errored

richardboehme:
Previously when running a test that throws an error, the summary message
of the spec reporter would report one passed and one errored spec. This
fixes this behavior by substracting the number of errored specs from the
passed spec count.


seven1m:
Thanks! That's been bugging me haha.

PR #56, opened 2021-07-24
Implement Enumerable#sort

richardboehme:
This was a tough one 😅 

Basically Enumerable#sort ideally should just call `to_a.sort`. However, Array#sort did not support passing a block to it. So I tried to change that.

Please take this PR as a proposal for an implementation. My C++ is not that great (basic university level), so I expect that my code may break some conventions or even has some critical bugs. 

To make everything clear I want to outline my thoughts on this:
* Vector#sort takes a SortComparator that has a `cmp` function. This function should evaluate the block in each iteration.
* The previous definition of the compare function in `array_value.cpp` creates a lambda. I tried modifying the lambda to evaluate the block, but this does not work, because we have to capture the block which is not allowed when passing the lambda as a function pointer.
* Easiest way around this would be to expand SortComparator to take a block, but since the vector class seems to be independent of Natalie I did not want to bring the Ruby context to it.
* My idea: Change SortComparator to have a function `compare` which calls the `cmp` member. Now it's possible to inherit from SortComparator and override the `compare` function. I created ArraySortComparator which has a block member and overrides the `compare` function to evaluate the block if needed.
* To make this work I had to change Vector#sort to take a pointer to a SortComparator object. Otherwise it would not be possible to override the `compare` function.

Everything seems to work fine, but as I said I'm not sure if the implementation is ideal.

The array spec is mostly green. I had to skip some tests, because of unimplemented features in Natalie (recursive arrays, Comparable module, ...). Two specs operate on a big array which makes the spec really slow.


seven1m:
I love your approach! Great work <3

seven1m:
btw I removed the `skip-spec` line from `spec/core/enumerable/sort_spec.rb` in fc90ee3a2891f88639e9bd6331d65da0b9f0c547. All the tests are passing; I assume you just forgot to do it in this PR.

seven1m:
One more thing... I checked in IRB and `Object#<=>` actually lives in the Kernel module:

```ruby
irb(main):001:0> Object.new.method(:<=>)
=> #<Method: Object(Kernel)#<=>(_)>
irb(main):002:0> Object.new.method(:<=>).owner
=> Kernel
```

So I added a fixme note for that in 265f14e8f24e2e114dee6816bed33194c8197e80. (No need to fix it now; I just wanted to show you how to easily check where a method should be added!)

richardboehme:
> btw I removed the `skip-test` line from `spec/core/enumerable/sort_spec.rb` in [fc90ee3](https://github.com/seven1m/natalie/commit/fc90ee3a2891f88639e9bd6331d65da0b9f0c547). All the tests are passing; I assume you just forgot to do it in this PR.

Thanks for catching that!



> One more thing... I checked in IRB and `Object#<=>` actually lives in the Kernel module:
> 
> ```ruby
> irb(main):001:0> Object.new.method(:<=>)
> => #<Method: Object(Kernel)#<=>(_)>
> irb(main):002:0> Object.new.method(:<=>).owner
> => Kernel
> ```
> 
> So I added a fixme note for that in [265f14e](https://github.com/seven1m/natalie/commit/265f14e8f24e2e114dee6816bed33194c8197e80). (No need to fix it now; I just wanted to show you how to easily check where a method should be added!)

I see... It's the same for Kernel#itself which I put in the Object class as well. My reference was https://rubyapi.org/2.7/o/object. I might need to investigate why the methods are listed there and not on https://rubyapi.org/2.7/o/kernel.

seven1m:
> My reference was https://rubyapi.org/2.7/o/object. 

Hey, it's the same on ruby-doc.org! Maybe that's why I was in the habit of using irb... :man_shrugging: 

Anyway, no biggie, and we'll get it all right eventually... :wink:

richardboehme:
> Hey, it's the same on ruby-doc.org! Maybe that's why I was in the habit of using irb... 🤷‍♂️

Just took a quick look and this is so weird because it's even implemented in [object.c](https://github.com/ruby/ruby/blob/v3_0_1/object.c#L1632). But taking a closer look where the method is bound one can see it's bound to the Kernel module (see https://github.com/ruby/ruby/blob/v3_0_1/object.c#L4539).

PR #57, opened 2021-07-25
Implement Enumerable#sort_by and Array#sort_by!

richardboehme:
This also fixes a small bug with Stubs and introduces Stub#once and Stub#and_yield.


seven1m:
Sweet!

PR #58, opened 2021-07-27
Implement Enumerable#reverse_each

borisromanov:
This attempts to implement the `Enumerable#reverse_each` method.
[#19]

### DIFF
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -529,11 +529,18 @@ module Enumerable
   end
 
   def reverse_each(*args)
-    reversed_arr = to_a.reverse
-    return enum_for(:reverse_each, reversed_arr) unless block_given?
+    return enum_for(:reverse_each, *args) unless block_given?
 
-    reversed_arr.each do |item|
-      yield item
+    if args.size > 0
+      reversed = args.to_a.reverse
+      reversed.each do |item|
+        yield item
+      end
+    else
+      reversed = to_a.reverse
+      reversed.each do |item|
+        yield item
+      end
     end
   end
 
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -531,16 +531,15 @@ module Enumerable
   def reverse_each(*args)
     return enum_for(:reverse_each, *args) unless block_given?
 
-    if args.size > 0
-      reversed = args.to_a.reverse
-      reversed.each do |item|
-        yield item
-      end
-    else
-      reversed = to_a.reverse
-      reversed.each do |item|
-        yield item
-      end
+    ary = []
+    each(*args) do |*item|
+      ary << item
+    end
+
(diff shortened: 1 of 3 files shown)


src/enumerable.rb
```diff
@@ -528,6 +528,15 @@ def reject
     ary
   end
 
+  def reverse_each(*args)
+    reversed_arr = to_a.reverse
+    return enum_for(:reverse_each, reversed_arr) unless block_given?
```
seven1m, line 533: This isn't quite right. The `*args` should be passed in here. I'm not sure why the specs don't test this, but I wrote up a little Ruby program to prove it:

```ruby
class MyEnum
  include Enumerable

  def each(*args)
    args.each do |arg|
      yield arg
    end
  end
end

MyEnum.new.reverse_each(1, 2, 3) do |i|
  p i
end

e = MyEnum.new.reverse_each(1, 2, 3)
e.each do |i|
  p i
end
```

This should print:

```
3
2
1
3
2
1
```

seven1m (changes requested):
Almost there! The `*args` need to be passed along to the enumerator.

Also, can you word your commit messages in the imperative mood? i.e. "Implement Enumerable#reverse_each". You can put "[#19]" in the body of the commit message if you want.

borisromanov:
Hi Tim, 

I appreciate the review, as well as, you pointing me in the right direction. I made some updates to my implementation of `#reverse_each`. Due to the specs [provided at](https://github.com/ruby/spec/blob/master/core/enumerable/reverse_each_spec.rb) being deceptively rudimentary and not actually testing a decent chunk of the expected `#reverse_each` functionality. How would you like to handle this gap in coverage here in your codebase? Would you like me to modify/update the `reverse_each_spec.rb` file, or perhaps handle it a different way?

src/enumerable.rb
borisromanov: Got it, I appreciate you pointing me in the correct directions, the existing specs really led me to take a naive approach.

seven1m:
If you want, you could add the sample Ruby code I posted to the `reverse_each_spec` or to a new file in `test/natalie`. Your call.

borisromanov:
> If you want, you could add the sample Ruby code I posted to the `reverse_each_spec` or to a new file in `test/natalie`. Your call.

Sorry for the big delay, didn't get a chance to pick any of this up on the weekend. I went ahead and added a `reverse_each_test.rb` file to `test/natalie`. Let me know if those are acceptable tests. I took a quick stab a the description in the it block, let me know if you'd want something more to the point.

seven1m:
This was pretty close, but I made a tweak so that the `*args` are passed to `#each`. Pretty sure that's how it's supposed to work. The test I wrote for you just didn't prove it well enough. :smile:

PR #59, opened 2021-07-27
Issue #19: Patchup Enumerable#partition to be spec compliant

borisromanov:
#19

### DIFF
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -531,13 +531,15 @@ module Enumerable
   def partition
     return enum_for(:partition) unless block_given?
 
+    gather = ->(obj) { obj.size <= 1 ? obj.first : obj }
+
     left = []
     right = []
-    to_a.each do |item|
-      if yield(item)
-        left << item
+    each do |*item|
+      if yield(gather.(item))
+        left << gather.(item)
       else
-        right << item
+        right << gather.(item)
       end
     end
     [left, right]
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -529,9 +529,11 @@ module Enumerable
   end
 
   def partition
+    return enum_for(:partition) unless block_given?
+
     left = []
     right = []
-    each do |item|
+    to_a.each do |item|
       if yield(item)
         left << item
       else
--- a/spec/core/enumerable/partition_spec.rb
+++ b/spec/core/enumerable/partition_spec.rb
@@ -1,5 +1,3 @@
-# skip-test
-
 require_relative '../../spec_helper'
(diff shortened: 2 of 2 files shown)


src/enumerable.rb
```diff
@@ -529,9 +529,11 @@ def reject
   end
 
   def partition
+    return enum_for(:partition) unless block_given?
+
     left = []
     right = []
-    each do |item|
+    to_a.each do |item|
```
seven1m, line 536: Why `to_a` here? It makes something that can be computed lazily over large enumerators and makes it build everything into an array first. Did you do this to make it pass a spec?

I don't think `partition` should need to generate the whole array first in order to do its work.

seven1m (changes requested):
Thanks for jumping in! I'd love to check this one off the list, but I have one concern about using `to_a` here.

src/enumerable.rb
borisromanov: Thanks for raising this concern. If I recall I was just hacking against the specs and landed on this as something that passed the spec. Let me do some more digging and try to come up with a less naive approach.

borisromanov: After digging in a little more, natalie seems to be behaving differently when running a script that uses `#partition` and uniquely in the context of the spec.

Executing the following script with `./bin/natalie` on a clean build of the master branch, I get the expected result
```rb
class AnEnum
  include Enumerable

  p [[1,2], [3,4,5], [6,7,8,9]].partition {|e| e == [3, 4, 5] }
end
```

However, whenever I run the spec I get some weird behavior around the yield [here](https://github.com/seven1m/natalie/blob/3ac606e8088ae4e0e7f063d68a3068b19bdd4a0e/src/enumerable.rb#L535). I am perplexed and a little lost. Maybe I'll pick this up again in the morning.

seven1m: I don't think that Ruby code does what you think it does. That script is effectively just this:

```ruby
  p [[1,2], [3,4,5], [6,7,8,9]].partition {|e| e == [3, 4, 5] }
```

The class is irrelevant. Maybe that's what's confusing?

----

As for the `to_a` thing, I think you just needed to add a `gather.()` call in there like we did with the other methods. This will fix the issue with yielding multiple values.

borisromanov: You were correct. `to_a` was never needed and the lazy naive approach. I implemented the gather proc pattern and then noticed that the splat was missing when performing the do loop. A combination of those two things should resolve the issue I ran into last week, and get this spec passing.

seven1m:
This is good. Thank you for finishing it!

PR #61, opened 2021-07-30
Trying to fix the argument count when calling a method with a splat

### LINKED ISSUE #60: Issue with processing default arguments with an argument splat
After playing around a bit with implementing some more Enumerable methods I think I found a natalie bug with the handling of default arguments. The following code prints 2 in Ruby and 1 in Natalie:

```ruby
def foo(a = 1, *args)
  puts a
end
foo(2)
```

I tried to understand where to look for this issue but I did not really find the position where this issue occurs. Maybe a small hint could help me finding the issue here :^)

richardboehme:
This commit tries to fix #60, by calculating the `total_count` of arguments differently. 

Previously we only rejected keyword arguments which results in splats being counted into the total number of arguments.
This results in choosing the default value even though an argument is passed. 

See the following example:

```ruby
def foo(a = 2, *args)
  puts [a, args]
end
foo(1)
```

The [if-condition](https://github.com/seven1m/natalie/blob/fix-default-with-splat/src/natalie.cpp#L122) evaluates to `true` because `remain` is calculated by subtracting the number of arguments passed and the number of required arguments which is calculated by the `total_count` passed. 

Here `ary_len` would be 1, `total_count` would be 2, which results in `required_count` being 1.

`remain` would then evaluate to 0, which makes the if-condition evaluate to `true`, even though it should not. 

Rejecting the splat arguments from the `total_count` would result in it being 1, `required_count` would be 0 and `remain` would be 1, which seems to be the behavior that we want.

Also this let's us get rid of the subtraction when we assume a splat prior to the currently evaluated argument.

### DIFF
--- a/lib/natalie/compiler/method_args.rb
+++ b/lib/natalie/compiler/method_args.rb
@@ -63,8 +63,8 @@ module Natalie
                 @required_args += 1
                 default_value = 'nullptr'
               end
-              non_kwargs = names.reject { |name| name.sexp_type == :kwarg }
-              value = s(:arg_value_by_path, :env, value_name, default_value, :false, non_kwargs.size, defaults.size, defaults_on_right ? :true : :false, 0, path.size, *path)
+              total_argument_count = names.reject { |name| name.sexp_type == :kwarg || name.sexp_type == :splat }.size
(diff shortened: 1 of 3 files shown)


seven1m:
Wow, I'm impressed you were able to make sense of that garbage function! :laughing:

PR #62, opened 2021-07-31
Implement Enumerable#slice_when

richardboehme:
This also fixes `Enumerable#first`, sets the default argument for `Kernel#enum_for` and adds an alias to `Kernel#to_enum`.


seven1m:
Nice! :heart_eyes:

PR #63, opened 2021-08-03
Adds aliases for Enumerable#find_all and Enumerable#filter

borisromanov:
[#19]

Adds a couple of method aliases for `#find_all` and `#filter`

[ruby 2.7.3 Enumerable#filter](https://ruby-doc.org/core-2.7.3/Enumerable.html#method-i-filter)
[ruby 2.7.3 Enumerable#find_all](https://ruby-doc.org/core-2.7.3/Enumerable.html#method-i-find_all)


seven1m:
Well that was easy :wink:

PR #64, opened 2021-08-04
Implement Enumerable#filter_map

borisromanov:
[#19]

### DIFF
--- a/src/enumerable.rb
+++ b/src/enumerable.rb
@@ -314,6 +314,19 @@ module Enumerable
     nil
   end
 
+  def filter_map
+    return enum_for(:filter_map) unless block_given?
+
+    gather = ->(item) { item.size <= 1 ? item.first : item }
+
+    arr = []
+    each do |*item|
+      item = yield(gather.(item))
+      arr << item if item
+    end
+    arr
+  end
+
   alias detect find
 
   def grep(pattern)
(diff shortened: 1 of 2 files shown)


src/enumerable.rb
```diff
@@ -314,6 +314,20 @@ def find(if_none = nil)
...
 
+  def filter_map
+    return enum_for(:filter_map) unless block_given?
+
+    gather = ->(item) { item.size <= 1 ? item.first : item }
+
+    arr = []
+    each do |*item|
+      if yield(gather.(item))
+        arr << yield(gather.(item))
```
seven1m, line 325: I could be wrong, but I don’t think we should yield to the block twice. What if the block has side effects?

seven1m (changes requested):

seven1m:
Also, can you word your commit message in imperative from? i.e. “Implement Enumerable#filter_map”

Thanks!

seven1m:
Sorry to nag, but you did it again: in your commit message, "Avoiding" should be "Avoid". 

Side note: Ideally this PR would just be one commit since it's one logical chunk of work, but I'd settle for imperative voice in the commit message subject. Thanks!

src/enumerable.rb
borisromanov: Yup, no reason to call yield twice here.

seven1m:
Thanks!

PR #65, opened 2021-08-05
Implement Enumerable#slice_after and Enumerable#slice_before

richardboehme:
This patch implements Enumerable#slice_after, Enumerable#slice_before and some more stuff in spec.rb.

I'm particularly not happy with the argument checking... Also I do not really get the issue with the FIXME above it. When moving the condition directly into the if-statement some specs are not passing...

### DIFF
--- a/spec/core/enumerable/slice_before_spec.rb
+++ b/spec/core/enumerable/slice_before_spec.rb
@@ -1,5 +1,3 @@
-# skip-test
-
 require_relative '../../spec_helper'
 require_relative 'fixtures/classes'
 require_relative 'shared/enumerable_enumeratorized'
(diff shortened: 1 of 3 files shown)


seven1m:
I found the bug causing the FIXME you mentioned.

```ruby
# wat.rb
def x
  puts 'should only print once'
  false
end

if x && false
  puts 'true'
else
  puts 'false'
end
```

```
→ ruby wat.rb
should only print once
false

→ bin/natalie wat.rb
should only print once
should only print once
false
```

`&&` is causing the left side to be eval'd twice when false. The bug is in `process_and` in pass 1 of the compiler I believe. I'll work on getting that fixed up!

seven1m:
I was able to fix that bug in 8ff368d9ecadf59cd95e207e84aa6cc0e752fde7. Thanks for noticing it!

seven1m:
Hmmm, that fix didn't seem to fix the bug here. Must be a different bug... :thinking:

PR #68, opened 2021-08-06
Implement Enumerable#entries

borisromanov:
[#19]

Let me know if this is a stretch. Looking over the ruby [`#entries`](https://ruby-doc.org/core-2.7.3/Enumerable.html#method-i-entries) documentation, it seemed like the entries implementation didn't seem to far off from `#to_a`.


seven1m:
Works for me!

PR #69, opened 2021-08-07
Implement Enumerable#uniq

richardboehme:
There are some points were I'm not sure if the implementation fits our needs:

* In Ruby `[].method(:uniq).owner` will be Array even though the method is pretty much the same than the one in Enumerable. Should we also keep a copy of `Enumerable#uniq` in Array?
* I'm not very confident about the change on the Hash class. It's not very clear how Ruby does it but with this solution I could make the specs pass.
* I implemented `Enumerable#uniq` by using `to_a.uniq!`. However, `Array#uniq!` returns `nil` if no change was made so there is a little workaround there to return the unchanged Array if `to_a.uniq!` returns `nil`. Not sure if this is the best solution for this.


src/hash_value.cpp
```diff
@@ -43,7 +47,11 @@ ValuePtr HashValue::set_default(Env *env, ValuePtr value) {
 void HashValue::put(Env *env, ValuePtr key, ValuePtr val) {
     assert_not_frozen(env);
     Key key_container;
+    if (key->is_string() && !key->is_frozen()) {
+        key = key->as_string()->dup(env);
+    }
```
seven1m, line 52: I had no idea Ruby did this. Seems so inefficient!

Here's something else that's weirding me out:

```ruby
s = "foo"
s2 = s.dup

h1 = { s => "bar" }
h2 = { s => "bar" }
h3 = { s2 => "bar" }
h4 = { "foo" => "bar" }

p s.object_id             # => 60
p h1.keys.first.object_id # => 80
p h2.keys.first.object_id # => 80
p h3.keys.first.object_id # => 80
p h4.keys.first.object_id # => 80
```

It's as if there is a cache of these `dup`d string keys and MRI somehow reuses them. :exploding_head:

richardboehme: Yeah that was a bit weird to me too. I just added it because the specs demanded for it 😅

PR #70, opened 2021-08-08
Add Enumerable#max_by and Enumerable#min_by methods

kuzux:
The "min/max n elements" implementation is not the most efficient, but their implementations within Enumerable#max and Enumerable#min did not seem the veryefficient either, so I figured this was "good enough"

### DIFF
--- a/spec/core/enumerable/max_by_spec.rb
+++ b/spec/core/enumerable/max_by_spec.rb
@@ -1,5 +1,3 @@
-# skip-test
-
 require_relative '../../spec_helper'
 require_relative 'fixtures/classes'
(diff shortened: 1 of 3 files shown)


seven1m:
Awesome! Can you please remove the `# skip-test` line from the top of the two spec files (`max_by_spec.rb` and `min_by_spec.rb`) so the test suite will run them?

seven1m:
Oh, also, can you please word your commit message in the imperative mood? e.g. "Add Enumerable#max_by and..." ("Add" instead of "Added"). :pray: Thanks!

kuzux:
Just did these :+1:

seven1m:
Nice!!!

PR #71, opened 2021-08-10
Implement Enumerable#sum

richardboehme:
The specs needed to be modified because Rational is not implemented yet.

[#19]


richardboehme:
Mh some check failed, but I don't think it's related to this patch?

seven1m:
`core/enumerable/cycle_spec.rb` has been failing in CI intermittently, and I haven't figured out why. So I've been rerunning the tests when it happens. 😬

PR #73, opened 2021-08-10
Implement Enumerable#find_index

richardboehme:
One quick note: Ideally we would return directly from the `each` block but somehow one spec is failing when doing so...


seven1m:
You're on a roll! <3

PR #74, opened 2021-08-15
Make Enumerator more spec-compliant

richardboehme:
I'm currently fiddling with Enumerator::Lazy and discovered that I actually do not know a lot about Enumerator itself. So I decided to run some specs and try to fix some issues that came up with it.


seven1m:
🎉🎉

Issue #39, opened 2021-06-26
Implementing methods of array


davidot:
Not all the methods as specified in: https://ruby-doc.org/core-3.0.1/Array.html are implemented in natalie at the moment.
Here is a list of the methods with status.
If you want to implement one please respond to this issue or make a PR such that people are all doing the same thing.
(Note i removed the already implemented methods to clean this up)
Also some of the methods must return an Enumerable see #19)
Legend:

Symbol | Meaning
--- | ------------
:heavy_check_mark: | Implemented completely (after this issue was made)
:warning: | Partial implementation
:gear: | Being worked on
:grey_question: | Status unclear
(Nothing) | Nothing implemented or tested

Methods

Status | Method(s) | Notes, PR
|---- | :------------- | :----
| :warning: | `::[]` |
| :warning: | `::new` |
| | `::try_convert` |
| | `#&` |
| | `#*` |
| :warning: | `#[]` | Does not return nil for `a[(size + 1)..x]`
| | `#[]=` | Does not work for negative index, does not work for ranges, ...
| :warning: | `#append, #push` | Append is not an alias for push
| | `#assoc` |
| | `#at` |
| | `#bsearch` |
| | `#bsearch_index` |
| | `#clear` |
| | `#collect, #map` | Collect is not an alias for map. Map cannot be called without block 
| | `#combination` |
| | `#compact!` |
| | `#concat` |
| | `#count` |
| | `#cycle` |
| | `#deconstruct` |
| | `#delete` |
| | `#delete_at` |
| | `#delete_if` |
| | `#difference` |
| | `#dig` |
| | `#drop` |
| | `#drop_while` |
| :warning: | `#each` | Cannot modify array while iterating
| | `#each_index` |
| | `#fetch` |
| | `#fill` |
| | `#filter!, #select!` |
| | `#find_index` | find_index is not an alias for index
| | `#flatten` |
| | `#flatten!` |
| | `#hash` |
| :grey_question: | `#initialize_copy` | Missing spec
| | `#insert` | 
| | `#intersection` |
| :warning: | `#join` | Recursive has brackets
| | `#keep_if` |
| | `#map!` |
| :warning: | `#max` | With `n` argument not implemented 
| :warning: | `#min` | With `n` argument not implemented
| | `#minmax` |
| | `#none?` |
| :grey_question: | `#old_to_s` | Do we want/need this? Not implemented (alias for `to_s`)
| | `#one?` |
| | `#pack` |
[...]

PR #75, opened 2021-08-21
Implement Array::flatten

TheGrizzlyDev:
Partially implements Array::flatten from #39 with the exception of:
- Returning an instance of a subtype of Array when flatten is invoked on said subtype
- Make use of respond_to_missing and method_missing in value for when respond_to would otherwise return false. These methods seem to be more something we should implement in Value and in a separate PR/Issue, but if preferred I'll gladly cover both in this PR.

I'm quite new to both Ruby and C++ so I'll be extremely glad to take any suggestion and critique :)

### DIFF
--- a/include/natalie/array_value.hpp
+++ b/include/natalie/array_value.hpp
@@ -157,7 +157,12 @@ private:
         : Value { Value::Type::Array, GlobalEnv::the()->Array() }
         , m_vector { std::move(vector) } { }
 
+    ArrayValue(ClassValue *klass)
+        : Value { Value::Type::Array, klass } { }
+
     Vector<ValuePtr> m_vector {};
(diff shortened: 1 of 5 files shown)


seven1m:
Thanks for lending a hand! This is good enough to merge as-is.

If you end up working on the `Array#flatten!` method, you might look at `Array::sort` and `Array::sort_in_place` (in C++) as a pattern to follow.

Also, you might consider using a `Hashmap<ArrayValue*>` as a hash set that will be a more efficient data structure for your `visited_arrays` variable. You can call `set()` and `get()` on it, and iterate over it. Let me know if you want some example code!

But this is a great start and definitely moves us forward. :hugs:

PR #76, opened 2021-08-22
Array flatten in place

TheGrizzlyDev:
Partially implements Array::flatten! from #39 with the exception of:
- Make use of respond_to_missing and method_missing in value for when respond_to would otherwise return false. These methods seem to be more something we should implement in Value and in a separate PR/Issue, but if preferred I'll gladly cover both in this PR.

Followed suggestion from #75 and replaced vector with Hashmap.

### DIFF
--- a/include/tm/vector.hpp
+++ b/include/tm/vector.hpp
@@ -115,10 +115,12 @@ public:
     }
 
     T pop_front() {
-        return pop_at(0);
+        T val = m_data[0];
+        remove(0);
+        return val;
     }
 
-    T pop_at(size_t index) {
+    void remove(size_t index) {
         assert(m_size > index);
         T val = m_data[index];
 
@@ -126,8 +128,6 @@ public:
             m_data[i - 1] = m_data[i];
         }
         --m_size;
-
-        return val;
     }
 
     bool is_empty() const { return m_size == 0; }
--- a/include/tm/vector.hpp
+++ b/include/tm/vector.hpp
@@ -100,21 +100,29 @@ public:
     }
 
     void push_front(T val) {
+        insert(0, val);
+    }
+
+    void insert(size_t index, T val) {
         if (m_size >= m_capacity) {
             grow_at_least(m_size + 1);
         }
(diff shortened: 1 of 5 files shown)


src/array_value.cpp
```diff
@@ -341,8 +341,18 @@ ValuePtr ArrayValue::first(Env *env, ValuePtr n) {
...
-    ArrayValue *array = new ArrayValue(klass());
+    ArrayValue *copy = new ArrayValue { *this };
+    copy->flatten_in_place(env, depth);
+    return copy;
+}
+
+ValuePtr ArrayValue::flatten_in_place(Env *env, ValuePtr depth) {
+    if (is_frozen()) {
+        env->raise("FrozenError", "can't modify frozen Array: {}", inspect_str(env));
+        return nullptr;
```
seven1m, line 352: No need to return anything here. `Env::raise()` is no return.

include/tm/vector.hpp
```diff
@@ -100,21 +100,29 @@ class Vector {
...
+        m_data[index] = val;
     }
 
     T pop_front() {
-        assert(m_size != 0);
-        T val = m_data[0];
+        return pop_at(0);
+    }
+
+    T pop_at(size_t index) {
```
seven1m, line 121: I guess it's personal preference, but I've never seen a Vector `pop_at` method before and its use does seem limited. You're not using the "popped" value in your code anyway, so let's rename this to `remove()` and have it be a void return.

seven1m (changes requested):
This is looking great. Just a couple small changes and then I'll merge...

include/tm/vector.hpp
TheGrizzlyDev: I agree and it makes even more sense since I've ranamed it's counterpart to insert, thanks!

TheGrizzlyDev:
> This is looking great. Just a couple small changes and then I'll merge...

Nice, I've addressed both comments, let me know what you think :)

PR #77, opened 2021-08-22
add Array::sum spec 

TheGrizzlyDev:
Addresses Array::sum for #39 
nothing to implement as sum is already provided by Enumerable and Array includes Enumerable.

### DIFF
--- a/include/natalie/array_value.hpp
+++ b/include/natalie/array_value.hpp
@@ -136,6 +136,7 @@ public:
     ValuePtr slice_in_place(Env *, ValuePtr, ValuePtr);
     ValuePtr sort(Env *, Block *);
     ValuePtr sub(Env *, ValuePtr);
+    ValuePtr sum(Env *, size_t, ValuePtr *, Block *);
     ValuePtr union_of(Env *, ValuePtr);
     ValuePtr union_of(Env *, size_t, ValuePtr *);
     ValuePtr uniq_in_place(Env *, Block *);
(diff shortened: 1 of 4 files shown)


seven1m:
While Array does include Enumerable, and Enumerable does provide a `sum` method, it seems that MRI implements its own `Array#sum` method:

```ruby
irb(main):001:0> [1].method(:sum)
=> #<Method: Array#sum(*)>
irb(main):002:0> [1].method(:sum).owner
=> Array
```

Do you think it is helpful for this spec to be committed and exercising the wrong method? I'm kinda torn here. Passing tests are usually just fine and dandy, but in this case, I think the passing test does a disservice.

TheGrizzlyDev:
If that's the case then I'd rather not merge this just yet and work on the method, I don't like the idea of a passing test giving you the impression that things are working well or else this could turn into a nasty little bug we might forget about eventually. On the other hand it might make sense for the implementation of sum to still delegate to the Enumerable one, so it claims ownership but we avoid duplicating the logic, what do you think?

seven1m:
There is a pattern of delegating work from Array to Enumerable. It's not perfect, but it's good enough for now. If you want to utilize it in this case, I'd be fine with that.

There is an example of delegating to an Enumerable method here: https://github.com/seven1m/natalie/blob/56a8941e72c2bd5bb0688a892dd4d90caab04e32/src/array_value.cpp#L987-L992

TheGrizzlyDev:
Looks very good, since it looks like it's been repeated multiple times and could be a more generic mechanism to delegate to a module, would it make sense to make it a protected method for Value?

seven1m:
No, it's a broken pattern that I want to stick out like it does. We need an alternative, but I don't have one yet.

TheGrizzlyDev:
Done, anyhow if you have time I'd be curious about why this pattern doesn't quite work, thanks!

seven1m:
Sure, here is why it doesn't accurately represent what MRI does:

```ruby
module Enumerable
  def one?
    false # overridden by the user
  end
end

p [1].one? # => prints true in MRI
```

In Natalie, the above will print `false` because we do not have two separate implementations.

PR #79, opened 2021-08-22
add spec for Array::compact and compact! and implement Array compact!

TheGrizzlyDev:
Implement Array::compact! from #39 and adds the spec for both compact and compact!

compact was already implemented and I kept it as it easy as it is much simpler to add elements to a new array rather than creating a copy of the current one and then compacting that.


seven1m:
This is great, thanks!

PR #80, opened 2021-08-22
Implement Enumator::Lazy

richardboehme:
This is a draft PR for keeping track of the development of Enumerator::Lazy. As this class behaves differently than the "normal" Enumerator we have to reimplement some Enumerator methods.

The following methods have to be implemented:

- [ ] chunk
- [ ] chunk_while
- [ ] collect
- [ ] collect_concat
- [ ] drop
- [ ] drop_while
- [ ] eager
- [ ] enum_for
- [ ] filter
- [ ] filter_map
- [ ] find_all
- [ ] flat_map
- [ ] force
- [ ] grep
- [ ] grep_v
- [ ] lazy
- [ ] map
- [ ] reject
- [ ] select
- [ ] slice_after
- [ ] slice_before
- [ ] slice_when
- [ ] take
- [ ] take_while
- [ ] to_a
- [ ] to_enum
- [ ] uniq
- [ ] with_index
- [ ] zip


seven1m:
:scream: oh boy I had no idea. Feel free to break this into smaller PRs if you'd rather. I know it can be overwhelming to have a gigantic draft PR that seems to go on forever.

richardboehme:
> 😱 oh boy I had no idea. Feel free to break this into smaller PRs if you'd rather. I know it can be overwhelming to have a gigantic draft PR that seems to go on forever.

Yeah I guess that's a good idea. However, currently there is so much that seems to be broken and I want to get stuff right before having one of those PRs merged. So managing one draft PR makes more sense for me now... Maybe after I have got some more methods implemented I feel confident enough to merge. After that we can create an issue for implementing the other methods (like with Enumerable or Array).

richardboehme:
I think I implemented the most critical part of the `Enumerator::Lazy` class. With one of the last commits I changed quite a lot related to how the lazy enumerator (specifically the passed enum block) works and I'm not really happy about the change. I think it may be unfortunate that I implemented the class in ruby as we don't really have unexposed/private state there. Now I find myself using `instance_variable_set` here and there. Maybe you have a better idea to this? @seven1m 

Otherwise I think we this is ready for review!

(some specs will probably fail due to #81)

seven1m:
The clang build keeps failing. I'll try to take a look at it soon...

PR #85, opened 2021-08-28
Implement Enumerator::Lazy#drop and Enumerator::Lazy#drop_while

richardboehme:
This also removes the binding for Range#first, because Range includes the Enumerable module that provides this method.


lib/natalie/compiler/binding_gen.rb
```diff
@@ -603,7 +603,6 @@ def generate_name
 
 gen.binding('Range', 'initialize', 'RangeValue', 'initialize', argc: 2..3, pass_env: true, pass_block: false, return_type: :Value)
 gen.binding('Range', 'begin', 'RangeValue', 'begin', argc: 0, pass_env: false, pass_block: false, return_type: :Value)
-gen.binding('Range', 'first', 'RangeValue', 'begin', argc: 0, pass_env: false, pass_block: false, return_type: :Value)
```
seven1m, line 606: Actually, Range has its own `first` method:

```ruby
irb(main):001:0> r = 1..10
=> 1..10
irb(main):002:0> r.method(:first)
=> #<Method: Range#first(*)>
irb(main):003:0> r.method(:first).owner
=> Range
```

richardboehme: Mh yeah.. I'll look into a possible alternative implementation for Range#first

richardboehme: Added a new implementation for Range#first that supports an integer argument.

PR #86, opened 2021-08-30
Implement Enumerator::Lazy#grep and #grep_v

richardboehme:
This also fixes a critical bug in Enumerator::Lazy which I described in the commit message and added a test for. Basically the following example failed:

```ruby
enum = [1,2].lazy
select = enum.select { true }
select.to_a
# => [1,2]
enum.to_a
# => []
```

In MRI Ruby both calls to #to_a returned `[1, 2]`. This happened because we consumed Enumerator objects when passed to Enumerator::Lazy. Now we rewind them before looping which makes this work.


seven1m:
Enumerator::Lazy is pretty cool!

PR #93, opened 2021-09-01
Array zip implementation

TheGrizzlyDev:
Array::zip from #39 
The implementation itself is trivial as it relies on zip from enumerable, but that and some good old investigation on MRI revealed that in fact, the ruby spec was not explicitly covering the scenario where calling :each would return a non-array value, which is in fact supported by MRI

```ruby
class E 
 include Enumberable

 def each
  yield 1
  yield 2
  yield 3
 end
end

irb(main):025:0> E.new.zip(E.new)
=> [[1, 1], [2, 2], [3, 3]]

# Natalie would have returned an error saying not every item was an array, lots of fun :)
```

### DIFF
--- a/include/natalie/array_value.hpp
+++ b/include/natalie/array_value.hpp
@@ -146,6 +146,7 @@ public:
     ValuePtr union_of(Env *, ValuePtr);
     ValuePtr union_of(Env *, size_t, ValuePtr *);
     ValuePtr uniq_in_place(Env *, Block *);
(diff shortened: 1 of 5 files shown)


seven1m:
Nice catch! 😅

TheGrizzlyDev:
> Nice catch!

Thanks! Learned from my past mistake and now I always double check with MRI ahahahha

PR #94, opened 2021-09-01
add spec for to_a and to_ary

TheGrizzlyDev:
as per #39.
The implementation was already good so I've just added the specs, making sure to exclude recursive equality checks which would always end up in an endless loop.


seven1m:
Nice!

PR #97, opened 2021-09-02
Array to_s and make inspect spec-compliant

TheGrizzlyDev:
Implement Array::to_s and inspect from #39 
Added spec for to_s and inspect. To_s really just calls inspect so I made inspect spec compliant. Classical case of to_birds with one stone. Sorry...

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -33,24 +33,34 @@ ValuePtr ArrayValue::initialize(Env *env, ValuePtr size, ValuePtr value, Block *
 }
 
 ValuePtr ArrayValue::inspect(Env *env) {
+    return _inspect(env);
+}
+
+ValuePtr ArrayValue::_inspect(Env *env, Hashmap<ArrayValue *> visited) {
     StringValue *out = new StringValue { "[" };
+    visited.set(this);
     for (size_t i = 0; i < size(); i++) {
         ValuePtr obj = (*this)[i];
-        auto inspected_repr = obj.send(env, SymbolValue::intern("inspect"));
-        SymbolValue *to_s = SymbolValue::intern("to_s");
 
-        if (!inspected_repr->is_string()) {
-            if (inspected_repr->respond_to(env, to_s)) {
-                inspected_repr = obj.send(env, to_s);
-            }
-        }
-
-        if (inspected_repr->is_string()) {
-            out->append(env, inspected_repr->as_string());
+        if (obj->is_array() && visited.get(obj->as_array()) != nullptr) {
+            out->append(env, "[...]");
         } else {
-            char buf[1000];
-            sprintf(buf, "#<%s:%p>", inspected_repr->klass()->class_name_or_blank()->c_str(), (void *)&inspected_repr);
-            out->append(env, buf);
+            auto inspected_repr = obj.send(env, SymbolValue::intern("inspect"));
+            SymbolValue *to_s = SymbolValue::intern("to_s");
+
+            if (!inspected_repr->is_string()) {
+                if (inspected_repr->respond_to(env, to_s)) {
+                    inspected_repr = obj.send(env, to_s);
+                }
+            }
+
+            if (inspected_repr->is_string()) {
+                out->append(env, inspected_repr->as_string());
+            } else {
+                char buf[1000];
+                sprintf(buf, "#<%s:%p>", inspected_repr->klass()->class_name_or_blank()->c_str(), (void *)&inspected_repr);
+                out->append(env, buf);
+            }
         }
 
         if (i < size() - 1) {
@@ -58,6 +68,7 @@ ValuePtr ArrayValue::inspect(Env *env) {
         }
     }
     out->append_char(env, ']');
+    visited.remove(this);
     return out;
 }
 
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -36,8 +36,23 @@ ValuePtr ArrayValue::inspect(Env *env) {
     StringValue *out = new StringValue { "[" };
     for (size_t i = 0; i < size(); i++) {
         ValuePtr obj = (*this)[i];
-        StringValue *repr = obj.send(env, SymbolValue::intern("inspect"))->as_string();
-        out->append(env, repr);
+        auto inspected_repr = obj.send(env, SymbolValue::intern("inspect"));
+        SymbolValue *to_s = SymbolValue::intern("to_s");
+
+        if (!inspected_repr->is_string()) {
+            if (inspected_repr->respond_to(env, to_s)) {
+                inspected_repr = obj.send(env, to_s);
+            }
+        }
+
+        if (inspected_repr->is_string()) {
+            out->append(env, inspected_repr->as_string());
+        } else {
+            char buf[1000];
+            sprintf(buf, "#<%s:%p>", inspected_repr->klass()->class_name_or_blank()->c_str(), (void *)&inspected_repr);
+            out->append(env, buf);
+        }
+
         if (i < size() - 1) {
             out->append(env, ", ");
         }
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -42,8 +42,13 @@ ValuePtr ArrayValue::_inspect(Env *env, Hashmap<ArrayValue *> visited) {
     for (size_t i = 0; i < size(); i++) {
         ValuePtr obj = (*this)[i];
 
-        if (obj->is_array() && visited.get(obj->as_array()) != nullptr) {
-            out->append(env, "[...]");
+        if (obj->is_array()) {
+            auto array_val = obj->as_array();
+            if (visited.get(array_val) != nullptr) {
+                out->append(env, "[...]");
+            } else {
+                out->append(env, array_val->_inspect(env, visited));
+            }
         } else {
             auto inspected_repr = obj.send(env, SymbolValue::intern("inspect"));
             SymbolValue *to_s = SymbolValue::intern("to_s");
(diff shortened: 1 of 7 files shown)


src/array_value.cpp
```diff
@@ -33,16 +33,42 @@ ValuePtr ArrayValue::initialize(Env *env, ValuePtr size, ValuePtr value, Block *
 }
 
 ValuePtr ArrayValue::inspect(Env *env) {
+    return _inspect(env);
+}
+
+ValuePtr ArrayValue::_inspect(Env *env, Hashmap<ArrayValue *> visited) {
```
seven1m, line 39: I don't dislike the `visited` hashmap you used here (it work fine), but for consistency, what do you think about using the `has_inspecting_flag()` mechanism I used over in HashValue?

https://github.com/seven1m/natalie/blob/38f60f0094433ac0669b78bcfd2c6276e00bb97d/src/hash_value.cpp#L226-L228

TheGrizzlyDev: So the reason why I did that is quite simple actually :) with arrays you can have neighbors that are the same without making that a loop.
If a=[b, b] and b is any non-recoursive array, then this won't be a loop, so you'd end up having to handle that additional case. Now you could argue that we are doing that regardless with set and remove itself on each inspect on an array, but the advantage of this is that we are not changing the state of array which for an array would mutate depending on the method call, whereas in an hash that won't happen since neighbors won't exist. I'm not entirely sure I made myself clear, so let me know if I wasn't :) but great question anyway!

seven1m: My way still works with `a=[b, b]`. The `b` object (assuming it's a Hash or Array) will be marked with the "inspecting" flag once for the first time it's inspected (then removed), then again marked for the second time it's inspected (then removed again).

I'm not following your logic on Hash vs Array and neighbors, actually. :man_shrugging: 

Unless you can think of another example where my way won't work, I think I prefer it since it doesn't require the overhead of a new data structure.

TheGrizzlyDev: Sorry, I think I get the idea now, no worries, will have a look at it tomorrow then, thanks 👍

seven1m: Thank you!!

TheGrizzlyDev: I do have a doubt now though :) What if calling to_s on a child fails while doing an inspect? If we use a flag that is a member of Array, would we not end up with a dirty context? That would cause the next call to behave differently as the inspecting_flag would be true. A stateless approach would handle such cleanup gracefully without any extra intervention as visited is recreated each and every time we run inspect or any similarly recursive method :)

TheGrizzlyDev: Here is an example of this issue with stateful tracking for recursion loops :) 
```ruby
class Boom
    attr :count

    def initialize
        @count = 0
    end

    def inspect
        @count += 1
        if @count % 2 == 0
            raise StandardError.new "I'll just go ka-boom" 
        end
        "Up and counting #{@count}"
    end
end

a = [[Boom.new]]

p a.inspect
begin
    p a.inspect
rescue StandardError
    p "It went boom!"
end
p a.inspect

```


 % ruby ./dirty_inspect.rb       
"[[Up and counting 1]]"
"It went boom!"
"[[Up and counting 3]]"


 % ./bin/natalie dirty_inspect.rb
"[[Up and counting 1]]"
"It went boom!"
"[[...]]"


I guess your solution would not be wrong if there were no potential side-effects in inspect, but since we call to_s and inspect and those can be changed, I reckon a stateless approach that does not need any cleanup would work best. What do you think? :)

TheGrizzlyDev:
```ruby
class Boom
    attr :count

    def initialize
        @count = 0
    end

    def inspect
        @count += 1
        if @count % 2 == 0
            raise StandardError.new "I'll just go ka-boom" 
        end
        "Up and counting #{@count}"
    end
end

a = { :b => { :c => Boom.new } }

p a.inspect
begin
    p a.inspect
rescue StandardError
    p "It went boom!"
end
p a.inspect
```

 % ruby ./dirty_inspect.rb
"{:b=>{:c=>Up and counting 1}}"
"It went boom!"
"{:b=>{:c=>Up and counting 3}}"

 % ./bin/natalie dirty_inspect.rb
"{:b=>{:c=>Up and counting 1}}"
"It went boom!"
"{...}"

Same issue with hash actually :|

seven1m:
WELL, ok, you got me with that one! :stuck_out_tongue_closed_eyes: 

Question though... what I don't understand about your solution is how do you pass the `visited` hash down to children when calling inspect on those?

If you have the following data structure... won't your solution get stuck recursively calling each other?

```ruby
# MRI

a = []
b = [a]
a[0] = b

p a.inspect # => "[[[...]]]"
p b.inspect # => "[[[...]]]"
```

src/array_value.cpp
seven1m: WELL, ok, you got me with that one! :stuck_out_tongue_closed_eyes: 

Question though... what I don't understand about your solution is how do you pass the `visited` hash down to children when calling inspect on those?

If you have the following data structure... won't your solution get stuck recursively calling each other?

```ruby
# MRI

a = []
b = [a]
a[0] = b

p a.inspect # => "[[[...]]]"
p b.inspect # => "[[[...]]]"
```

TheGrizzlyDev: > WELL, ok, you got me with that one! stuck_out_tongue_closed_eyes

WELL, thanks for making it sound like something I thought about since the get go ahhahaha. I actually realised this issue right before pushing the change to a flag based detection ahhahaha

in a way it really is the same as the flag. You flag (set) and unflag (remove) yourself when you are an array and instead of checking if you are already in there, you just check that for your children instead and don't even invoke inspect if it's a recursive loop :)

TheGrizzlyDev: `p a.inspect # => "[[[...]]]"`

(visited = [])
a -> puts itself in visited (visited = [a])
   -> checks if b is already in visited
   -> since it isn't it calls inspect
       b -> puts itself in visited (visited = [a, b])
           ->  checks if a is already in visited
           -> since it is we just append "[...]"
           -> removes itself from visited (visited = [a])
   -> removes itself from visited (visited = [a])

seven1m: But you're not passing the `visited` hashmap down from `a` to the child `b` -- `b` cannot add itself to the same `visited` hashmap -- it will end up creating its _own_ `visited` hashmap.

I checked out your branch and tried this:

```sh
tim ~/pp/natalie array-to-s 0 → bin/natalie
nat> a=[]
[]
nat> b=[a]
[[]]
nat> a[0]=b
	5: from bin/natalie:215:in `<main>'
	4: from bin/natalie:85:in `run'
	3: from /Users/tim/pp/natalie/lib/natalie/repl.rb:34:in `go'
	2: from /Users/tim/pp/natalie/lib/natalie/repl.rb:34:in `loop'
	1: from /Users/tim/pp/natalie/lib/natalie/repl.rb:59:in `block in go'
/Users/tim/pp/natalie/lib/natalie/repl.rb:59:in `call': stack level too deep (SystemStackError)
```

seven1m: For my `add_inspecting_flag()` solution, in order to address the issue you brought up about exceptions leaving the object in a dirty state... I'm considering adding a Defer class, a la Serenity's [ScopeGuard](https://github.com/SerenityOS/serenity/blob/master/AK/ScopeGuard.h) class, that would let us set up some clean up code to run when an exception is raised and the scope gets cleaned up.

I think that would resolve the issue you mentioned...

I'm still open to use your `visited` Hashmap approach, but I'm not convinced it will easily work, unless you can find a way to pass the `visited` pointer down to all the children... 🤔

TheGrizzlyDev: > But you're not passing the visited hashmap down from a to the child b -- b cannot add itself to the same visited hashmap -- it will end up creating its own visited hashmap.

That's called 'I screwed up' actually! I didn't notice I was not passing it down since the tests were passing! Thanks a lot for that. 
PS: really like the idea of a guard and would even like a go-ish deferring mechanism bound to env. An API like

env.defer([]() {
// Do something here :) 
})

TheGrizzlyDev: @seven1m fixed the bug, thanks!

seven1m: Better! Though, now you're bypassing the `send()` call and the user's own overridden `inspect` (if any) will not be called. That's why I said I wasn't sure how you were going to pass it down to the child array... 😆

TheGrizzlyDev: That would actually be extremely interesting for classes extending Array. In that case, would MRI just call the right inspect or detect the loop and back-off with "[...]"? If it's the former then, yeah, my solution falls short in that sense, but otherwise I might have just ended up mimicking the right behaviour by mistake.

PS: I though something as trivial as a stringifying method would have been boring, but this is getting rather interesting, so cool :)

TheGrizzlyDev:
```ruby
class Arrayz < Array
  def inspect
     return "[?!?]"
  end
end

irb(main):008:0> a = Arrayz.new
=> [?!?]
irb(main):009:0> b = [a]
=> [[?!?]]
irb(main):010:0> a[0]=b
=> [[?!?]]
```

oh, this is beyond weird...

TheGrizzlyDev: it seems like it called Array::inspect, which called Array::inspect and then, only then, called Arrayz::inspect, right? Or am I getting confused?

seven1m: Oh what have we gotten ourselves into... 🤦‍♂️😆

Yeah that's super weird!

seven1m: OK, given that weirdness, which actually seems like a bug in MRI, let's go with your solution for now. If you get time and feel like  it, would you also copy your solution over to the HashValue class and make sure these two data structures can be nested + inspected?

seven1m (approved):

seven1m:
I approved the PR. You can merge when you think it's ready!

src/array_value.cpp
TheGrizzlyDev: Yeah, makes sense and if I can find a nice enough way I'll introduce a more generic way to handle recursion guards. At some point I will also test this with later and earlier versions of ruby to see how this changes over time and see if it's in fact an actual bug :)

PR #98, opened 2021-09-02
implement Array at

TheGrizzlyDev:
Implement Array::at from #39 
Added index to_int conversion when parameter is not an int


seven1m:
Well that was easy :laughing:

PR #104, opened 2021-09-05
Generic recursion guard

TheGrizzlyDev:
Added a simple and probably naive mechanism to handle recursion guarding across various objects. Can handle the now infamous dirty_inspect.rb case described in #97 

 % ./bin/natalie ./dirty_inspect.rb 
"[[Up and counting 1]]"
"It went boom!"
"[[Up and counting 3]]"


seven1m:
I love this! <3

PR #105, opened 2021-09-05
Array equality

TheGrizzlyDev:
Implemented both eql? and == from #39 .

The test "compares with an equivalent Array-like object using #to_ary" highlighted an issue on how we use respond_to? in natalie's standard library. We rely on calling the method on the c++ impl rather than send the event "respond_to? :method_name" so mocks won't be able to correctly override such value and neither will anything else :) We should replace all the direct respond_to with the ruby method and maybe create a utility that wraps `send(env, Symbol::intern("respond_to?")...)` so the ergonomics won't change much for us but natalie will in fact behave correctly. What do you think about this?

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -329,12 +329,12 @@ ValuePtr ArrayValue::eq(Env *env, ValuePtr other) {
                                               ->send(env, equality, { item->send(env, object_id) });
 
                 // This allows us to check NAN equality and other potentially similar constants
-                if (same_object_id->type() == Value::Type::True)
+                if (same_object_id->is_true())
                     continue;
             }
 
             ValuePtr result = this_item.send(env, equality, 1, &item, nullptr);
-            if (result->type() == Value::Type::False)
+            if (result->is_false())
                 return result;
         }
 
@@ -366,7 +366,7 @@ ValuePtr ArrayValue::eql(Env *env, ValuePtr other) {
         for (size_t i = 0; i < size(); ++i) {
             ValuePtr item = (*other_array)[i];
             ValuePtr result = (*this)[i].send(env, SymbolValue::intern("eql?"), 1, &item, nullptr);
-            if (result->type() == Value::Type::False)
+            if (result->is_false())
                 return result;
         }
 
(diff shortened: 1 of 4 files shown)


seven1m:
You're totally right! Good catch. Let's create a little convenience method on Value to do that. What to call it though? 🤔

seven1m:
Let's rename the existing `respond_to` to `respond_to_method` (there is prior art for this, e.g. `new_method`) so it is clear it's what the bindings point to.

Then we can make a new convenience method called `respond_to` that does `send(env, Symbol::intern("respond_to?")...)`.

Thoughts?

TheGrizzlyDev:
> Let's rename the existing `respond_to` to `respond_to_method` (there is precedent for this naming suffix, e.g. `ClassValue::new_method`, `ModuleValue::private_method`, etc.) so it is clear it's what the bindings point to.
> 
> Then we can make a new convenience method called `respond_to` that does `send(env, Symbol::intern("respond_to?")...)`.
> 
> Thoughts?

sounds lovely :) I'll create an issue then and later on I'll fix the broken build on this PR :)

src/array_value.cpp
```diff
@@ -295,17 +295,47 @@ ValuePtr ArrayValue::any(Env *env, size_t argc, ValuePtr *args, Block *block) {
...
+            ValuePtr item = (*other_array)[i];
+
+            ValuePtr same_object_id = this_item
+                ->send(env, object_id)
+                ->send(env, equality, {
+                    item->send(env, object_id)
+                });
+
+            // This allows us to check NAN equality and other potentially similar constants
+            if (same_object_id->type() == Value::Type::True)
```
seven1m, line 329: We also have `Value::is_true()` and `Value::is_false()` if you want to use them... :wink:

seven1m (approved):
Merge when you're ready!

TheGrizzlyDev:
> Merge when you're ready!

Sure, right now it's causing some tests to fail (I think because we had some equality checks on arrays in the testing framework ?), so I'll fix that and merge later on ;) thanks though!

PR #107, opened 2021-09-05
replace respond_to with the actual respond_to? method by sending it

### LINKED ISSUE #106: Send respond_to to the right receiver from the core API
Found: https://github.com/seven1m/natalie/pull/105

The test "compares with an equivalent Array-like object using #to_ary" highlighted an issue on how we use respond_to? in natalie's standard library. We rely on calling the method on the c++ impl rather than send the event "respond_to? :method_name" so mocks won't be able to correctly override such value and neither will anything else :) We should replace all the direct respond_to with the ruby method and maybe create a utility that wraps send(env, Symbol::intern("respond_to?")...) so the ergonomics won't change much for us but natalie will in fact behave correctly. 

We can rename the existing respond_to to respond_to_method (there is precedent for this naming suffix, e.g. ClassValue::new_method, ModuleValue::private_method, etc.) so it is clear it's what the bindings point to.

Then we can make a new convenience method called respond_to that does send(env, Symbol::intern("respond_to?")...).

TheGrizzlyDev:
our internal respond_to_method is still used in the abscence of the ruby method as for BaseObject. This fixed #106


src/value.cpp
```diff
@@ -485,6 +485,14 @@ bool Value::is_a(Env *env, ValuePtr val) {
 }
 
 bool Value::respond_to(Env *env, ValuePtr name_val) {
+    if (respond_to_method(env, SymbolValue::intern("respond_to?")))
+        return send(env, SymbolValue::intern("respond_to?"), { name_val })->is_true();
```
seven1m, line 489: omg I totally forgot we had an override of send that takes an initializer list. 🤦‍♂️

brb, about to do a big find+replace... 😆

seven1m (approved):

PR #108, opened 2021-09-05
Use template param for callback in RecursionGuard

seven1m:
This lets us get rid of the `functional` STL header and the `ReturnType` template param. Also, I moved the static did_run Hashmap into a static member of the RecursionGuard class.

Last, I moved this into the TM namespace because I think it's fairly generic and fits better there. The exception to this is that it includes `natalie/hashmap.hpp`, but I plan to rewrite the Hashmap implementation from scratch and put it in the TM namespace soon... :-)


TheGrizzlyDev (approved):
Very cool! Love seeing an abstraction getting improved at such an early stage!

PR #110, opened 2021-09-06
defer: Forward the callback object instead of copying it

alimpfard:
Just a minor thing I noticed while watching @seven1m's latest video, nothing too major :)

This is not going to make a difference in most cases, but there are some
possible scenarios where copying the callable object would not be
possible (e.g. when the lambda captures a smart pointer that doesn't
allow copying, like `std::unique_ptr`).


seven1m:
Cool thanks!

PR #111, opened 2021-09-06
add spec for Array::allocate

TheGrizzlyDev:
and another one down from #39 , one of the hardest ones :)

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -7,6 +7,14 @@
 
 namespace Natalie {
 
+ValuePtr ArrayValue::allocate(Env *env, size_t argc, ValuePtr *args) {
+    if (argc > 0)
+        env->raise("ArgumentError",
+            "wrong number of arguments (given {}, expected 0)",
+            argc);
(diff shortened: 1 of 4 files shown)


src/array_value.cpp
```diff
@@ -7,6 +7,14 @@
 
 namespace Natalie {
 
+ValuePtr ArrayValue::allocate(Env *env, size_t argc, ValuePtr *args) {
+    if (argc > 0)
+        env->raise("ArgumentError",
+            "wrong number of arguments (given {}, expected 0)",
+            argc);
```
seven1m, line 14: We have some convenience methods for this pattern: https://github.com/seven1m/natalie/blob/03602356b9fd89a30e79c00dba855b2b40477ab5/include/natalie/env.hpp#L67-L70

seven1m (approved):

src/array_value.cpp
TheGrizzlyDev: oh that's pretty neat!

PR #115, opened 2021-09-06
added Array::assoc spec and fixed test

TheGrizzlyDev:
as per #39


seven1m:
Wow it was close! :laughing:

PR #117, opened 2021-09-07
add Array::delete_if spec

TheGrizzlyDev:
add Array::delete_if from #39 
add empty array matcher

### DIFF
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -334,7 +334,7 @@ class EqlExpectation
   end
 end
 
-class BeEmpyExpectation
+class BeEmptyExpectation
   def match(subject)
     if (subject.length > 0)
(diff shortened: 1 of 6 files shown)


test/support/spec.rb
```diff
@@ -334,6 +334,20 @@ def inverted_match(subject)
   end
 end
 
+class BeEmpyExpectation
```
seven1m, line 337: Typo "empy" haha

seven1m (changes requested):

test/support/spec.rb
TheGrizzlyDev: Hahahaha nice catch!

seven1m (approved):

PR #118, opened 2021-09-07
Array delete method

TheGrizzlyDev:
another one for #39

### DIFF
--- a/include/natalie/array_value.hpp
+++ b/include/natalie/array_value.hpp
@@ -106,6 +106,7 @@ public:
     ValuePtr compact_in_place(Env *);
     ValuePtr concat(Env *, size_t, ValuePtr *);
     ValuePtr delete_if(Env *, Block *);
+    ValuePtr delete_item(Env *, ValuePtr, Block *);
     ValuePtr dig(Env *, size_t, ValuePtr *);
     ValuePtr drop(Env *, ValuePtr);
     ValuePtr each(Env *, Block *);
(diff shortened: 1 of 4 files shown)


seven1m:
This will be inefficient for large arrays where many items are bing removed, because all the items will be shifted left each time an item is removed.

It's good enough for now. I think we can come up with a better algorithm later...

TheGrizzlyDev:
> This will be inefficient for large arrays where many items are bing removed, because all the items will be shifted left each time an item is removed.
> 
> It's good enough for now. I think we can come up with a better algorithm later...

I somewhat agree with the thought but I am also a believer in 'good enough' for the moment :) Right now optimising isn't as important as covering is, but I reckon it won't hurt to come up with some ideas for later (it will at least be interesting :^) ).

In general I reckon that for all the add/delete/update operations that can happen in batches we should have a first phase 'planning' the actions to be later 'committed' in a single swoop. In this case we'd mark cells as to be deleted and then remove them at the end all together, so we only have a single resize and the least amount of left shifts :) 

Since we're talking about this topic: it might be interesting for aggregated operations like max, sum, min, etc. to be precomputed during insert for the simplest scenario (usually the one calling without a block) as that might speed everything up (think about all the to_int we won't have to rerun on demand each time!). Then again, the first implementation will be the dumbest possible and then we can iterate over that, but regardless it still is a lot of fun to talk about this :^)

PR #122, opened 2021-09-08
Change build tool from CMake to Rake

seven1m:
I've been considering this for awhile, so I finally bit the bullet and rewrote our CMake config for Rake.

I really like the idea of using Ruby for our build scripts, since, you know, we're working on Ruby. :laughing: 

Rake is more than capable of doing everything we need, with exception (maybe) of generating configs for ninja or some other build tool. I personally do not use ninja; I get plenty of use out of ccache that ninja doesn't provide any extra speed improvement. 

- Rake is Ruby, and we love Ruby. <3
- Rake is more flexible than CMake, at least for someone who already knows Ruby. :-)
- CMake is another technology to learn, and I don't want to.
- One less dependency to install.


seven1m:
@TheGrizzlyDev @richardboehme take note, as this will affect your workflow locally. You'll need to run `rake bulid && bin/natalie ...` instead of `make build`. :wink:

TheGrizzlyDev:
> @TheGrizzlyDev @richardboehme take note, as this will affect your workflow locally. You'll need to run `rake bulid && bin/natalie ...` instead of `make build`. :wink: 

Thanks for the heads up, then I better start working on the ARM port as it'd be a good chance to understand the new build :)

richardboehme:
> @TheGrizzlyDev @richardboehme take note, as this will affect your workflow locally. You'll need to run `rake bulid && bin/natalie ...` instead of `make build`. 😉

That's great, thanks for the note! Should we add rake as a dependency to the Gemfile? @seven1m

seven1m:
Sure, we can. But I think it's built into Ruby now too?

richardboehme:
Mh.. If I run `bundle exec rake` in the natalie repo I get
```
        4: from /home/richard/.rvm/gems/ruby-2.7.3/bin/ruby_executable_hooks:22:in `<main>'
        3: from /home/richard/.rvm/gems/ruby-2.7.3/bin/ruby_executable_hooks:22:in `eval'
        2: from /home/richard/.rvm/gems/ruby-2.7.3/bin/rake:23:in `<main>'
        1: from /usr/share/rvm/rubies/ruby-2.7.3/lib/ruby/2.7.0/bundler/rubygems_integration.rb:402:in `block in replace_bin_path'
/usr/share/rvm/rubies/ruby-2.7.3/lib/ruby/2.7.0/bundler/rubygems_integration.rb:374:in `block in replace_bin_path': can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
```

Maybe I'm just using it wrong, but I guess this should *theoretically* work?

seven1m:
Oh, you're right. Bundler doesn't like that. I've been running just `rake`. So yeah, to avoid that little oddity, let's add it to the Gemfile. :-)

PR #123, opened 2021-09-08
WIP - Port Natalie to aarch64

TheGrizzlyDev:
- setup build and CI to support aarch64

### DIFF
--- a/.github/workflows/build aarch64.yml
+++ b/.github/workflows/build aarch64.yml
@@ -1,0 +1,47 @@
+name: Build and Test on aarch64
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  DOCKER_FLAGS: ""
+  ARCH: "aarch64"
+  CXX: "aarch64-linux-gnu-g++" 
+  CC:  "aarch64-linux-gnu-gcc"
+
+jobs:
+  aarch64_gcc:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: checkout submodules
+      run: git submodule update --init --recursive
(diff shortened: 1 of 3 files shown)


.github/workflows/build aarch64.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: test release build
+      run: rake docker_test_release
+  aarch64_ruby3:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v1
+    - name: checkout submodules
+      run: git submodule update --init --recursive
+    - name: test with Ruby 3.0
+      run: rake docker_test_gcc
```
seven1m, line 47: This file seems to have a space in the name? Not sure if that will work right, and probably best to avoid it just in case.

Rakefile
```diff
@@ -144,10 +144,11 @@ file 'build/libonigmo.a' do
   build_dir = File.expand_path('build/onigmo', __dir__)
   rm_rf build_dir
   cp_r 'ext/onigmo', build_dir
+  arch_conf = "--host #{ENV['ARCH']}" if ENV['ARCH']
   sh <<-SH
     cd #{build_dir} && \
     sh autogen.sh && \
-    ./configure --with-pic --prefix #{build_dir} && \
+    ./configure --with-pic #{arch_conf} --prefix #{build_dir} && \
```
seven1m, line 151: Nice!

seven1m:
This is great, though I don't think we can merge as-is since it would break the build? I'm going to close this for now... let me know if you want to make this a nightly build instead, like we do with the full ruby/spec suite. Setting this up that way would allow us to work on it incrementally!

PR #128, opened 2021-09-11
make Array constructor spec compliant

TheGrizzlyDev:
Another one down from #39 , the implementation was correct but was not passing down the correct subclass as MRI does.


seven1m:
Nice! Glad you found `pass_klass` :wink:

TheGrizzlyDev:
> Nice! Glad you found `pass_klass` :wink: 



Ez

PR #139, opened 2021-09-12
implement Array hash

TheGrizzlyDev:
As per #39 
Currently excluding tests that are using Array::fill and tests requiring the correct implementation of Hash::hash

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -5,6 +5,7 @@
 #include <natalie/string_value.hpp>
 #include <natalie/symbol_value.hpp>
 #include <random>
+#include <stdlib.h>
 
 namespace Natalie {
 
@@ -1210,6 +1211,7 @@ ValuePtr ArrayValue::hash(Env *env) {
     return guard.run([&](bool is_recursive) {
         if (is_recursive)
             return ValuePtr::integer(0);
+        srand(0);
         nat_int_t accumulator = 0;
         auto hash_method = SymbolValue::intern("hash");
         auto to_int = SymbolValue::intern("to_int");
@@ -1219,7 +1221,7 @@ ValuePtr ArrayValue::hash(Env *env) {
             if (! current_hash->is_integer() && current_hash->respond_to(env, to_int))
                 current_hash = current_hash->send(env, to_int);
 
-            accumulator += current_hash->as_integer()->to_nat_int_t();
+            accumulator += current_hash->as_integer()->to_nat_int_t() * (rand() % size() + 1);;
         }
 
         return ValuePtr::integer(accumulator);
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -5,7 +5,6 @@
 #include <natalie/string_value.hpp>
 #include <natalie/symbol_value.hpp>
 #include <random>
-#include <stdlib.h>
 
 namespace Natalie {
 
@@ -1211,17 +1210,18 @@ ValuePtr ArrayValue::hash(Env *env) {
     return guard.run([&](bool is_recursive) {
         if (is_recursive)
             return ValuePtr::integer(0);
-        srand(0);
         nat_int_t accumulator = 0;
         auto hash_method = SymbolValue::intern("hash");
         auto to_int = SymbolValue::intern("to_int");
-        for (auto &item : *this) {
+        for (size_t i = 0; i < size(); ++i) {
+            auto item = (*this)[i];
             auto current_hash = item->send(env, hash_method);
 
             if (! current_hash->is_integer() && current_hash->respond_to(env, to_int))
                 current_hash = current_hash->send(env, to_int);
 
-            accumulator += current_hash->as_integer()->to_nat_int_t() * (rand() % size() + 1);;
+            auto pre_hash = current_hash->as_integer()->to_nat_int_t() + hashmap_hash_string(&i);
+            accumulator += hashmap_hash_string(&pre_hash);
         }
 
         return ValuePtr::integer(accumulator);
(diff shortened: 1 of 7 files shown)


TheGrizzlyDev:
Just noticed an issue with this

```
nat> [1,2].hash
30368179215
nat> [2,1].hash
30368179215
```

```
irb(main):001:0> [1,2].hash
=> -1930233398629545187
irb(main):002:0> [2,1].hash
=> -1827901450314679638
```

My implementation is ignoring ordering. In order to have a stable enough implementation with an acceptable risk of collision I'll just add a random value on each turn of hash when iterating over its elements and seed it at the start of hash with 0

TheGrizzlyDev:
Fixed the comment above :^)

src/array_value.cpp
```diff
@@ -1205,6 +1206,28 @@ ValuePtr ArrayValue::rassoc(Env *env, ValuePtr needle) {
...
+        nat_int_t accumulator = 0;
+        auto hash_method = SymbolValue::intern("hash");
+        auto to_int = SymbolValue::intern("to_int");
+        for (auto &item : *this) {
+            auto current_hash = item->send(env, hash_method);
+
+            if (! current_hash->is_integer() && current_hash->respond_to(env, to_int))
+                current_hash = current_hash->send(env, to_int);
+
+            accumulator += current_hash->as_integer()->to_nat_int_t() * (rand() % size() + 1);;
```
seven1m, line 1224: While I appreciate the ingenuity :nerd_face:  here of using `rand()`, I think we should use a proper hash algorithm. We have a hash alg for nat_int_t [here](https://github.com/seven1m/natalie/blob/master/src/hashmap.cpp#L581-L588), which I borrowed from Stack Overflow. Maybe we copy-paste that and use it here?

TheGrizzlyDev: oh boy you just saved me from reimplementing this myself ahahahha. I was going to do a followup PR, thanks!

PS: in hindsight just copying a similar use case would have made sense :P

TheGrizzlyDev: I think it is worth just extracting those functions as the same problem will exist for Hash and other objects. A good approach could just be:

`hash(array[index].hash + hash(index))`

What do you think about it? 
PS: since this is already blowing up a bit instead of being incremental as I wanted I could just go ahead and cover Hash::hash as well which would cover an additional scenario here :)

seven1m: Yeah I think that sounds good!

TheGrizzlyDev: finally got back to it and hopefully addressed your comment! Let me know what you think, please :D 

On a sidenote: hashing functions are quite generic and used throughout various hashing methods, would it make sense to extract them out Hashmap?

TheGrizzlyDev: Snap it's broken, the idea still stands, but I will be fixing the tests tomorrow :D

seven1m:
@TheGrizzlyDev I wonder what you think about this implementation of ArrayValue::hash:

```cpp
ValuePtr ArrayValue::hash(Env *env) {
    TM::RecursionGuard guard { this };
    return guard.run([&](bool is_recursive) {
        if (is_recursive)
            return ValuePtr { NilValue::the() };
        nat_int_t hash = 100019; // just a prime number :-)
        auto hash_method = SymbolValue::intern("hash");
        auto to_int = SymbolValue::intern("to_int");

        for (auto &item : *this) {
            auto item_hash = item->send(env, hash_method);

            if (item_hash->is_nil())
                continue;

            if (!item_hash->is_integer() && item_hash->respond_to(env, to_int))
                item_hash = item_hash->send(env, to_int);

            nat_int_t h = item_hash->as_integer()->to_nat_int_t();
            hash ^= h * 207269; // another prime number to prevent bias toward zero
        }

        return ValuePtr::integer(hash);
    });
}
```

It doesn't yet handle recursive arrays, but it will easily once we do the work I mentioned in Discord to augment the RecursionGuard... :wink:

TheGrizzlyDev:
> @TheGrizzlyDev I wonder what you think about this implementation of ArrayValue::hash...
> 
> ```c++
> ValuePtr ArrayValue::hash(Env *env) {
>     TM::RecursionGuard guard { this };
>     return guard.run([&](bool is_recursive) {
>         if (is_recursive)
>             return ValuePtr { NilValue::the() };
>         nat_int_t hash = 100019; // just a prime number :-)
>         auto hash_method = SymbolValue::intern("hash");
>         auto to_int = SymbolValue::intern("to_int");
> 
>         for (auto &item : *this) {
>             auto item_hash = item->send(env, hash_method);
> 
>             if (item_hash->is_nil())
>                 continue;
> 
>             if (!item_hash->is_integer() && item_hash->respond_to(env, to_int))
>                 item_hash = item_hash->send(env, to_int);
> 
>             nat_int_t h = item_hash->as_integer()->to_nat_int_t();
>             hash ^= h * 207269; // another prime number to prevent bias toward zero
>         }
> 
>         return ValuePtr::integer(hash);
>     });
> }
> ```
> 
> It doesn't yet handle recursive arrays, but it will easily once we do the work I mentioned in Discord to augment the RecursionGuard...  (it passes the other tests)

Thanks a lot! Sure, I like it :D also, finally back from a pretty busy week and hopefully now I should have some time to get some more of this array done!

seven1m:
I tweaked your `ArrayValue::hash` method in b8b82829cb7ad3b5d1ec34a66154c48d0f9a1648 and merged most of this to master. I did not include the `HashValue::hash` stuff as I think that can be another PR. I'm going to close this as it is mostly done. Feel free to open a new PR for `HashValue::hash`!

PR #143, opened 2021-09-12
Add a Github Action workflow to run all ruby specs

richardboehme:
We are constantly working on improving spec compliance for Natalie. I think it is very interesting to see the progress we are making with this. This Github Action workflow will run one time each day and run all specs from ruby/spec against Natalie. 

The current results are:
```
Success: 2525
Failures: 1389
Errors: 6202
Compile Errors: 1403
Crashes: 847
Timeouts: 7
```

Compile Errors, Crashes and Timeouts count per file while Success, Failures and Errors count per example. The results can only be seen manually by looking at the action log for now.

### DIFF
--- a/.github/workflows/run_all_specs.yml
+++ b/.github/workflows/run_all_specs.yml
@@ -1,0 +1,29 @@
+name: Run all specs from ruby/spec
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+env:
+  BUNDLE_WITH: run_all_specs
+
+jobs:
+  specs:
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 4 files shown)


richardboehme:
Also note that the specs are run against the ruby version from https://github.com/seven1m/natalie/blob/master/.ruby-version

We may consider running it with 2.7 and 3.0? 🤔

seven1m:
I think just 2.7 is fine for now -  our main suite should (eventually) catch differences between hosting Ruby versions.

That said, @TheGrizzlyDev and I have been discussing going all in on Ruby 3, so I can see us switching everything to Ruby 3 soon-ish...

For now this is fantastic!

PR #144, opened 2021-09-14
Implement Hash#to_proc

richardboehme:
Please not that this targets Ruby 2.7. With Ruby version > 3 this should return a lambda instead of a proc.

### DIFF
--- a/spec/core/hash/to_proc_spec.rb
+++ b/spec/core/hash/to_proc_spec.rb
@@ -1,0 +1,101 @@
+require_relative '../../spec_helper'
+require_relative 'fixtures/classes'
+
+describe "Hash#to_proc" do
+  before :each do
+    @key = Object.new
+    @value = Object.new
+    @hash = { @key => @value }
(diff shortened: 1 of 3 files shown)


richardboehme:
Mh I expected the ruby3 job to fail. Is there even a difference in the configuration between the ruby3 job and the standard gcc job?

seven1m:
@richardboehme the ruby3 job is just the *hosting* Ruby version -- not the Ruby version Natalie reports to the spec runner. Natalie still thinks she's 2.7. 😉

richardboehme:
> @richardboehme the ruby3 job is just the _hosting_ Ruby version -- not the Ruby version Natalie reports to the spec runner. Natalie still thinks she's 2.7. 😉

Ah ok, thanks. I still don't see the configuration difference between the gcc test and the ruby 3 test. Both are running the same rake command, aren't they?

richardboehme:
> `docker_test_ruby3`

Fixed with #148

PR #150, opened 2021-09-15
Implement all Hash#each* methods

richardboehme:
Implements:
* `Hash#each_key`
* `Hash#each_value`

Add spec for `Hash#each_pair`

### DIFF
--- a/spec/core/hash/each_key_spec.rb
+++ b/spec/core/hash/each_key_spec.rb
@@ -1,0 +1,23 @@
+require_relative '../../spec_helper'
+require_relative 'fixtures/classes'
+require_relative 'shared/iteration'
+require_relative '../enumerable/shared/enumeratorized'
+
+describe "Hash#each_key" do
+  it "calls block once for each key, passing key" do
+    r = {}
+    h = { 1 => -1, 2 => -2, 3 => -3, 4 => -4 }
+    h.each_key { |k| r[k] = k }.should equal(h)
+    r.should == { 1 => 1, 2 => 2, 3 => 3, 4 => 4 }
+  end
+
+  it "processes keys in the same order as keys()" do
+    keys = []
+    h = { 1 => -1, 2 => -2, 3 => -3, 4 => -4 }
+    h.each_key { |k| keys << k }
+    keys.should == h.keys
+  end
+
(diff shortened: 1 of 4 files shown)


src/hash_value.cpp
```diff
@@ -378,6 +378,22 @@ ValuePtr HashValue::each(Env *env, Block *block) {
     return this;
 }
 
+ValuePtr HashValue::each_key(Env *env, Block *block) {
+    if (!block)
+        return send(env, SymbolValue::intern("enum_for"), { SymbolValue::intern("each_key") });
+
+    keys(env)->send(env, SymbolValue::intern("each"), {}, block);
+    return this;
+}
```
seven1m, line 387: What do you think about implementing this in `hash.rb` so we can avoid the creation of the array of keys:

```ruby
def each_key
  each do |key, _|
    yield key
  end
end
```

src/hash_value.cpp
```diff
@@ -378,6 +378,22 @@ ValuePtr HashValue::each(Env *env, Block *block) {
...
+
+ValuePtr HashValue::each_value(Env *env, Block *block) {
+    if (!block)
+        return send(env, SymbolValue::intern("enum_for"), { SymbolValue::intern("each_value") });
+
+    values(env)->send(env, SymbolValue::intern("each"), {}, block);
+    return this;
+}
```
seven1m, line 395: Same:

```ruby
def each_value
  each do |_, value|
    yield value
  end
end
```

seven1m (changes requested):
Just some thoughts I had about avoiding the creation of the array in both cases.

I'm thinking it would be more efficient when the user program breaks out of the loop:

```ruby
h = {...} # some huge hash
h.each_key do |key|
  # some work...
  break if key == :foo
end
```

src/hash_value.cpp
seven1m: I should say, you can obviously do that in C++ too, but it was just easier for me to illustrate in Ruby quickly. :wink: Really whichever language you want to implement it in...

richardboehme:
Yeah you're absolutely right. I'm going to change that!

seven1m (approved):

PR #151, opened 2021-09-15
Implement Hash#except

richardboehme:
I do not know if there is an easier way of copying the hash. I cannot use the constructor like `new HashValue { *this }`, because it expects the new Hash to be of class Hash (and not of a sub class). Is there an easier way to do it?

### DIFF
--- a/include/natalie/hash_value.hpp
+++ b/include/natalie/hash_value.hpp
@@ -132,6 +132,7 @@ public:
     bool eq(Env *, ValuePtr, SymbolValue *);
     bool eq(Env *, ValuePtr);
     bool eql(Env *, ValuePtr);
+    ValuePtr except(Env *, size_t, ValuePtr *);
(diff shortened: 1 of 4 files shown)


seven1m:
I'm surprised that `new HashValue { env, *this }` didn't work, but I don't fully understand what you mean about the class.

Your code essentially does the same thing anyway, so I think this is fine...

seven1m (approved):

richardboehme:
> I'm surprised that `new HashValue { env, *this }` didn't work, but I don't fully understand what you mean about the class.
> 
> Your code essentially does the same thing anyway, so I think this is fine...

I think when using `new HashValue { env, *this }` `Value` will be initialized using the current hash value with it's class type. I'm talking specifically about the spec `spec/core/hash/except_spec.rb:23`. Basically `Hash#except` must always return a class of type hash (not a subclass or anything like that).

PR #152, opened 2021-09-15
Implement Hash#fetch and Hash#fetch_values

richardboehme:
A question that turned up during development of this: Do we have a generic way of printing a warning to the user? In *.rb files we just print to $stderr, so I tried to mimic this behavior. Maybe we could add a simpler helper or something?

### DIFF
--- a/include/natalie/env.hpp
+++ b/include/natalie/env.hpp
@@ -65,6 +65,14 @@ public:
         raise(class_name, message);
     }
 
+    void warn(const String *);
+
+    template <typename... Args>
+    void warn(const char *format, Args... args) {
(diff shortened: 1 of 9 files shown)


seven1m (approved):
I think what you have is fine! If you want to make a more convenient function/method for it (especially if you are needing it in more than a few places), then by all means go for it! :smile:

richardboehme:
> I think what you have is fine! If you want to make a more convenient function/method for it (especially if you are needing it in more than a few places), then by all means go for it! 😄

I've just quickly moved it over to the Env class. We will probably need it in the future (for example when implementing `Kernel#warn`)! 😄

PR #167, opened 2021-09-19
Fix behavior of Hash#value? if value is falsey

richardboehme:
Noticed this bug while working on `Hash#slice`:

Previously we double-negated the returned key from `Hash#key`. However,
this method returns `nil` when no key is found. As hash keys can also
be `nil`, we cannot differentiate between a found key that's `nil` and
no found key.

### DIFF
--- a/test/natalie/hash_test.rb
+++ b/test/natalie/hash_test.rb
@@ -282,4 +282,11 @@ describe 'hash' do
       h.values_at(:a, :b, :c).should == [1, 2, 3]
     end
   end
+
+  describe '#value?' do
+    it "work with falsey values" do
+      h = { nil => :a }
+      h.value?(:a).should == true
+    end
+  end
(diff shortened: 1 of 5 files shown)


test/natalie/hash_test.rb
```diff
@@ -282,4 +282,11 @@ def hash
       h.values_at(:a, :b, :c).should == [1, 2, 3]
     end
   end
+
+  describe '#value?' do
+    fit "work with falsey values" do
```
seven1m, line 287: focused test 😉

seven1m (changes requested):

test/natalie/hash_test.rb
richardboehme: oopsie

seven1m (approved):

PR #169, opened 2021-09-19
Implement Hash#rehash and Hash#compare_by_identity(?)

richardboehme:
Instead of using the `hash` method I'm using the `object_id` when `Hash#compare_by_identity` is called as the hash for the hash map. This seems to be working fine....

I'm not sure about the `rehash` implementation. Maybe we can make this more elegant?

### DIFF
--- a/src/hash_value.cpp
+++ b/src/hash_value.cpp
@@ -21,11 +21,34 @@ bool HashValue::compare(const void *a, const void *b, void *env) {
     return a_p->key.send((Env *)env, SymbolValue::intern("eql?"), { b_p->key })->is_truthy();
 }
 
+ValuePtr HashValue::compare_by_identity(Env *env) {
+    assert_not_frozen(env);
+    this->m_is_comparing_by_identity = true;
+    this->rehash(env);
+    return this;
+}
+
+ValuePtr HashValue::is_comparing_by_identity() {
+    if (m_is_comparing_by_identity) {
+        return TrueValue::the();
+    } else {
+        return FalseValue::the();
+    }
+}
+
 ValuePtr HashValue::get(Env *env, ValuePtr key) {
     Key key_container;
     key_container.key = key;
-    key_container.hash = key.send(env, SymbolValue::intern("hash"))->as_integer()->to_nat_int_t();
+    key_container.hash = generate_key_hash(env, key);
     return m_hashmap.get(&key_container, env);
+}
+
+nat_int_t HashValue::generate_key_hash(Env *env, ValuePtr key) {
+    if (m_is_comparing_by_identity) {
+        return TM::Hashmap<void *>::hash_ptr(key.value());
+    } else {
+        return key.send(env, SymbolValue::intern("hash"))->as_integer()->to_nat_int_t();
+    }
 }
 
 ValuePtr HashValue::get_default(Env *env, ValuePtr key) {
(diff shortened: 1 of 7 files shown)


richardboehme:
Mh I don't really know why the release build is failing in this one rehash spec case. It seems to be random but also kind of consistent sometimes? Could it be a GC thing?

src/hash_value.cpp
```diff
@@ -21,13 +21,36 @@ bool HashValue::compare(const void *a, const void *b, void *env) {
...
     Key key_container;
     key_container.key = key;
-    key_container.hash = key.send(env, SymbolValue::intern("hash"))->as_integer()->to_nat_int_t();
+    key_container.hash = generate_key_hash(env, key);
     return m_hashmap.get(&key_container, env);
 }
 
+nat_int_t HashValue::generate_key_hash(Env *env, ValuePtr key) {
+    if (m_is_comparing_by_identity) {
+        return key->object_id();
```
seven1m, line 48: This won't return a value that is evenly distributed, so we'll have to do something like:

```suggestion
        return TM::Hashmap<void*>::hash_ptr(key);
```

seven1m (changes requested):
I'm not sure about the release build -- I'd have to break out gdb and see what's going on. Sounds like a fun debugging session. :wink: 

Try making my suggested change and see if it goes away for now... :grimacing:

src/hash_value.cpp
richardboehme: I'm not sure if it's the same but the suggestion you did would not compile.. I changed it to 
```cpp
return TM::Hashmap<void*>::hash_ptr(key.value());
```
Is this fine?

I guess it won't change the behavior of the rehash spec that is failing though, because the spec does not use `compare_by_identity`.

seven1m: Yes, that is correct. I typed it without checking.

I'll take a look at the failing spec with a debugger a little later!

richardboehme: Great! I definitely have to get into this debugging thing myself at some point...

seven1m (approved):
@richardboehme you might take a look at the commit I added. It makes the specs pass, but I wonder if I should also copy of `m_default_value` etc. too... :thinking:

richardboehme:
> @richardboehme you might take a look at the commit I added. It makes the specs pass, but I wonder if I should also copy over `m_default_value` etc. too... 🤔

To be honest, I'm not really sure if I even understand how your commit fixes it... It creates a new HashValue and assigns it to the current one. Beforehand we only had this one HashValue, so it should not be collected by the GC, right?. I'm sorry if this looks dumb but my knowledge of C++ and our GC is very small yet, so could you elaborate further how your change fixes it?

seven1m:
@richardboehme sure thing, let me try to explain a bit better than I did...

Your code for HashValue#rehash looked like this (with annotations added by me):

```c++
ValuePtr HashValue::rehash(Env *env) {
    assert_not_frozen(env);

    auto old_hashmap = TM::Hashmap<Key *, Value *> { m_hashmap }; // creates a *copy* of the m_hashmap
    clear(env); // m_hashmap no longer points to any Values

    // <------- For simplicity, let's say the GC chooses to run right *here*.

    for (auto pair : old_hashmap) {
        put(env, pair.first->key, pair.second);
    }

    return this;
}
```

Right after you run `clear(env)`, the GC happens to run. It looks through all the of the things accessible from the stack, heap, registers, etc, and it finds *this* HashValue, the one we are operating on, but this HashValue does not point to any items! It is empty at this particular moment. :grimacing: 

Now, you say, "but the `old_hashmap` object still points to all the items!". Yes, it does, but the GC does not know how to walk TM::Hashmaps -- it only knows how to walk classes that descend from `Cell`.

So, if the GC runs at that particular moment (and by chance, that's what happened in the "release" build and caused the test to segfault), then it will see that a bunch of items need to be cleaned up. But we're still using them!

The fix is to use a "real" `HashValue` to store the items temporarily. That way, if the GC runs, it still "sees" the items are accessible from running code and it won't clean them up.

Finally, in order to overwrite a HashValue with another (temporary) HashValue, I implemented the assignment operator which "moves" the `m_hashmap` from one (the temporary `old_hashmap` one) over to the actual HashValue we care about.

I hope that makes sense!!

richardboehme:
That makes so much sense! Thank you very much for explaining this to me!

PR #171, opened 2021-09-19
Implement all operators (<, <=, >, >=) for Hash

richardboehme:


src/hash_value.cpp
```diff
@@ -371,6 +371,56 @@ bool HashValue::eql(Env *env, ValuePtr other_value) {
...
+    return true;
+}
+
+bool HashValue::lt(Env *env, ValuePtr other) {
+    if (!other->is_hash() && other->respond_to_method(env, SymbolValue::intern("to_hash")))
+        other = other->send(env, SymbolValue::intern("to_hash"));
+
+    other->assert_type(env, Value::Type::Hash, "Hash");
+
+    return lte(env, other) && other->as_hash()->size() != size();
```
seven1m, line 421: Can you explain how this works? At first glance, it seems wrong to me...

One of the specs says `it 'returns true when receiver is a subset of argument'` and this doesn't seem to check for that?

richardboehme: One hash a is "less than" another hash b if each element of a is in b (checked by `lte`) and they are not equal. The trick is instead of checking for equality we just test if the size of a and b are different because if they are and all elements of a are in b (which we checked by `lte`), they must be equal. Does this make any sense? 😅 

Do you have an idea of making the code more understandable?

seven1m: No I think I get it. Works for me!

PR #176, opened 2021-09-22
Array prepend and unshift

TheGrizzlyDev:
Implement Array::unshift and add prepend as an alias as per #39

### DIFF
--- a/spec/core/array/shared/unshift.rb
+++ b/spec/core/array/shared/unshift.rb
@@ -1,0 +1,46 @@
+describe :array_unshift, shared: true do
+  it "prepends object to the original array" do
+    a = [1, 2, 3]
+    a.send(@method, "a").should equal(a)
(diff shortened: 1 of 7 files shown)


spec/core/array/shared/unshift.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  it "properly handles recursive arrays" do
+    empty = ArraySpecs.empty_recursive_array
+    empty.send(@method, :new).should == [:new, empty]
+
+    array = ArraySpecs.recursive_array
+    array.send(@method, :new)
+    array[0..5].should == [:new, 1, 'two', 3.0, array, array]
+  end
```
seven1m, line 36: This just worked!? NICE

seven1m (approved):
Looks good assuming the build passes! I'm heading to work :)

PR #178, opened 2021-09-22
Array filter select

TheGrizzlyDev:
add specs for filter and select of Array (#39) and implement their in-place counterparts, making sure the non in place version is backed by those and runs on a copy of the original array.


seven1m:
@TheGrizzlyDev sorry to snipe your PR, but I had a different (hopefully better) idea for doing the select in place. Rather than creating a vector of indexes to remove and then removing them one-at-a-time (which has to shift the vector entries down one-at-a-time, and is inefficient), I figured we should make an assignment operator for ArrayValue and effectively _replace_ its vector with a newly-built one.

TheGrizzlyDev:
> @TheGrizzlyDev sorry to snipe your PR, but I had a different (hopefully better) idea for doing the select in place. Rather than creating a vector of indexes to remove and then removing them one-at-a-time (which has to shift the vector entries down one-at-a-time, and is inefficient), I figured we should make an assignment operator for ArrayValue and effectively _replace_ its vector with a newly-built one.

Hey, no need to be sorry, it's cool! The only question I'd have, not knowing rube (and neither c++ to be honest ahahahha) would be if there are similar flags to frozen in array and if so how should we behave? Frozen itself would not be a problem by the way, so no need to check that :)

seven1m:
Yeah I believe I took care of copying flags, ivars, etc over by calling `Value::operator=()`.

PR #185, opened 2021-09-23
Implement Hash#deconstruct_keys

richardboehme:
The method does not seem to have any use and the spec only runs for
rubies with version 2.7, so I'm not sure what the purpose is. However,
here it is :^)


seven1m:
Here's a bit about it: https://rubyreferences.github.io/rubyref/language/pattern-matching.html#matching-non-primitive-objects-deconstructkeys-and-deconstruct

Your implementation here looks correct. 😜

richardboehme:
Yeah it's weird that the spec isn't even running when on version 3 😅

PR #186, opened 2021-09-23
Make Hash#initialize spec compliant

richardboehme:
I did not see another way of defining a private method, then defining it and making it private afterwards. Is there better way to do this?


seven1m:
This is what I was thinking you would do! <3

Issue #188, opened 2021-09-24
`Array#hash` producing too many collisions


richardboehme:
I'm currently tinkering around with `Hash#hash` and noticed this bug in `Array#hash`:
```ruby
[2, 2].hash == [1, 1].hash
# => true
```

`[2, 2].hash` will return 100019 which is the starting point of the hash calculation, because `(2.hash * 207269) ^ (2.hash * 207269)` will eliminate each other. My first thought of fixing this is multiplying with a random prime number instead of a fixed one. Would this work?

seven1m:
Whoops! @TheGrizzlyDev and I worked on that, but that mistake is mine. The recursive array specs are mind bending, but I feel confident adding a random number won't fix it. 😕

I will try to take another look at this soon.

TheGrizzlyDev:
> I feel confident adding a random number won't fix it. confused
I feel like we've been here already ahhahaha. A similar but less naive approach would be to just shift at each iteration :)

seven1m:
@TheGrizzlyDev you're a genius. (That worked.)

TheGrizzlyDev:
> @TheGrizzlyDev you're a genius. (That worked.)

Ahahahaha thanks! I've spent enough time on it to think in hashes

richardboehme:
Thanks for fixing this so fast! 😄 Sadly, I think this is not enough yet. Firstly it does not work with nested arrays:

```ruby
[[1, 2], [3, 2]].hash == [[1, 3], [2, 3]]
# => true
```
The two 2s and the 3s are eliminating each other.

Secondly, as nat_int_t is "only" 64 bit, it will not work if we shift more than 64 times (when the array is bigger than 64):

```ruby
def generate_array(i)
  [].tap do |out|
    66.times { out << i }
  end
end

generate_array(2).hash == generate_array(3).hash
# => true
```

seven1m:
Bummer. Well I'll have to take another stab at it later. If you have ideas, by all means, go for it. 😉

TheGrizzlyDev:
this keeps getting better and better ahahhahah. How about we take a page from DES/AES which have the same issue with using multiple rounds and having just a single key (in this case our prime number)?
We could:
- select a big enough prime number (or any number big enough that if taken in smaller bits would still have a somewhat reasonable distribution of 0s and 1s and at least with more bits than nat_int_t)
- for each item we get the hash of, we xor it with a prime number AND then xor it with a subkey of the aforementioned number of 64 bits in length
[...]

PR #190, opened 2021-09-25
make Array::hash more reliable on arrays containing only duplicate elements

TheGrizzlyDev:
This is a follow-up to the latest issue in #188, the best issue ever on this project :^). @seven1m @richardboehme please have a look and let me know if you can break this one too ;)

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -1254,7 +1254,8 @@ ValuePtr ArrayValue::hash(Env *env) {
     return guard.run([&](bool is_recursive) {
         if (is_recursive)
             return ValuePtr { NilValue::the() };
-        nat_int_t hash = 100019; // just a prime number :-)
+
+        HashBuilder hash {};
         auto hash_method = SymbolValue::intern("hash");
         auto to_int = SymbolValue::intern("to_int");
 
@@ -1274,12 +1275,10 @@ ValuePtr ArrayValue::hash(Env *env) {
             if (!item_hash->is_integer() && item_hash->respond_to(env, to_int))
                 item_hash = item_hash->send(env, to_int);
 
-            nat_int_t h = item_hash->as_integer()->to_nat_int_t();
-
-            hash += (hash << 5) ^ h;
+            hash.append(item_hash->as_integer()->to_nat_int_t());
         }
 
-        return ValuePtr::integer(hash);
+        return ValuePtr::integer(hash.digest());
     });
 }
 
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -1250,8 +1250,6 @@ ValuePtr ArrayValue::rassoc(Env *env, ValuePtr needle) {
 }
 
 ValuePtr ArrayValue::hash(Env *env) {
-    constexpr const int PRIME_COUNT = 10;
-    nat_int_t primes[PRIME_COUNT] = { 55108187, 37599817, 22555241, 54071879, 15925463, 35080337, 38524007, 19435781, 97744909, 32643563 };
     TM::RecursionGuard guard { this };
     return guard.run([&](bool is_recursive) {
         if (is_recursive)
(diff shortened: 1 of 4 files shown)


src/array_value.cpp
```diff
@@ -1277,8 +1275,8 @@ ValuePtr ArrayValue::hash(Env *env) {
                 item_hash = item_hash->send(env, to_int);
 
             nat_int_t h = item_hash->as_integer()->to_nat_int_t();
-            hash ^= i;
-            hash ^= h * primes[i % PRIME_COUNT]; // another prime number to prevent bias toward zero
+
+            hash += (hash << 5) ^ h;
```
seven1m, line 1279: How did you come up with `5` here?

seven1m (commented):
It would be nice to add in a test case or two to prove this is superior...

src/array_value.cpp
TheGrizzlyDev: any number that would not the number of bits was not divisible by in order to avoid periodicity issues (or really, just delay them by 5 * nat_int_t_bits) that was also not too high and would not completely invalidate the effect of the left-most bits of the previous hash (yeah, sums puts those back, but the unshifted part still had more influence). Another candidate by that train of thought was 3, but then again, 5 is almost twice as much effective in terms of periodicity :) 

Both are conjectures though, I really need to verify how good is the distribution of this :^)

seven1m: OK that makes sense!

TheGrizzlyDev:
> It would be nice to add in a test case or two to prove this is superior...

that's coming :D that though begs the question, this would be a quite slow test, should it be part of the main test suite or should we add it separately?

seven1m:
@TheGrizzlyDev I'm not saying you should include hundreds of fuzzing style tests -- just one or two cases (if you can find them) where this algorithm produces a better result. Would that be possible?

TheGrizzlyDev:
> @TheGrizzlyDev I'm not saying you should include hundreds of fuzzing style tests -- just one or two cases (if you can find them) where this algorithm produces a better result. Would that be possible?

oh for that I could just use the test suite I had in the last comment of the issue, that runs fast, would that work?

PS: if we need something more that's also cool, as I plan on doing a similar work for Hash and maybe refactor all of this logic into an HashBuilder that can just be fed new chuncks and spit out a single hash at the end. Any suggestion?

seven1m:
Yeah that'd be fine.

TheGrizzlyDev:
> Yeah that'd be fine.

done and took the chance to pull in the small refactoring I was talking about, but we will probably have to iterate over hash and test it some more I guess :)

test/natalie/array_test.rb
```diff
@@ -44,6 +44,20 @@
...
+      (0..100).map do |length| 
+        current_hash = (0..length).to_a.map { |x| 0 }.hash 
+        if found_hashes.find_index current_hash
+          fail
+        else
+          found_hashes << current_hash
+        end 
+      end
+    end
+  end
```
seven1m, line 59: Can you move this down with the other `describe '#hash'` block?

TheGrizzlyDev: Oooops sorry, will do!

TheGrizzlyDev: fixed, thanks ;)

test/natalie/array_test.rb
```diff
@@ -44,6 +44,20 @@
...
+  describe 'hash' do
+    it 'does not collide when array contains duplicate elements' do
+      found_hashes = []
+      (0..100).map do |length| 
+        current_hash = (0..length).to_a.map { |x| 0 }.hash 
+        if found_hashes.find_index current_hash
+          fail
+        else
+          found_hashes << current_hash
+        end
```
seven1m: `fail` is an alias for `raise`, so I think to fail the spec properly we should use an expectation:

```suggestion
        found_hashes.find_index(current_hash).should be_nil
        found_hashes << current_hash
```

PR #191, opened 2021-09-25
Array repeated combination

TheGrizzlyDev:
Another one for #39 . The current implementation is somewhat naive, even aside from the fact I'd try to avoid recursion altogether, there is still the issue of how this just returns a non-lazy iterator because I do not know how to calculate all the possible combinations.

### DIFF
--- a/spec/core/array/repeated_combination_spec.rb
+++ b/spec/core/array/repeated_combination_spec.rb
@@ -1,0 +1,84 @@
+require_relative '../../spec_helper'
+
+describe "Array#repeated_combination" do
+  before :each do
+    @array = [10, 11, 12]
+  end
+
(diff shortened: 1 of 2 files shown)


seven1m (approved):
Works for me. We can optimize the enumerator later if needed. Though I'm confident I've never seen this method used in production code. 😛

TheGrizzlyDev:
> Works for me. We can optimize the enumerator later if needed. Though I'm confident I've never seen this method used in production code. 😛

I actually plan on using it to generate test cases for Array::hash ahahhahaha. I want to make an array of a large enough amount of primitive values and then get various lengths of combinations and us it to see how well distributed hashes are and create a decent enough test-bench! Here you go, a production use-case ahahahah

PR #193, opened 2021-09-25
Add specs for Array#take and Array#take_while

borisromanov:
#39 

I believe this just uses the `Enumerable#take` and `Enumerable#take_while`, methods. Is that okay to do here or would it be preferable for array_value.cpp to have take and take_while methods defined?


seven1m:
This is fine for now as it gets specs passing. We will need to, at some point, loop back and figure out how exactly Array has so many of its own methods, if they are just aliases, etc. But I'm not too worried about it right now.

PR #194, opened 2021-09-26
Implement test to check that our collision rate for Array::hash is comparable or lower MRI's collision rate

### LINKED ISSUE #188: `Array#hash` producing too many collisions

TheGrizzlyDev:
The test is still quite naive and only generates around 50k test cases and those are all very short arrays, so it is quite biased. Regardless it is still nice to have and see that we actually outperform MRI in this scenario (0.9190165737176968% vs 11.5%), which implies that this is 'good enough' for ruby and in theory won't need any further improvement in the foreseeable future. With this we should be able to close #188

### DIFF
--- a/test/natalie/array_test.rb
+++ b/test/natalie/array_test.rb
@@ -1822,10 +1822,42 @@ describe 'array' do
 
     it 'does not collide when array contains duplicate elements' do
       found_hashes = []
-      (0..100).map do |length| 
+      (0..100).each do |length| 
         current_hash = (0..length).to_a.map { |x| 0 }.hash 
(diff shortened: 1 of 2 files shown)


seven1m:
COOL! :sunglasses:

PR #198, opened 2021-09-26
make Array::clone spec compliant

TheGrizzlyDev:
Added spec for Array::clone from #39 and added the 'copy singleton methods' behaviour MRI has. Since this does not look restricted to just Array, I've done that in Kernel itself and said behaviour does not exist with dup, so it's just for clone :)


seven1m (approved):
Nice!

PR #205, opened 2021-09-28
Implement Random

richardboehme:
This PR adds some basics for the Random class.

Please have a closer look on this, as I may missed important parts of adding a new value class. The following issues are known:

* when passing a float range, exclusiveness is not guaranteed
* we cannot pass ranges of objects that respond to +/- yet, because I'm not sure what we are supposed to do with those objects

Also I added three new allocators because a RandomValue is > 5000 Bytes. Not sure if this is good...

### DIFF
--- a/include/natalie/global_env.hpp
+++ b/include/natalie/global_env.hpp
@@ -47,6 +47,9 @@ public:
     ClassValue *Object() { return m_Object; }
     void set_Object(ClassValue *Object) { m_Object = Object; }
 
+    ClassValue *Random() { return m_Random; }
+    void set_Random(ClassValue *Random) { m_Random = Random; }
+
     ClassValue *Regexp() { return m_Regexp; }
     void set_Regexp(ClassValue *Regexp) { m_Regexp = Regexp; }
 
@@ -88,6 +91,7 @@ private:
     ClassValue *m_Integer { nullptr };
     ClassValue *m_Module { nullptr };
     ClassValue *m_Object { nullptr };
+    ClassValue *m_Random { nullptr };
     ClassValue *m_Regexp { nullptr };
     ClassValue *m_String { nullptr };
     ClassValue *m_Symbol { nullptr };
--- a/include/natalie/random_value.hpp
+++ b/include/natalie/random_value.hpp
@@ -1,0 +1,52 @@
+#pragma once
+
+#include <utility>
+
+#include "natalie/class_value.hpp"
+#include "natalie/forward.hpp"
+#include "natalie/global_env.hpp"
+#include "natalie/integer_value.hpp"
+#include "natalie/value.hpp"
+#include <random>
+
+namespace Natalie {
+
+using namespace TM;
+
+class RandomValue : public Value {
+public:
+    RandomValue(ClassValue *klass)
+        : Value { Value::Type::Random, klass } { }
+
+    RandomValue()
+        : Value { Value::Type::Random, GlobalEnv::the()->Random() } { }
+
+    virtual void visit_children(Visitor &visitor) override final {
+        Value::visit_children(visitor);
+    }
+
+    ValuePtr initialize(Env *, ValuePtr);
+
+    ValuePtr rand(Env *, ValuePtr);
+    ValuePtr seed() { return new IntegerValue { m_seed }; }
+
+    virtual void gc_inspect(char *buf, size_t len) const override {
+        snprintf(buf, len, "<Random %p seed=%lld>", this, m_seed);
+    }
+
+private:
+    nat_int_t m_seed;
+    std::mt19937 m_generator;
+
+    ValuePtr generate_random(double min, double max) {
+        std::uniform_real_distribution<double> random_number(min, max);
+        return new FloatValue { random_number(m_generator) };
+    }
+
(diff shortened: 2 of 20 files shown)


include/natalie/random_value.hpp
```diff
@@ -0,0 +1,52 @@
...
+    ValuePtr rand(Env *, ValuePtr);
+    ValuePtr seed() { return new IntegerValue { m_seed }; }
+
+    virtual void gc_inspect(char *buf, size_t len) const override {
+        snprintf(buf, len, "<Random %p seed=%lld>", this, m_seed);
+    }
+
+private:
+    nat_int_t m_seed;
+    std::mt19937 m_generator;
```
seven1m, line 39: @richardboehme let's change this into a pointer, allocate it in the constructor, and deallocate it in a (to-be-added) destructor. That way we don't need to have the GC overhead of those huge Cell size Allocators.

```suggestion
    std::mt19937 *m_generator;
```

I tested dynamically allocating it in a simple cpp program and it seemed to work:

```cpp
#include <random>   
#include <stdio.h>                                                     
                                                                       
int main() {                                                           
    auto *mt_rand = new std::mt19937(time(0));                 
    printf("rand = %lu\n", (*mt_rand)());                                                                                                                                                                          
    return 0;                                                                                                                                                                                                      
}
```

include/natalie/random_value.hpp
```diff
@@ -0,0 +1,52 @@
...
+public:
+    RandomValue(ClassValue *klass)
+        : Value { Value::Type::Random, klass } { }
+
+    RandomValue()
+        : Value { Value::Type::Random, GlobalEnv::the()->Random() } { }
+
+    virtual void visit_children(Visitor &visitor) override final {
+        Value::visit_children(visitor);
+    }
```
seven1m, line 26: You can remove this virtual method since it's just calling the superclass' visit_children.

include/natalie/global_env.hpp
```diff
@@ -88,6 +91,7 @@ class GlobalEnv : public Cell {
     ClassValue *m_Integer { nullptr };
     ClassValue *m_Module { nullptr };
     ClassValue *m_Object { nullptr };
+    ClassValue *m_Random { nullptr };
     ClassValue *m_Regexp { nullptr };
     ClassValue *m_String { nullptr };
     ClassValue *m_Symbol { nullptr };
```
seven1m, line 97: Also you need to visit the Random class from `GlobalEnv::visit_children`.

seven1m (changes requested):
This is really cool! Just a couple small changes.

Getting the random number generator out of the Random class will let us have smaller allocators and reduce some load on the GC.

include/natalie/random_value.hpp
richardboehme: I thought about the general layout of the class. Do you think it makes sense, to store the seed (`m_seed`) as an IntegerValue directly? This way we would not need to create a new Value each time someone calls `Random#seed`. Of course we would need to collect it during GC then...

include/natalie/random_value.hpp
richardboehme: That's a great idea! 👍

include/natalie/random_value.hpp
seven1m: Integers are free if you use `ValuePtr::integer`, since they're encoded in the smart pointer. So either way is fine.

include/natalie/random_value.hpp
```diff
@@ -21,6 +21,8 @@ class RandomValue : public Value {
     RandomValue()
         : Value { Value::Type::Random, GlobalEnv::the()->Random() } { }
 
+    ~RandomValue() { free(m_generator); }
```
seven1m, line 24: Let's use `delete m_generator` here to match the use of `new` in the initializer.

include/natalie/random_value.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+    virtual void visit_children(Visitor &visitor) override final {
+        Value::visit_children(visitor);
+    }
+
+    ValuePtr initialize(Env *, ValuePtr);
+
+    ValuePtr rand(Env *, ValuePtr);
+    ValuePtr seed() { return new IntegerValue { m_seed }; }
```
seven1m, line 33:
```suggestion
    ValuePtr seed() { return ValuePtr::integer(m_seed); }
```

`ValuePtr::integer` doesn't cause an allocation and can save some memory in certain optimized circumstances.

include/natalie/random_value.hpp
```diff
@@ -0,0 +1,56 @@
...
+    ValuePtr generate_random(double min, double max) {
+        assert(m_generator);
+        std::uniform_real_distribution<double> random_number(min, max);
+        return new FloatValue { random_number(*m_generator) };
+    }
+
+    ValuePtr generate_random(nat_int_t min, nat_int_t max) {
+        assert(m_generator);
+        std::uniform_int_distribution<nat_int_t> random_number(min, max);
+        return new IntegerValue { random_number(*m_generator) };
```
seven1m, line 52:
```suggestion
        return new ValuePtr::integer(random_number(*m_generator));
```

seven1m (approved):

PR #208, opened 2021-09-30
make Array initialize spec compliant

TheGrizzlyDev:
from #39

### DIFF
--- a/spec/core/array/initialize_spec.rb
+++ b/spec/core/array/initialize_spec.rb
@@ -1,0 +1,156 @@
+require_relative '../../spec_helper'
+require_relative 'fixtures/classes'
+
+describe "Array#initialize" do
+  before :each do
+    ScratchPad.clear
+  end
+
+  it "is private" do
+    Array.should have_private_instance_method("initialize")
+  end
+
+  it "is called on subclasses" do
+    b = ArraySpecs::SubArray.new :size_or_array, :obj
+
+    b.should == []
+    ScratchPad.recorded.should == [:size_or_array, :obj]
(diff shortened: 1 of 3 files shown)


spec/core/array/initialize_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+require_relative 'fixtures/classes'
+
+describe "Array#initialize" do
+  before :each do
+    ScratchPad.clear
+  end
+
+  #FIXME add back once we support have_private_instance_method
+  xit "is private" do
+    Array.should have_private_instance_method("initialize")
```
richardboehme, line 11: I think we can still make `Array#initialize` private by using the new `:visibility` keyword on the bindings :^) This would not make the spec pass, but as soon as we support #private_instance_method? it will just pass

TheGrizzlyDev: Fair point, I am in general not a fan of writing code without tests or that cannot be tested, but in this case it shouldn't hurt :^) Thanks!

richardboehme: Yeah I thought about implementing `#private_instance_method?` for the `Hash#initialize` spec, but I decided to hack around it in the spec (https://github.com/seven1m/natalie/blob/master/spec/core/hash/initialize_spec.rb#L7).

TheGrizzlyDev: Fixed it, but this really makes me feel like implementing this missing bit in module, I will probably create an issue to make Module spec compliant :)

PR #209, opened 2021-09-30
make Array pop spec compliant

TheGrizzlyDev:
#39


src/array_value.cpp
```diff
@@ -951,11 +951,32 @@ void ArrayValue::push_splat(Env *env, ValuePtr val) {
...
+            count = count.send(env, to_int);
+
+        count->assert_type(env, ValueType::Integer, "Integer");
+        auto c = count->as_integer()->to_nat_int_t();
+
+        if (c < 0)
+            env->raise("ArgumentError", "negative array size");
+
+        if (c > (nat_int_t)size())
+            c = (nat_int_t)size();
```
seven1m, line 970: Any time I coerce a `size_t` to `nat_int_t` I add an `assert(size() <= NAT_INT_MAX);` just prior. I suppose that's overkill since an array of that size would likely consume all the memory in my machine. But still, seems like a good habit to make. Thoughts?

TheGrizzlyDev: Makes sense, will fix it if I start cleaning up Array or when I implement the next method :) Thanks!

PR #212, opened 2021-10-01
Try to fix keyword argument processing

richardboehme:
Fix "FIXMEs" of natalie's method test. This contains fixes for the following scenarios:

When passed a Hash to a method that has default arguments, we would assign the hash to the default argument even if it has keyword arguments that should be matched to the hash.

```ruby
def foo(a, b = 'b', **c)
  [a, b, c]
end
foo(1, c: 'c', d: 'd').should == [1, 'b', { c: 'c', d: 'd' }
```

The second scenario is that keyword arguments that were successfully extracted from the keyword hash, should not be present in the "rest" (when using double-splats).

```ruby
def foo(a: 'a', **b)
  [a, b]
end
foo(a: 1, b: 2, c: 3).should == [1, { b: 2, c: 3 }] 
```

### DIFF
--- a/src/natalie.cpp
+++ b/src/natalie.cpp
@@ -281,13 +281,13 @@ static ValuePtr splat_value(Env *env, ValuePtr value, size_t index, size_t offse
     return splat;
 }
 
-ValuePtr arg_value_by_path(Env *env, ValuePtr value, ValuePtr default_value, bool splat, int total_count, int default_count, bool defaults_on_right, int offset_from_end, size_t path_size, ...) {
+ValuePtr arg_value_by_path(Env *env, ValuePtr value, ValuePtr default_value, bool splat, bool has_kwargs, int total_count, int default_count, bool defaults_on_right, int offset_from_end, size_t path_size, ...) {
     va_list args;
     va_start(args, path_size);
     bool has_default = !!default_value;
     if (!default_value) default_value = NilValue::the();
     bool defaults_on_left = !defaults_on_right;
-    int required_count = total_count - default_count;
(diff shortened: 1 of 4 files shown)


src/natalie.cpp
```diff
@@ -430,7 +435,7 @@ HashValue *kwarg_hash(ArrayValue *args) {
 
 ValuePtr kwarg_value_by_name(Env *env, ArrayValue *args, const char *name, ValuePtr default_value) {
     auto hash = kwarg_hash(args);
-    ValuePtr value = hash->as_hash()->get(env, SymbolValue::intern(name));
+    ValuePtr value = hash->as_hash()->remove(env, SymbolValue::intern(name));
```
seven1m, line 438: Nice!

test/natalie/method_test.rb
```diff
@@ -442,9 +442,8 @@ def method_with_kwargs11(a:, **b)
     method_with_kwargs9('a', b: 'b').should == ['a', 'b']
     method_with_kwargs10(b: 'b').should == [1]
 
-    # FIXME: failing:
-    #method_with_kwargs9(b: 'b').should == [1, 'b']
-    #method_with_kwargs11(a: 'a', b: 'b').should == ['a', { b: 'b' }]
+    method_with_kwargs9(b: 'b').should == [1, 'b']
+    method_with_kwargs11(a: 'a', b: 'b').should == ['a', { b: 'b' }]
```
seven1m, line 446: Can you sort these with the list above 🙏

seven1m (approved):
Nice work!

PR #216, opened 2021-10-03
Add initial support for Bignums

richardboehme:
As discussed on Natalie's discord server, we are adding support for bignums by using a library that does most of the work for us. The current implementation stores a flag to know if the IntegerValue is a bignum or a fixnum. Not sure if there is a better implementation approach. 

To test the implementation, I added support for Integer#+ and made it spec compliant.


seven1m:
This is awesome!! :tada: Going to look at it in more detail a bit later...

seven1m:
I really like your work here and it's exciting that this actually works as well as it does. The way you integrated it into IntegerValue feels really natural to me. Though, I think we should consider subclassing IntegerValue to make BignumValue. This would only be an implementation detail and wouldn't be exposed inside Ruby, but would allow us to save the space taken up by Bignum when it's not needed.

But the big blocker on this for me is the use of std::string. You'll notice that Natalie does not use std::string -- we have our own String class. The reason is compilation time. And I don't mean the time it takes to run `rake build`. I mean the time it takes gcc/clang to compile and link our generated code when we run `bin/natalie script.rb`.

I have found that any time I include std::string, std::unordered_map, and some others I cannot think of right now, compilation time takes a nose dive. Here is a quick example I just ran:

**master branch:**

```sh
time bin/natalie examples/boardslam.rb 3 5 1
...snip...
bin/natalie examples/boardslam.rb 3 5 1  3.42s user 0.23s system 99% cpu 3.652 total
```

**bignums branch:**

```sh
time bin/natalie examples/boardslam.rb 3 5 1
...snip...
bin/natalie examples/boardslam.rb 3 5 1  4.06s user 0.26s system 99% cpu 4.323 total
```

So, as it is, I don't think we can merge this. :disappointed: 

I see a couple of options for moving forward with Bignums:
1. Find another library that does not use std::string. I was a bit surprised to see this one uses std::string to store the value. The other libraries I have seen use tree of nodes, which seems more efficient.
2. Tweak this library so it uses our own Natalie::String.
3. Make our own bignum library from scratch.

Again, your work here is excellent, and I really really want to get Bignum support merged. But not at the expense of compilation time.

Let me know your thoughts!

richardboehme:
Thanks for the feedback. I also noticed, that they are using a string to store the bignum which seemed odd. But I only thought about it impacting runtime performance for Bignums which was not too much of a problem for now. 

I guess the best solution would be to switch the libraries. I found some more which may be working better. I'll work on integrating them.

I also thought about separating Bignum/Fixnum but I'm not sure how we dynamically change between them. Let's say we subclass IntegerValue as BignumValue. If we have a bignum and substract from it, so that it would become a fixnum. How do we "change" class types without changing the memory address?
Mh... Maybe we do not even need that. It seems like ruby is allocating a new object (also on modifying operators?):

```ruby
a = 1000000000000000000000000000000000000
a.object_id
# => 360
a -= 1
a.object_id
# => 380
```

So maybe it's fine like you said?

seven1m:
>I guess the best solution would be to switch the libraries. I found some more which may be working better. I'll work on integrating them.

🎉 

>Maybe we do not even need that. It seems like ruby is allocating a new object (also on modifying operators?):

Yeah, I don't believe integers are mutable ever. Even `a -= 1` is just `a = a - 1` in the AST:

```
bin/natalie --ast -e "a -= 1"
s(:block, s(:lasgn, :a, s(:call, s(:lvar, :a), :-, s(:lit, 1))))
```

richardboehme:
I replaced it with another lib, removed all std::string related stuff (from string, to string) and compilation time got even worse 🙃

richardboehme:
I added another library. The only library that it depends on is Natalie's String. Still the compile time is about 0.5 seconds slower than in the master branch... Can you tell me how to debug this? 😕

TheGrizzlyDev:
> I added another library. The only library that it depends on is Natalie's String. Still the compile time is about 0.5 seconds slower than in the master branch... Can you tell me how to debug this? confused

what'd happen if we used the current library and changed it to use natalie's String under the hood? How much would we be missing from String and would it take longer to compile?

richardboehme:
> what'd happen if we used the current library and changed it to use natalie's String under the hood? How much would we be missing from String and would it take longer to compile? Probably a naive question but I was curious, anyhow, nice work and looking forward to it!

Well I've already forgot about this option, but I quickly looked into it. It uses a lot of string's api (for example comparison operators). We could implement that but I'm not even sure it would have an effect...

TheGrizzlyDev:
> > what'd happen if we used the current library and changed it to use natalie's String under the hood? How much would we be missing from String and would it take longer to compile? Probably a naive question but I was curious, anyhow, nice work and looking forward to it!
> 
> Well I've already forgot about this option, but I quickly looked into it. It uses a lot of string's api (for example comparison operators). We could implement that but I'm not even sure it would have an effect...

would it not make sense to merge as is and then fix that forward? If it doesn't add too much time to the build and searching for alternatives is so hard, maybe we should just take the simple route, merge this then try to use TM::String after (which would imply adding quite a few API) and then if that works call it a day or keep optimising it? That, although not optimal, would unblock so many things that rely on bignum and that'd be quite cool :^)

seven1m:
>Still the compile time is about 0.5 seconds slower than in the master branch... Can you tell me how to debug this?

Is there a difference if you build in release mode? (`rake clean build_release`)

>would it not make sense to merge as is and then fix that forward?

If there is a clear boundary between the library and our own code, I'm ok with merging what you have done now. But when I looked at this PR it seemed there were assumptions built into our own IntegerValue based on what the library did and its interface. Build a bunch of work to have to turn around and do it over seems like a bad idea...

seven1m:
@richardboehme I don't yet know how to instrument compile time and where time is being spent. But I want to learn! I can take a look at it if you want to push up whatever branch you like the best aesthetically.

richardboehme:
> @richardboehme I don't yet know how to instrument compile time and where time is being spent. But I want to learn! I can take a look at it if you want to push up whatever branch you like the best aesthetically.

Aesthetically speaking I think the library we have in this PR is better, because it's less complex and easier to read (also easier to expand I think). However it may be less performant (in terms of runtime performance), because it's storing the bignum as a string. Additionally, it may be easier to optimize in terms of compile time, if std::string really makes a big impact.

seven1m:
Ship it! :shipit:

PR #218, opened 2021-10-05
require/load/require_relative now return a status

TheGrizzlyDev:
First of a bunch of PRs to make our requiring mechanism closer to MRI's one in terms of behaviour, without changing natalie's compiler nature. The focus for this PR is really just to make natalie's REPL run require/load expressions. Previously those were just expanding the required code's ast into the current node and thus the REPL would wrap that into a 
`p [your_expression].inspect` 
which would yield no result and evaluate to a method not found. 

By wrapping this in an if expression we first fix the issue above as we get the result of evaluating the if, but we also get closer to a more correct implementation of require_once, which should not be taken into account account by the compiler, but rather by the runtime. Following an example of that

```
if either_true_or_false
  # Do something 
  require "module_that_does_stuff"
else
  # Do something else
  require "module_that_does_stuff" 
end 
```

Natalie would always import the module with the first require in the AST, whereas MRI would do that based on the require invoked at runtime. Since require can alter the current execution context we want to make sure we keep said ordering, so Natalie could just always expand but then we can generate code where that if does not just evaluates to true, but checks if said module has been in fact loaded and if so jumps on the else, does nothing and returns false.


TheGrizzlyDev:
From the looks of it, the tests are failing because 
```ruby
if RUBY_ENGINE != 'natalie'
  # need this so we can run this script from any directory (even if that directory has a Gemfile of its own)
  ENV['BUNDLE_GEMFILE'] = File.expand_path('../Gemfile', __dir__)
  require 'bundler/setup'
end
```

was not evaluated before but now it is. Turns out that we aren't really able to resolve bundler yet, so that'd need extra work. @seven1m since this wasn't used, should I just comment it out for now, add a FIXME and come back later once we've fully implemented require/load/require_relative/autload?

seven1m:
This code is use on the MRI side, so we'll need a way to exclude code from compiling/running with Natalie.

seven1m:
This (`if RUBY_ENGINE...`) is a common pattern throughout the compiler - if we can detect this and *not* compile the require'd library for Natalie, that would be cool.

TheGrizzlyDev:
I am not sure I like this, it sounds quite complex... I'd rather go for a different route: we keep compiling as is and instead of throwing loaderror on compilation, we add that as a node to the AST in place of where the node containing the module would have been. This would also allow us to more closely imitate MRI while still keeping Natalie's nature of being a compiler unaltered. In that case though I'd add an optional flag to re-enable compile time loaderrors and another one to log them. What do you think about this?

seven1m:
So we'd rescue the LoadError like this?...

```ruby
begin
  require 'library_for_mri'
rescue LoadError
  if RUBY_ENGINE == 'natalie'
    # expected, carry on
  else
    raise
  end
end
```

(Or some such)

Yeah, I can go with that. I don't like that the compiler is doing unnecessary work, but someday we can add compile-time macros to Natalie and clean that up.

Let's go with your suggestion... 🤗

TheGrizzlyDev:
@seven1m I've added the LoadError mechanism to get closer to MRI's behaviour and get the tests to pass. I think this will influence the final executable size, so I'd definitely like to go for some kind of macro for natalie at some point (if you've got any idea this would make a very interesting video, just saying :D ). Aside from that we need to improve a few things:
- the once mechanism is not bound to compile time but rather runtime, so we should always load something at compile time, but then in that if check if something is loaded or not when needed
- loading a module with 'load' does not seem to influence require, so we are currently not behaving the same way (ie: something like load "pp"; require "pp" would return false in natalie and true in mri)
- we still need to add the runtime API in Kernel, but those could use the mechanism implemented for the issues above :)

PR #234, opened 2021-10-13
Add Array#cycle

richardboehme:
This just delegates to Enumerable right now. This also adjusts
Enumerable#cycle to be more precise in terms of MRI compatibility and
to make Array's specs pass.

[#39]


seven1m:
Thanks for fixing my typos 🤗

PR #238, opened 2021-10-14
Make Array#replace spec-compliant

imustafin:
[#39]


imustafin:
Right. I'll remove the code and keep only the spec then.

PR #239, opened 2021-10-15
Re-add skipped Array Specs

richardboehme:
I went through all array specs and un-skipped a lot of them. Most of them depended on other Array methods that weren't implemented yet or on bignum support.

To make the `Array#slice` spec work I added `Float#to_i` which was a bit hard to figure out. Please take a closer look at what I did there. Maybe there is a better way of doing it...


seven1m:
That's fine, but I will also try to figure out what the GC is doing. 😉

PR #240, opened 2021-10-15
Extract implicit conversion try_to_ary

imustafin:
This PR extracts the repeated implicit to array conversion to a separate function.

Unfortunately, there are some cases where we try to convert but on error we don't raise but do something else:
* `Array#<=>` tries to convert the argument but simply returns `nil` on failure ([spec](https://github.com/seven1m/natalie/blob/dfe11e9fb85db8086224e5d01117bc5132c220bd/spec/core/array/comparison_spec.rb#L94))
* `Array#initialize` tries to convert the first argument to an array or else to an integer ([to_ary](https://github.com/seven1m/natalie/blob/dfe11e9fb85db8086224e5d01117bc5132c220bd/src/array_value.cpp#L29), [to_int](https://github.com/seven1m/natalie/blob/dfe11e9fb85db8086224e5d01117bc5132c220bd/src/array_value.cpp#L40))
* And maybe more I haven't listed or not yet implemented in Natalie

# Solutions
I see several solutions:
1. Keep `try_to_ary` as is, such unusual conversions are coded as is

    **pros** When you see a `try_to_ary`, you know you don't want to catch it

    **cons** Coding unusual conversions produces duplicate code
2. Return some special ConversionResult from `try_to_ary` which has the conversion result and can raise the exception itself

    **pros** No exception performance penalty (is this really relevant?)

    **cons** Less readable code. The code becomes something like:
    ```cpp
    ConversionResult x = val->try_to_ary(env);
    if (!x.converted) x.raise();
    ArrayValue *ary = x.converted; // conversion ok, use the ary
    ```
3. Raise a special exception inheriting from ExceptionValue. It can optionally have the `#to_ary` result and can be rethrown.
    
    **pros** no-catch usage is still concise code-wise

    **cons** exception performance penalty (is this really relevant?)

I don't have much experience with cpp, but I think that the exception handling overhead is negligible in this case.

Currently I prefer the 3rd approach (throw specialized exceptions).

What do you think? Also, it might be worth merging this PR as-is too :)

### DIFF
--- a/src/array_value.cpp
+++ b/src/array_value.cpp
@@ -591,30 +591,9 @@ bool ArrayValue::_flatten_in_place(Env *env, nat_int_t depth, Hashmap<ArrayValue
         }
 
         if (!item->is_array()) {
-            auto sym_to_ary = SymbolValue::intern("to_ary");
-            auto sym_to_a = SymbolValue::intern("to_a");
-            auto original_item_class = item->klass();
-            ValuePtr new_item;
+            ValuePtr new_item = try_convert(env, item);
 
-            if (item->respond_to(env, sym_to_ary)) {
-                new_item = item.send(env, sym_to_ary);
-            } else if (item->respond_to(env, sym_to_a)) {
-                new_item = item.send(env, sym_to_a);
-            }
-
-            if (new_item != nullptr && !new_item->is_nil()) {
-                if (!new_item->is_array()) {
-                    auto original_item_class_name = original_item_class->class_name_or_blank();
-                    auto new_item_class_name = item->klass()->class_name_or_blank();
-                    env->raise(
-                        "TypeError",
-                        "can't convert {} to Array ({}#to_ary gives {})",
-                        original_item_class_name,
-                        original_item_class_name,
-                        new_item_class_name);
-                    return false;
-                }
-
+            if (!new_item->is_nil()) {
                 item = new_item;
             }
         }
--- a/test/natalie/array_test.rb
+++ b/test/natalie/array_test.rb
@@ -79,6 +79,35 @@ describe 'array' do
     end
   end
 
+  describe '#flatten' do
+    it 'raises TypeError if #to_ary of an element returns a non-array' do
+      x = Object.new
+      def x.to_ary
+        :sym
+      end
+
+      ->{ [x].flatten }.should raise_error(TypeError, "can't convert Object to Array (Object#to_ary gives Symbol)")
+    end
+
+    it 'keeps elements returning nil from #to_ary' do
+      x = Object.new
+      def x.to_ary
+        nil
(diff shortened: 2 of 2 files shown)


seven1m:
I agree the Conversion enum seems best and matches the existing pattern we used in `to_symbol`. For that matter, I would like to see this renamed `to_ary` if possible, just to match the existing `to_symbol`. You can rename the existing `ArrayValue::to_ary` to `to_ary_method` (and rename it also in `binding_gen.rb` to match) that way there isn't a conflict.

I really like this!

imustafin:
How are we going to distinguish "no conversion is possible" from "conversion gives nil" with the enum?

Some test code:
```ruby
objects = {}

objects[:absent] = Object.new

gives_ary = Object.new
def gives_ary.to_ary
  []
end
objects[:gives_ary] = gives_ary

gives_nil = Object.new
def gives_nil.to_ary
  nil
end
objects[:gives_nil] = gives_nil

gives_sym = Object.new
def gives_sym.to_ary
  :sym
end
objects[:gives_sym] = gives_sym

[:<=>, :replace].each do |method|
  objects.each do |name, obj|
    begin
      result = [].send(method, obj)
    rescue StandardError => e
      result = e
    end

    puts [method, name, '=>', result.nil? ? 'nil' : result.to_s].join(' ')
  end
end
```

This prints
```
<=> absent => nil
<=> gives_ary => 0
<=> gives_nil => nil
<=> gives_sym => can't convert Object to Array (Object#to_ary gives Symbol)

replace absent => no implicit conversion of Object into Array
replace gives_ary => []
replace gives_nil => can't convert Object to Array (Object#to_ary gives NilClass)
replace gives_sym => can't convert Object to Array (Object#to_ary gives Symbol)
```


Usually methods raise when there is no `to_ary` or if `to_ary` gives not an array, nil included (`#replace` example).

But `Array#<=>` returns `nil` if there is no `to_ary` or `to_ary` gives `nil`. If `to_ary` gives not an array, it raises.

Indeed, usually the conversion either succeeds or fails (as in `replace`), but sometimes we need more control and just having one `nullptr` is not enough the describe what happened.

What do you think? Maybe such unusual methods will occur rarely and we can go with the Conversion enum.

I hope I don't overthink this, please tell me if I do :)

seven1m:
I think I see what you mean. What if we add a new enum...

```cpp
enum class Conversion {
    Strict,
    NullAllowed,
};

enum class ToAryConversionMode {
    Strict,
    NilAllowed,
};

// ...

ArrayValue *to_ary(Env *env, Conversion conversion = Conversion::Strict, ToAryConversionMode conversion_mode = ToAryConversionMode::Strict);
```

seven1m:
This is really handy! 🤛

PR #242, opened 2021-10-16
Setup build and CI jobs to run aarch64

TheGrizzlyDev:


seven1m:
I tried to get this to work myself, but I didn't succeed. Honestly I think we should just wait for Sourcehut to add arm64 architecture support to their build system. Supposedly [it's on its way](https://man.sr.ht/builds.sr.ht/compatibility.md):

>Note: support for multi-arch builds is underway, but not yet available.

PR #243, opened 2021-10-16
Make Integer comparison spec compliant 

richardboehme:
This also changes the way Floats are converted to bignums. Instead of using dtoa I now use snprintf (see the new String constructors). This seems to yield similar results to ruby. For example applying dtoa to 9223372036854775847.0 yields 9223372036854776000.0 but snprintf yields 9223372036854775808.0 which is what ruby returns.

### DIFF
--- a/spec/core/integer/equal_value_spec.rb
+++ b/spec/core/integer/equal_value_spec.rb
@@ -1,0 +1,6 @@
+require_relative '../../spec_helper'
+require_relative 'shared/equal'
+
+describe "Integer#==" do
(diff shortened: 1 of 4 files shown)


seven1m:
This looks great. I originally imported dtoa in [this commit](https://github.com/seven1m/natalie/commit/0928b65d2f5c2fbf404d7312d67f23c2a6f26ebe) because I couldn't get snprintf to produce the same results as Ruby. But if you figured it out (seems the specs agree with you :smile:), then this is great. Can we remove dtoa.c entirely?

PR #244, opened 2021-10-16
Clean up spec files as we go

seven1m:
Hoping this will fix the nightly build:

Screenshot from 2021-10-16 16-14-58

### DIFF
--- a/spec/support/ruby_spec_runner.rb
+++ b/spec/support/ruby_spec_runner.rb
@@ -46,6 +46,8 @@ Dir[specs].each do |path|
       end
     }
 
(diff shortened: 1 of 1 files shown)


seven1m:
@richardboehme can you think of any other files we might need to clean up to conserve disk space?

richardboehme:
> @richardboehme can you think of any other files we might need to clean up to conserve disk space?

Well we could remove the *.rb spec file if it was compiled... But the whole ruby/spec repo is just 52 MB so I don't know if it would really help us. Unlinking the binaries should be fine for now.

Issue #195, opened 2021-09-26
Implement Array#pack


seven1m:
1. Remove "skip-test" comment from top of a `spec/core/array/pack/*_spec.rb` file.
2. Modify ArrayPacker and/or StringPacker to make the specs pass
3. Commit! (EASY! :laughing:)

- [x] a_spec.rb
- [x] at_spec.rb
- [x] b_spec.rb
- [x] buffer_spec.rb
- [x] comment_spec.rb
- [x] c_spec.rb
- [x] d_spec.rb
- [x] empty_spec.rb
- [x] e_spec.rb
- [x] f_spec.rb
- [x] g_spec.rb
- [x] h_spec.rb
- [x] i_spec.rb
- [x] j_spec.rb
- [x] l_spec.rb
- [x] m_spec.rb
- [x] n_spec.rb
- [x] percent_spec.rb
- [x] p_spec.rb
- [x] q_spec.rb
- [x] s_spec.rb
- [x] u_spec.rb
- [x] v_spec.rb
- [x] w_spec.rb
- [x] x_spec.rb
- [x] z_spec.rb

Crystalsage:
@seven1m It seems that directives n+N and p+P  have been implemented in PR #668. We can check them off the list.

seven1m:
🎉

PR #257, opened 2021-10-31
Array pack u directive

TheGrizzlyDev:
Added support for the u/U directives for Array#pack (#195):
- u -> uu-encoding of a string (https://en.wikipedia.org/wiki/Uuencoding)
- U -> UTF-8 encoding of a number

There is quite a bit going on in here :) 
- I've simplified the index tracking in StringHandler as in fact only a single index was needed to begin with
- Now be_computed_by expands the subject if it contains args to be passed to the method to test (an example of this can be found in the u spec ;) )
- Added an IntegerHandler that only currently supports the U directive
- Packer now returns a StringValue instead of a String so directives can now change the encoding (both u and U do, but u uses an ASCII encoding we do not support yet)
- When parsing tokens we now skip the count modifier of a directive if it is a zero byte (MRI does too, but it's only present in a subset of specs we did not work on until now, I found that rather funny and there's definitely lots of improvements we could eventually bring to ruby/spec :^) )

### DIFF
--- a/include/natalie/array_packer/integer_handler.hpp
+++ b/include/natalie/array_packer/integer_handler.hpp
@@ -1,0 +1,64 @@
+#pragma once
+
+#include "natalie/array_packer/tokenizer.hpp"
+#include "natalie/env.hpp"
+#include "natalie/string.hpp"
+
+namespace Natalie {
+
+namespace ArrayPacker {
+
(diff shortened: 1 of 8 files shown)


seven1m:
Nice! I like your refactoring too. 😍

PR #258, opened 2021-11-01
Don't look through all blocks when allocating new cells

richardboehme:
Previously we were looping through each block to find the next one with
a free cell. We had a early return but on some measurements I did the
loop ran on a average 60 iterations.

This patch introduces a list for keeping track which block has free
cells. Each time we want to allocate something we can just use the last
block of this list and do not have to iterate anymore.

Right now I use `push` and `pop` to add and remove from the vector. However, this means that a block newly pushed to the list will be picked first (LIFO). I made this decision because adding to and removing from the end of a vector is much more efficient but I'm not sure if the above is a problem for us. What do you think?

## Results:

I've looked at two scripts:

### Each:
```ruby
ary = []
(0..100_000).each { |n| ary << n + 1 }
```

#### master

Execution Time (lowest I got): 
```
real    0m0.184s
user    0m0.143s
sys     0m0.042s
```

This is the profile I got:

master-each

As we can see 19% of the instruction were in Heap::allocate.

#### After the patch

Execution Time (lowest I got): 
```
real    0m0.093s
user    0m0.092s
sys     0m0.001s
```

This is the profile I got:

new-each

As we can see the amount spent in Heap::allocate decreased significantly, which led to the speedup.

### Map:

```ruby
(0..100_000).map { |n| n + 1 }
```

I'm going to skip the profiles for this one. Basically it's the same but we spend even more time in Heap::allocate (53%) because this allocates much more objects.

#### master

Execution Time (lowest I got): 
```
real    0m2.248s
user    0m2.208s
sys     0m0.040s
```

#### After the patch

Execution Time (lowest I got): 
```
real    0m0.228s
user    0m0.178s
sys     0m0.050s
```

The last one shows the real benefit of this patch. The time it takes to allocate does not linearly scale with the amount of objects we already have allocated. This leads to this immense speedup 🎉

### DIFF
--- a/include/natalie/gc/allocator.hpp
+++ b/include/natalie/gc/allocator.hpp
@@ -51,14 +51,21 @@ public:
     }
 
     void *allocate() {
-        for (auto block : m_blocks) {
-            if (block->has_free()) {
-                --m_free_cells;
-                return block->find_next_free_cell();
-            }
+
+        if (m_free_blocks.size() > 0) {
+            auto *block = m_free_blocks.last();
+            --m_free_cells;
+            auto *cell = block->find_next_free_cell();
+            if (!block->has_free())
+                m_free_blocks.pop();
+            return cell;
         }
         auto *block = add_heap_block();
         return block->find_next_free_cell();
+    }
+
+    void add_free_block(HeapBlock *block) {
+        m_free_blocks.push(block);
     }
 
     bool is_my_block(HeapBlock *candidate_block) {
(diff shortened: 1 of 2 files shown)


include/natalie/gc/heap.hpp
```diff
@@ -19,6 +19,8 @@ class Heap {
 public:
     NAT_MAKE_NONCOPYABLE(Heap);
 
+    static long long allocations;
+
```
seven1m, line 23: Was this just for debugging? Can it be removed?

seven1m (approved):
Nice!!!

include/natalie/gc/heap.hpp
richardboehme: Oh yeah I was debugging allocations by counting them...  Thought I removed this already 🤔

Issue #201, opened 2021-09-27
Implement methods of Integer


richardboehme:
Our main approach is to work on one spec from [ruby/spec](https://github.com/ruby/spec/tree/master/core/integer) at a time, implementing as much of the spec as possible, then check it off the list:

- [x] abs_spec.rb
- [x] allbits_spec.rb
- [x] anybits_spec.rb
- [x] bit_and_spec.rb
- [x] bit_length_spec.rb
- [x] bit_or_spec.rb
- [x] bit_xor_spec.rb
- [x] case_compare_spec.rb
- [x] ceil_spec.rb
- [x] chr_spec.rb
- [x] coerce_spec.rb
- [x] comparison_spec.rb
- [x] complement_spec.rb
- [x] constants_spec.rb
- [x] denominator_spec.rb
- [x] digits_spec.rb
- [x] div_spec.rb
- [x] divide_spec.rb
- [x] divmod_spec.rb
- [x] downto_spec.rb
- [x] dup_spec.rb
- [x] element_reference_spec.rb
- [x] equal_value_spec.rb
- [x] even_spec.rb
- [x] exponent_spec.rb
- [x] fdiv_spec.rb
- [x] floor_spec.rb
- [x] gcd_spec.rb
- [x] gcdlcm_spec.rb
- [x] gt_spec.rb
- [x] gte_spec.rb
- [x] integer_spec.rb
- [x] lcm_spec.rb
- [x] left_shift_spec.rb
- [x] lt_spec.rb
- [x] lte_spec.rb
- [x] magnitude_spec.rb
- [x] minus_spec.rb
- [x] modulo_spec.rb
- [x] multiply_spec.rb
- [x] next_spec.rb
- [x] nobits_spec.rb
- [x] numerator_spec.rb
- [x] odd_spec.rb
- [x] ord_spec.rb
- [x] plus_spec.rb
- [x] pow_spec.rb
- [x] pred_spec.rb
- [x] rationalize_spec.rb
- [x] remainder_spec.rb
- [x] right_shift_spec.rb
- [x] round_spec.rb
- [x] size_spec.rb
- [x] sqrt_spec.rb
- [x] succ_spec.rb
- [x] times_spec.rb
- [x] to_f_spec.rb
- [x] to_i_spec.rb
- [x] to_int_spec.rb
- [x] to_r_spec.rb
- [x] to_s_spec.rb
- [x] truncate_spec.rb
- [x] try_convert_spec.rb
[...]

PR #260, opened 2021-11-23
Implement `odd?` for the `Integer`

mateusdeap:
This should tick one more box in #201. @TheGrizzlyDev I've decided to open this PR just so we don't pollute the issue thread. Like I said in discord, this is all quite new to me.

I'm also making this a draft just to make it clear this is not ready for reviewing and merging.

That being said, I did try to lookup how to make odd work for Bignums by looking at the MRI, but I don't think I'm actually implementing odd right even for a simple Fixnum

~Basing myself on the other methods, I added and IntegerValue argument to odd, in order to test wether it's a bignum, but I'm unsure if that's the correct thing to do.~

~I've not actually implemented an `odd` method for Bignum because I'm still trying to understand what a Bignum is and what does it mean to ask wether it is an odd number or not, in the context of computing: from the MRI it would seem that I need to take a specific digit (the left most, apparently) from the number and `&` it with 1 and that would mean it's odd. Is that really it?~

EDIT: I've since done some research and I believe I found what could be the right way to test wether a bignum is odd or not. What I do is take the bignum string, get the last char, convert it to an int and I mod that with 2 and test wether it is different from 0.

Not sure if this is right, but it seemed to be what MRI does and bignums seem to be the complement of a value, at least in the case of the fixnum specs.

The case for the bignum context I'll add in a comment below, as it's quite lengthy

### DIFF
--- a/include/natalie/bignum_value.hpp
+++ b/include/natalie/bignum_value.hpp
@@ -23,6 +23,16 @@ public:
         if (m_bignum) delete m_bignum;
     }
 
+    bool is_odd() override {
+        if (m_bignum->to_string().length() != 0) {
+            int last_digit = m_bignum->to_string().last_char() - '0';
+            bool is_odd = last_digit % 2 != 0;
+            return is_odd;
+        }
+        
+        return true;
+    }
+
     ValuePtr add(Env *, ValuePtr) override;
     ValuePtr to_s(Env *, ValuePtr = nullptr) override;
 
--- a/include/natalie/integer_value.hpp
+++ b/include/natalie/integer_value.hpp
@@ -27,7 +27,7 @@ public:
         return m_integer == 0;
     }
 
-    bool is_odd() {
+    virtual bool is_odd() {
         return m_integer % 2 != 0;
     }
 
--- a/spec/core/integer/odd_spec.rb
+++ b/spec/core/integer/odd_spec.rb
@@ -10,7 +10,6 @@ describe "Integer#odd?" do
       1.odd?.should be_true
       2.odd?.should be_false
 
-      puts bignum_value(0)
       bignum_value(0).odd?.should be_false
       bignum_value(1).odd?.should be_true
 
--- a/spec/core/integer/odd_spec.rb
+++ b/spec/core/integer/odd_spec.rb
@@ -1,0 +1,39 @@
+require_relative '../../spec_helper'
+
+describe "Integer#odd?" do
+  context "fixnum" do
+    it "returns true when self is an odd number" do
+      (-2).odd?.should be_false
+      (-1).odd?.should be_true
+
+      0.odd?.should be_false
+      1.odd?.should be_true
+      2.odd?.should be_false
+
+      puts bignum_value(0)
+      bignum_value(0).odd?.should be_false
+      bignum_value(1).odd?.should be_true
+
+      (-bignum_value(0)).odd?.should be_false
+      (-bignum_value(1)).odd?.should be_true
+    end
+  end
+
+  context "bignum" do
+    it "returns true if self is odd and positive" do
+      (987279**19).odd?.should be_true
+    end
+
+    it "returns true if self is odd and negative" do
+      (-9873389**97).odd?.should be_true
+    end
+
+    it "returns false if self is even and positive" do
+      (10000000**10).odd?.should be_false
+    end
+
+    it "returns false if self is even and negative" do
+      (-1000000**100).odd?.should be_false
+    end
+  end
+end
--- a/spec/core/integer/odd_spec.rb
+++ b/spec/core/integer/odd_spec.rb
@@ -19,21 +19,22 @@ describe "Integer#odd?" do
     end
   end
 
-  context "bignum" do
-    it "returns true if self is odd and positive" do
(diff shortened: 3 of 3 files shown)


mateusdeap:
So, some specs still fail and, apparently, for odd reasons.

1. There are 2 specs that call `(-bignum_value(num))` on 0 and 1. What I find weird is that, in all my runs, BOTH end up calling the `is_odd` method in the `IntegerValue` class and the value of `m_integer` is 1 for both. I tried running this method in irb but I do not get the same result. I get, as one would expect, the value of calling `bignum(0)` but with a minus sign. However, if I puts the value of `(-bignum_value(0))` or `(-bignum_value(1))` I get 1, so there's something odd between running the method in irb and how it's run in the context of the spec...

2. Also, all of the specs in the `context "bignum"` block fail. Here, there are different huge numbers being generated, however, once execution goes through the cpp code, it never calls `is_odd` for `BignumValue` but only for `IntegerValue` and there, the value of `m_integer` is always `-9223372036854775808`, no matter what number it actually is. Maybe it's a limitation on the side of printf? I still don't know

TheGrizzlyDev:
@mateusdeap yeah, the idea looks good to me :) Are the problems above still there? If so I can try having a look at it later

mateusdeap:
@TheGrizzlyDev yes. I didn't have much time to figure them out since I posted my comment.

seven1m:
@mateusdeap this is a good start. In order to get the `-bignum_value(num)` thing to work, we would need to implement `BignumValue::negate` as a virtual method too. Currently, when `-` (negate) is called, the Integer version is being called, and it doesn't properly handle the bignum.

It is a similar problem for the other tests, which use `num**power`. The `**` operator is bound to the `pow()` method on Integer, which has not yet been made Bignum-aware. For an example of what I mean by that, check out the `IntegerValue::add` method. You will see it properly promotes an overflowed result to a Bignum. The `pow` (`**`) method doesn't do that yet.

If I were you, I would probably comment out the failing tests, adding a note like `NATFIXME: need to work on IntegerValue::pow() so it promotes to Bignum`. Or something like that.

mateusdeap:
I see. Will do.

mateusdeap:
So, I've commented the specs that depend either on `IntegerValue::negate` or `IntegerValue::pow` and added comment explaining that once those are fixed, these should be revisited

TheGrizzlyDev:
> So, @seven1m and @TheGrizzlyDev, I've commented the specs that depend either on `IntegerValue::negate` or `IntegerValue::pow` and added comments explaining that once those are fixed, these should be revisited

Looks great, but a bit in contrast with @seven1m instead of commenting out I usually prefer using xit to exclude the test so that it's marked as skipped.

seven1m:
Yeah, I said "comment out" but I guess I meant skip. Sorry!

mateusdeap:
Hm... true. Had forgotten that that was a possibility. And indeed, much better practice. I'll just fix it up real quick

mateusdeap:
I asked this in the discord, but maybe you'll see it sooner here: Is there a way to mark a specific assertion as pending? I tried RSpec's `pending` and `skip`, but I get `NoMethodError`s. Any advice? Or should I just leave them commented?

TheGrizzlyDev:
> I asked this in the discord, but maybe you'll see it sooner here: Is there a way to mark a specific assertion as pending? I tried RSpec's `pending` and `skip`, but I get `NoMethodError`s. Any advice? Or should I just leave them commented?

All the testing machinery is rewritten in Natalie (somewhere under the test folder) so you could just add that yourself, or for the sake of simplicity just using xit instead of it now and leaving the comments with NATFIXME, what do you think about it?

seven1m:
It would be good to have _some_ specs for this yeah?

mateusdeap:
@seven1m So, considering the conversation I had with @TheGrizzlyDev, we decided to mark them all pending while he added the methods to only mark specific assertions as pending in specs.

However, I could just revert the last commit and have the second `it` block marked as pending and the first one would run, but 2 of its assertions would be commented out.

TheGrizzlyDev:
> @seven1m So, considering the conversation I had with @TheGrizzlyDev, we decided to mark them all pending while he added the methods to only mark specific assertions as pending in specs.
> 
> However, I could just revert the last commit and have the second `it` block marked as pending and the first one would run, but 2 of its assertions would be commented out.

I am honestly not a fan of a slightly divergent spec (other than just either skipping or completely commenting out) so I was thinking of a slightly different solution, can you add more tests (not the ones from ruby/spec) in the test package? Just covering the things you know to be working and have implemented? @seven1m how does this sound?

mateusdeap:
Yeah, sure. No probs.

seven1m:
Everyone, I'm sorry we killed the momentum on this. I went ahead and tweaked the specs so we can run a bit of _something_ and merged it. Thanks for your patience @mateusdeap, and please, keep up the work now that we ironed out some kinks in the process. :hugs:

TheGrizzlyDev:
Very cool first PR @mateusdeap !

mateusdeap:
Hey, no worries, and thanks a lot! As soon as I can I'll try and maybe tackle the methods that need fixing for the specs here to work!

PR #263, opened 2021-12-04
Make Integer calculations spec compliant

richardboehme:
This makes `Integer#-`, `Integer#*` and `Integer#/` spec compliant and fixes a bad bug that I introduced with BigInt!

I'm currently quite busy with university which is why I could not contribute that much in the last weeks 😞


seven1m:
Thanks for fixing that! Good luck on your studies!

PR #264, opened 2021-12-04
Implement Array#pack c- and C-directives

richardboehme:
This also adds an `Encoding::BINARY` constant as it's needed for one of the specs.

### DIFF
--- a/test/natalie/array_test.rb
+++ b/test/natalie/array_test.rb
@@ -1180,6 +1180,17 @@ describe 'array' do
     end
   end
 
+  describe '#pack' do
+    describe 'c' do
+      fit 'handle bignums' do
+        [fixnum_max + 1].pack('c').should == "\x00".force_encoding(Encoding::BINARY)
+        [fixnum_max + 2].pack('c').should == "\x01".force_encoding(Encoding::BINARY)
+        [-fixnum_max - 1].pack('c').should == "\x00".force_encoding(Encoding::BINARY)
+        [-fixnum_max - 2].pack('c').should == "\xFF".force_encoding(Encoding::BINARY)
+      end
+    end
+  end
+
   describe '#append' do
       specify do
         a = []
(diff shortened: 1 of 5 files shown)


richardboehme:
This currently seems to be incomplete (even though the specs already pass)...

seven1m:
@richardboehme what do you mean? Should we merge this as-is? Or are you saying there's a problem?

richardboehme:
> @richardboehme what do you mean? Should we merge this as-is? Or are you saying there's a problem?

There's definitely some misbehavior regarding larger numbers (> 256) because the c-directive should print 8 bit integers. Also I guess there must be some difference between c and C even though the spec treats it in the same way... So I'd say to not merge this as-is. I want to correct the behavior first.

richardboehme:
> There's definitely some misbehavior regarding larger numbers (> 256) because the c-directive should print 8 bit integers. Also I guess there must be some difference between c and C even though the spec treats it in the same way... So I'd say to not merge this as-is. I want to correct the behavior first.

First point seemed to be an illusion.... String#inspect yielded different results for some larger numbers because we only print them in the format '\xXX' if the character is less than 32. This resulted in Natalie giving me "�" when MRI returned "\x02". But they are the same so this should be fine!

I tried finding a difference between the 'c'- and the 'C'-directive but despite the difference of their definition (signed vs unsigned char) there does not seem to be any difference in calculating the result for those? I'm just giving up on this until somebody finds a bug with it 😅 

And lastly I added support for Bignums to our IntegerPacker. I hope making m_source an IntegerValue* instead of a nat_int_t is fine for you?

From my point of view we could merge this now 😄

seven1m:
Great work!

test/natalie/array_test.rb
```diff
@@ -1180,6 +1180,17 @@ def x.to_a
     end
   end
 
+  describe '#pack' do
+    describe 'c' do
+      fit 'handle bignums' do
```
seven1m, line 1185: Just noticed this is focused @richardboehme :grimacing:

richardboehme: Fixed in #266 😅

PR #265, opened 2021-12-05
Fix overflow calculation of Integer#* and Integer#-

richardboehme:
Yesterday I messed up the overflow detection of some of the new Integer calculation methods. This should fix now...
This also adds tests to check if the detection works correctly.


richardboehme:
This also brought up some failing specs that just worked because over-/underflow happened before... 😄

PR #266, opened 2021-12-06
Remove focused Array#pack test

richardboehme:
Missed this guy. I probably need to create a linter which makes PRs fail with one of those 😅


richardboehme:
@seven1m What do you think about raising an exception when a focused test is running in CI? (maybe controlled by an environment variable)

seven1m:
Yeah I think that would be great!

PR #267, opened 2021-12-06
Raise an error if a focused spec is running in CI

richardboehme:
And here it is: My own little helper to prevent mistakes with focused specs 😅


TheGrizzlyDev (approved):
I absolutely love this kind of small things, very cool!

PR #268, opened 2021-12-07
Make String#strip spec compliant, implement String#lstrip and String#rstrip

linusg:


seven1m:
Dude! Thanks so much!

seven1m:
Oops, should have copied in the spec files from ruby/spec. But you didn't know that. I'll do it in another commit later...

seven1m:
I added the specs in c856b1a9ecbec18a4c29ca95cb17686ddf281f30.

Issue #269, opened 2021-12-07
Docs on running specs needed


linusg:
Disclaimer: I'm a complete ruby noob :)

Just wrote my first couple of lines of code for natalie & found several issues (#195, #201, #217) referring to specs. While I did manage to get it to run some tests, the whole process was & still is really unclear to me, as I couldn't find any docs.

Specifically:

1. From various other PR's I noticed that spec files are added on a case by cases basis:
  - Are these copied from https://github.com/ruby/spec?
  - If so: why not use a submodule or separate clone of that repository? Manually copying files initially & then again for every upstream change seems incredibly cumbersome and error prone. We've had a good experience with using a plain checkout of the upstream [test262](https://github.com/tc39/test262/) repository for Serenity's LibJS, both for local development and CI.
2. Simply doing what the GitHub actions workflow causes thousands of unstaged git changes. Surely that's not the right way for testing locally? Or is the whole idea that you only run a subset of known working tests locally (part of the natalie repo) and all of them on CI?
  https://github.com/seven1m/natalie/blob/d012e0f86304f4d1282dc77f6d588c49e31ecc40/.github/workflows/run_all_specs.yml#L21-L29
3. Is there a way to specify a subdirectory or specific spec file? I had to edit this line manually:
  https://github.com/seven1m/natalie/blob/bc22ef43ae1200eb57f33ac7459b3e3c9209f48f/spec/support/ruby_spec_runner.rb#L20
4. Is there a way of getting detailed test failure output? I only got this:
   ```
   Start running specs spec/core/string/lstrip_spec.rb
   Ran spec spec/core/string/lstrip_spec.rb
   {"Successful Examples"=>2, "Failed Examples"=>2, "Errored Examples"=>2, "Compile Errors"=>0, "Crashes"=>0, "Timeouts"=>0}
   ```
5. How do I install the needed dependencies for the spec runner? The command from the CI workflow (`bundle exec ruby spec/support/ruby_spec_runner.rb`) failed with:
   ```
   spec/support/ruby_spec_runner.rb:1:in `require': cannot load such file -- concurrent (LoadError)
	   from spec/support/ruby_spec_runner.rb:1:in `<main>'
   ```
   and running `bundle add concurrent-ruby` didn't fix that. `gem install concurrent-ruby` + `ruby spec/support/ruby_spec_runner.rb` worked however. Again, no idea what I'm doing. :)

Thanks!

seven1m:
[...]

PR #271, opened 2021-12-08
Add some docs about specs/tests

seven1m:
This is to address some of the questions in #269.


linusg (commented):
Perfect, thanks for adding this!

PR #272, opened 2021-12-08
Less integer hydration

awesomekling:
Hello friends! :^)

Here are some patches to keep us on the happy path more often (by avoiding hydrating `Type::Integer` `ValuePtr`s)

With these changes, a simple increment loop runs ~twice as fast (from ~300ms to ~150ms):

```ruby
i = 0
while i < 1000000
    i += 1
end
```

### DIFF
--- a/src/value_ptr.cpp
+++ b/src/value_ptr.cpp
@@ -50,6 +50,13 @@ bool ValuePtr::is_float() {
     return value()->is_float();
 }
 
+bool ValuePtr::is_bignum() {
+    if (m_type == Type::Integer)
+        return false;
+
+    return value()->is_integer() && value()->as_integer()->is_bignum();
+}
+
 void ValuePtr::assert_type(Env *env, ValueType type, const char *type_name) {
     if (m_type == Type::Integer && type == ValueType::Integer)
         return;
--- a/include/natalie/value_ptr.hpp
+++ b/include/natalie/value_ptr.hpp
@@ -79,6 +79,7 @@ public:
 
     bool is_integer();
     bool is_float();
+    bool is_bignum();
     void assert_type(Env *, ValueType, const char *);
     nat_int_t to_nat_int_t();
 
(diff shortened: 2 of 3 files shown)


src/value_ptr.cpp
```diff
@@ -50,6 +50,13 @@ bool ValuePtr::is_float() {
     return value()->is_float();
 }
 
+bool ValuePtr::is_bignum() {
+    if (m_type == Type::Integer)
```
TheGrizzlyDev, line 54: Hmmm, not quite right, Bignum is still an integer. The concept comes from ruby itself where an int is either a Fixnum or Bignum (if too big or too small for a Fixnum - in our case long long (?)). So basically m_type == Type::Integer is actually a condition to be a bignum :)

`ruby
return m_type == Type::Integer && value()->is_integer() && value()->as_integer()->is_bignum();
`

TheGrizzlyDev (commented):
I really like the whole idea/process you went through (watching this while reviewing and I honestly wish this was the case for all the PRs, love it!). 

PS: if you want a repl for MRI's ruby (the standard one) you can use `irb`

src/value_ptr.cpp
awesomekling: Are you sure? `ValuePtr::Type::Integer` means that we use `ValuePtr::m_integer` instead of hydrating a full `IntegerValue` object (that may be either Fixnum or Bignum)

Since `ValuePtr::m_integer` is a `long long`, it seems to me that it must always be a Fixnum. But maybe I'm misunderstanding something?

TheGrizzlyDev: Ah I see, this actually works in ValuePtr, the type I was looking at was ValueType (in which case it'd be integer for both), but when it comes to ValuePtr::Type then Fixnums are Integer and Bignums are Pointer so you are right!

TheGrizzlyDev (approved):

PR #275, opened 2021-12-09
Handle behaviours of Integer over when swapping between fixnum/bignum

TheGrizzlyDev:
By moving the checks before the actual overflow can happen when doing multiplications. This pattern could be applied more extensively and will also be needed for pow when we do implement it.

I've also updated the constants for min and max finums in Natalie.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -187,14 +187,14 @@ file 'build/generated/numbers.rb' do |t|
   f2 = Tempfile.create('numbers')
   f2.close
   begin
-    f1.puts "#include <limits.h>"
     f1.puts "#include <stdio.h>"
+    f1.puts "#include \"natalie/constants.hpp\""
     f1.puts "int main() {"
-    f1.puts "  printf(\"NAT_MAX_FIXNUM = %lli\\n\", LLONG_MAX);"
-    f1.puts "  printf(\"NAT_MIN_FIXNUM = %lli - 1\\n\", LLONG_MIN + 1);"
+    f1.puts "  printf(\"NAT_MAX_FIXNUM = %lli\\n\", NAT_MAX_FIXNUM);"
+    f1.puts "  printf(\"NAT_MIN_FIXNUM = %lli\\n\", NAT_MIN_FIXNUM);"
     f1.puts "}"
     f1.close
-    sh "#{cc} -x c -o #{f2.path} #{f1.path}"
+    sh "#{cc} -Iinclude -x c -o #{f2.path} #{f1.path}"
     sh "#{f2.path} > #{t.name}"
   ensure
     File.unlink(f1.path)
--- a/src/integer_value.cpp
+++ b/src/integer_value.cpp
@@ -115,25 +115,26 @@ ValuePtr IntegerValue::mul(Env *env, ValuePtr arg) {
     } else if (!arg.is_integer()) {
         arg = Natalie::coerce(env, arg, this).second;
     }
+
     arg.assert_type(env, Value::Type::Integer, "Integer");
 
-    if (is_zero() || arg->as_integer()->is_zero())
-        return new IntegerValue { 0 };
+    if (to_nat_int_t() == 0 || arg.to_nat_int_t() == 0)
+        return ValuePtr::integer(0);
 
-    if (arg.is_bignum()) {
-        auto other = arg->as_integer();
-        auto result = to_bignum() * other->to_bignum();
-        return new BignumValue { result };
-    }
+    if (arg.is_bignum())
+        return (BignumValue { to_bignum() }).mul(env, arg);
 
     auto ll_this = to_nat_int_t();
     auto ll_arg = arg.to_nat_int_t();
 
-    auto min_fraction = (NAT_MIN_FIXNUM - (NAT_MIN_FIXNUM % ll_arg)) / ll_arg;
-    auto max_fraction = (NAT_MAX_FIXNUM - (NAT_MAX_FIXNUM % ll_arg)) / ll_arg;
-    if (
-        (ll_this > 0 && ll_arg > 0 && max_fraction <= ll_this) || (ll_this > 0 && ll_arg < 0 && min_fraction <= ll_this) || (ll_this < 0 && ll_arg > 0 && min_fraction >= ll_this) || (ll_this < 0 && ll_arg < 0 && max_fraction >= ll_this)) {
-        return (new BignumValue { to_bignum() })->mul(env, arg);
+    auto min_fraction = (NAT_MIN_FIXNUM) / ll_arg;
+    auto max_fraction = (NAT_MAX_FIXNUM) / ll_arg;
+
+    if ((ll_this > 0 && ll_arg > 0 && max_fraction <= ll_this)
+        || (ll_this > 0 && ll_arg < 0 && min_fraction <= ll_this)
+        || (ll_this < 0 && ll_arg > 0 && min_fraction >= ll_this)
+        || (ll_this < 0 && ll_arg < 0 && max_fraction >= ll_this)) {
+        return (BignumValue { to_bignum() }).mul(env, arg);
     }
 
     return ValuePtr::integer(ll_this * ll_arg);
--- a/src/integer_value.cpp
+++ b/src/integer_value.cpp
@@ -1,5 +1,6 @@
 #include "natalie.hpp"
 
+#include "natalie/constants.hpp"
 #include <math.h>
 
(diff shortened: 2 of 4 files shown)


src/integer_value.cpp
```diff
@@ -116,27 +117,26 @@ ValuePtr IntegerValue::mul(Env *env, ValuePtr arg) {
...
 
-    if (overflowed) {
-        auto other = arg->as_integer();
-        auto result = to_bignum() * other->to_bignum();
-        return new BignumValue { result };
+    auto min_fraction = (NAT_MIN_FIXNUM - (NAT_MIN_FIXNUM % ll_arg)) / ll_arg;
+    auto max_fraction = (NAT_MAX_FIXNUM - (NAT_MAX_FIXNUM % ll_arg)) / ll_arg;
+    if (
+        (ll_this > 0 && ll_arg > 0 && max_fraction <= ll_this) || (ll_this > 0 && ll_arg < 0 && min_fraction <= ll_this) || (ll_this < 0 && ll_arg > 0 && min_fraction >= ll_this) || (ll_this < 0 && ll_arg < 0 && max_fraction >= ll_this)) {
+        return (new BignumValue { to_bignum() })->mul(env, arg);
```
richardboehme, line 136: I think we should unify this with the other bignum case above. We should either call the mul method on BignumValue or do the multiplication in here. What do you think?

Also the if statement above looks a bit weird formatting-wise or is this just me?

src/integer_value.cpp
```diff
@@ -116,27 +117,26 @@ ValuePtr IntegerValue::mul(Env *env, ValuePtr arg) {
     }
     arg.assert_type(env, Value::Type::Integer, "Integer");
 
+    if (is_zero() || arg->as_integer()->is_zero())
```
richardboehme, line 120: This would cause hydration of the argument if is_zero() is false, right? Maybe we could just check for to_nat_int_t() == 0?

src/integer_value.cpp
```diff
@@ -116,27 +117,26 @@ ValuePtr IntegerValue::mul(Env *env, ValuePtr arg) {
...
+    auto ll_arg = arg.to_nat_int_t();
 
-    if (overflowed) {
-        auto other = arg->as_integer();
-        auto result = to_bignum() * other->to_bignum();
-        return new BignumValue { result };
+    auto min_fraction = (NAT_MIN_FIXNUM - (NAT_MIN_FIXNUM % ll_arg)) / ll_arg;
+    auto max_fraction = (NAT_MAX_FIXNUM - (NAT_MAX_FIXNUM % ll_arg)) / ll_arg;
+    if (
+        (ll_this > 0 && ll_arg > 0 && max_fraction <= ll_this) || (ll_this > 0 && ll_arg < 0 && min_fraction <= ll_this) || (ll_this < 0 && ll_arg > 0 && min_fraction >= ll_this) || (ll_this < 0 && ll_arg < 0 && max_fraction >= ll_this)) {
```
richardboehme, line 135: This looks a bit different to what [ruby ](https://github.com/ruby/ruby/blob/cb4e2cb55a59833fc4f1f6db2f3082d1ffcafc80/internal/bits.h#L102)does. Basically they do not need this modulo operation. Can you explain to me why it's there?

richardboehme (commented):
Thanks for fixing this up Antonio! I've added some comments about some details that were not clear to me 😅

src/integer_value.cpp
TheGrizzlyDev: Yes, in fact I don't like that so I will be adding a way to check if a ValuePtr is zero and integer without hydration to check, and use that to check arg. Nice call ;) 

PS: the rest of this 0-check will just stay the same and it will do hydration on arg, but only if really needed :)

src/integer_value.cpp
TheGrizzlyDev: I am not sure I would like to unify it, the case above is a faster path and doesn't need us to calculate the fractions, plus I think it is nicely separated.

richardboehme: What I meant by this is that in this branch we use
```cpp
(new BignumValue { to_bignum() })->mul(env, arg)
```

In the other faster path above we use 

```cpp
        auto result = to_bignum() * other->to_bignum();
        return new BignumValue { result };
```

I'm not quite sure which variant I prefer. The first one might be more consistent (for example if BignumValue multiplication does more that just multiply (any checks or whatever)). The second variant however is probably a bit faster because we save some checks (like checking for floats and so on). What do you think?

src/integer_value.cpp
richardboehme: I'm not sure if I understand the "PS" section 😄 Why would we need to hydrate the argument here? In any case we can just check if it's zero by calling to_nat_int_t() on it right? Like `is_zero() || arg.to_nat_int_t() == 0`

src/integer_value.cpp
TheGrizzlyDev: I tend to like the consistency of the second one, but I do see your point! In fact, now I am also quite conflicted, @seven1m  any preference? We should align those anyway, the question is which variant we will end up choosing :)

src/integer_value.cpp
TheGrizzlyDev: Oh I see, I see, we can honestly just use to_nat_int_t and totally skip the is_zero since to_nat_int_t in ValuePtr doesn't hydrate and is smart enough to check value() if it isn't an int!

src/integer_value.cpp
TheGrizzlyDev: yeah, I think I just dumbly added that thinking you could have an issue when arg = NAT_MAX_FIXNUM, but it doesn't seem to be the case and really doesn't make much sense :) Thanks!

src/integer_value.cpp
```diff
@@ -116,27 +117,26 @@ ValuePtr IntegerValue::mul(Env *env, ValuePtr arg) {
     }
     arg.assert_type(env, Value::Type::Integer, "Integer");
 
+    if (is_zero() || arg->as_integer()->is_zero())
+        return new IntegerValue { 0 };
```
seven1m:
```suggestion
        return ValuePtr::integer(0);
```

This will avoid a heap allocation.

src/integer_value.cpp
seven1m:
```suggestion
    if (is_zero() || arg->to_nat_int_t() == 0)
```

Avoiding a call to hydrate().

src/integer_value.cpp
seven1m: Yeah I don't care, I'll let you guys decide. But in the first case, you shouldn't need the `new` keyword since the intermediate value is thrown away:

```c++
return (BignumValue { to_bignum() })->mul(env, arg);
```

src/integer_value.cpp
seven1m: Well, you either amended the commit, or I had a really outdated page in my browser. This looks resolved.

src/integer_value.cpp
seven1m: Well, you either amended the commit, or I had a really outdated page in my browser. This looks resolved.

Rakefile
```diff
@@ -187,14 +187,14 @@ file 'build/generated/numbers.rb' do |t|
...
   f2.close
   begin
-    f1.puts "#include <limits.h>"
     f1.puts "#include <stdio.h>"
+    f1.puts "#include \"natalie/constants.hpp\""
     f1.puts "int main() {"
-    f1.puts "  printf(\"NAT_MAX_FIXNUM = %lli\\n\", LLONG_MAX);"
-    f1.puts "  printf(\"NAT_MIN_FIXNUM = %lli - 1\\n\", LLONG_MIN + 1);"
+    f1.puts "  printf(\"NAT_MAX_FIXNUM = %lli\\n\", NAT_MAX_FIXNUM);"
+    f1.puts "  printf(\"NAT_MIN_FIXNUM = %lli\\n\", NAT_MIN_FIXNUM);"
```
seven1m, line 194: This was smart -- not sure why I didn't think about it. My brain was melting under all the meta!

src/integer_value.cpp
seven1m: To clarify, I don't think I have enough info to know which is the better course. I think we just need to pick a path and figure out later if it was a mistake. 😄

seven1m (approved):

Rakefile
TheGrizzlyDev: Thanks!

PR #277, opened 2021-12-09
Implement Integer#integer?

timcraft:
Also include Comparable in Integer so that `spec/core/integer/integer_spec.rb` passes fully.

### DIFF
--- a/src/natalie.cpp
+++ b/src/natalie.cpp
@@ -72,6 +72,7 @@ Env *build_top_env() {
     global_env->set_Integer(Integer);
     Object->const_set(SymbolValue::intern("Integer"), Integer);
     Object->const_set(SymbolValue::intern("Fixnum"), Integer);
+    Integer->include_once(env, Comparable);
 
     ClassValue *Float = Numeric->subclass(env, "Float", Value::Type::Float);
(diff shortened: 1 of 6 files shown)


src/natalie.cpp
```diff
@@ -72,6 +72,7 @@ Env *build_top_env() {
     global_env->set_Integer(Integer);
     Object->const_set(SymbolValue::intern("Integer"), Integer);
     Object->const_set(SymbolValue::intern("Fixnum"), Integer);
+    Integer->include_once(env, Comparable);
```
richardboehme, line 75: Hm I think we do not need to do that. Normally Integer inherits Comparable by its parent class Numeric (which includes Comparable). I guess we would need to fix Class#include? to check parent classes too. I guess we can keep this for now but we should probably change that in the future.

richardboehme (approved):

src/natalie.cpp
seven1m: Let's add a FIXME comment for that so we don't forget to come back and clean that up.

seven1m: I went ahead and made `Module#include?` do the right thing so we can remove the extra include here.

PR #281, opened 2021-12-10
Make Integer#negate spec compliant

richardboehme:
Thanks to @TheGrizzlyDev's work in the last PR this was an easy one 👌 

[#201]


TheGrizzlyDev (approved):
very cool!

PR #283, opened 2021-12-10
Make purpose of our spec runner clearer

richardboehme:
This patch renames the spec runner to make it clearer that this file is used in CI. It also adds a comment on top of the file to document that further.

I'm obviously open to other names for the runner but I think this one would make it clear enough already.

Also my editor seemed to auto-re-format the github workflow file. Hope that's fine...


seven1m (approved):
Perfect.

PR #284, opened 2021-12-11
Optimize hash map capacity calculations

squeek502:
This is something I noticed while watching @awesomekling's recent video. He briefly noted that `calculate_map_size` was taking up a decent amount of time in the debug build before his `hydrate`-avoidance changes. This stood out to me since I had previously done some [optimizations in the Zig standard library](https://github.com/ziglang/zig/pull/2630) that I thought might apply here as well. It didn't end up being quite as much of a difference as I expected, but thought I'd submit a PR in case it's still desirable.

From the commit message:

Instead of doing `pow(2, i) - 1` in a loop, we can instead calculate
the target capacity and then bump it up to the next power of two.
Plus, because capacity is now always a power of two, (it was previously
always 1 less than a power of two) we can use `(hash & (capacity - 1))`
instead of `hash % capacity` when converting a hash to an index to further
get a performance boost.

---

Some benchmarking:

With the simple loop code in https://github.com/seven1m/natalie/pull/272 but *without* the changes in that PR applied (since the PR removes the Hashmap usage entirely from the generated code):

Debug build:
```
Benchmark #1: ./loop.before.dbg
  Time (mean ± σ):     873.1 ms ±   7.5 ms    [User: 796.3 ms, System: 76.1 ms]
  Range (min … max):   863.6 ms … 883.4 ms    10 runs
 
Benchmark #2: ./loop.optimized.dbg
  Time (mean ± σ):     767.7 ms ±  11.1 ms    [User: 689.3 ms, System: 77.7 ms]
  Range (min … max):   748.1 ms … 786.9 ms    10 runs
 
Summary
  './loop.optimized.dbg' ran
    1.14 ± 0.02 times faster than './loop.before.dbg'
```

Release build:
```
Benchmark #1: ./loop.before
  Time (mean ± σ):     251.0 ms ±   4.4 ms    [User: 173.9 ms, System: 76.3 ms]
  Range (min … max):   242.5 ms … 258.7 ms    12 runs
 
Benchmark #2: ./loop.optimized
  Time (mean ± σ):     244.7 ms ±   3.3 ms    [User: 170.6 ms, System: 73.1 ms]
  Range (min … max):   240.5 ms … 252.1 ms    12 runs
 
Summary
  './loop.optimized' ran
    1.03 ± 0.02 times faster than './loop.before'
```

With a modified version of https://github.com/martinus/map_benchmark to use only this hash map and only run the 'insert 10M integers` benchmark:

Screenshot 2021-12-10 at 21-52-44 Screenshot
(blue is insert 10M integers, yellow is clear, green is re-insert 10M integers, red is remove 10M integers one-by-one)

---

Note that there are better implementations for bumping up to the next power of two (using things like `__builtin_clz`), but I'm not familiar enough with C++ to know what implementation would be best/desired ([relevant issue for when I was investigating this stuff in Zig](https://github.com/ziglang/zig/issues/2426)). There is a `std::bit_ceil` that would work but it requires C++20.

---

Note: I also confirmed that the resulting capacity is unchanged (except for it now always being a power of two instead of 1 less than a power of two):

```
num_items => capacity_before_this_pr -> capacity_after_this_pr
0 => 15 -> 16
8 => 31 -> 32
16 => 63 -> 64
32 => 127 -> 128
64 => 255 -> 256
128 => 511 -> 512
256 => 1023 -> 1024
512 => 2047 -> 2048
1024 => 4095 -> 4096
2048 => 8191 -> 8192
4096 => 16383 -> 16384
8192 => 32767 -> 32768
16384 => 65535 -> 65536
32768 => 131071 -> 131072
65536 => 262143 -> 262144
131072 => 524287 -> 524288
262144 => 1048575 -> 1048576
524288 => 2097151 -> 2097152
1048576 => 4194303 -> 4194304
2097152 => 8388607 -> 8388608
4194304 => 16777215 -> 16777216
8388608 => 33554431 -> 33554432
```

<details>
<summary>Capacity calculation test code</summary>

```c++
#include <algorithm>
#include <cassert>
#include <cmath>
#include <stdio.h>

static constexpr float HASHMAP_TARGET_LOAD_FACTOR = 0.5;
class MaxCapacityException {};

size_t calculate_map_size(size_t num_items) {
  for (size_t i = 4; i < 10000; ++i) {
    auto size = pow(2, i) - 1;
    if (num_items / size < HASHMAP_TARGET_LOAD_FACTOR)
      return size;
  }
  throw new MaxCapacityException;
}

static constexpr size_t HASHMAP_TARGET_LOAD_FACTOR_INT = 50;

size_t calculate_map_size2(size_t num_items) {
  size_t target_size =
      std::max<size_t>(4, num_items) * 100 / HASHMAP_TARGET_LOAD_FACTOR_INT + 1;

  // Round up to the next power of two (if the value is not already a power of
  // two)
  // TODO: This can be replaced by std::bit_ceil in C++20
  target_size--;
  for (size_t i = 1; i < sizeof(size_t) * 8; i *= 2) {
    target_size |= target_size >> i;
  }
  return ++target_size;
}

int main() {
  size_t last_size = 0;
  for (size_t i = 0; i < 10000000; ++i) {
    auto old_size = calculate_map_size(i);
    auto new_size = calculate_map_size2(i);
    if (old_size != last_size) {
      printf("%zu => %zu -> %zu\n", i, old_size, new_size);
      last_size = old_size;
    }
    assert(old_size + 1 == new_size);
  }
  return 0;
}
```

</details>

### DIFF
--- a/include/tm/hashmap.hpp
+++ b/include/tm/hashmap.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <algorithm>
 #include <math.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -19,9 +20,9 @@ public:
     using HashFn = size_t(const void *);
     using CompareFn = bool(const void *, const void *, void *);
 
-    static constexpr float HASHMAP_MIN_LOAD_FACTOR = 0.25;
-    static constexpr float HASHMAP_TARGET_LOAD_FACTOR = 0.5;
-    static constexpr float HASHMAP_MAX_LOAD_FACTOR = 0.75;
+    static constexpr size_t HASHMAP_MIN_LOAD_FACTOR = 25;
+    static constexpr size_t HASHMAP_TARGET_LOAD_FACTOR = 50;
+    static constexpr size_t HASHMAP_MAX_LOAD_FACTOR = 75;
 
     struct Item {
         KeyT key;
@@ -129,7 +130,7 @@ public:
     Item *find_item(KeyT key, size_t hash, void *data = nullptr) {
         if (m_size == 0) return nullptr;
         assert(m_map);
-        auto index = hash % m_capacity;
+        auto index = index_for_hash(hash);
         auto item = m_map[index];
         while (item) {
             if (hash == item->hash && m_compare_fn(key, item->key, data))
@@ -154,7 +155,7 @@ public:
             item->value = value;
             return;
         }
-        auto index = hash % m_capacity;
+        auto index = index_for_hash(hash);
         key = duplicate_key(key);
         auto new_item = new Item { key, value, hash };
         insert_item(m_map, index, new_item);
@@ -164,7 +165,7 @@ public:
     T remove(KeyT key, void *data = nullptr) {
         if (!m_map) return nullptr;
         auto hash = m_hash_fn(key);
-        auto index = hash % m_capacity;
+        auto index = index_for_hash(hash);
         auto item = m_map[index];
         if (item) {
             // m_map[index] = [item] -> item -> item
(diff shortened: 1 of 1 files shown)


seven1m:
Well ain't that fancy! I appreciate you working on this.

PR #286, opened 2021-12-11
Make `Integer::sqrt` spec compliant

ryangjchandler:
Leaving this as a draft until I finish it off, here's my list of things still left:

(Edit 12/12/21) - Need to figure out how to handle Bignum values...

### DIFF
--- a/spec/core/integer/shared/next.rb
+++ b/spec/core/integer/shared/next.rb
@@ -1,0 +1,25 @@
+describe :integer_next, shared: true do
+  it "returns the next larger positive Fixnum" do
+    2.send(@method).should == 3
+  end
+
+  it "returns the next larger negative Fixnum" do
+    (-2).send(@method).should == -1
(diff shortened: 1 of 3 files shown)


ryangjchandler:
Still not sure about type coercion here, as well as Bignum handling...

TheGrizzlyDev:
> Still not sure about type coercion here, as well as Bignum handling...

Since bignum internally uses bigint and bigint supports sqrt, what would be the issue? Also, in this case you don't need to detect any overflow as the sqrt of a number will always be smaller than the number itself :) Maybe you can optimise things by converting from bignum to fixnum the result of sqrt if said result is in the boundaries of a fixnum, but I wouldn't worry too much about that :) 

Or do you need help on something different?

ryangjchandler:
I'll come back to this at some point this week to rebase and finalise the `sqrt` spec :)

PR #287, opened 2021-12-11
Avoid repetition when fetching constants and symbols from c++

TheGrizzlyDev:
The goal of this PR is to reduce the repetition when fetching constants and getting symbols from c++. 
1. For the first problem, I've added 2 trivial utility functions to fetch a top-level or nested constant
2. Created a shorthand syntax to invoke SymbolValue::intern from a string (yeah, copied StringView from Natalie, I didn't even know you could do that :D )

Anything else that might be worth improving here?


include/natalie/dsl.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "symbol_value.hpp"
+#include "string.hpp"
+#include "value_ptr.hpp"
+#include "global_env.hpp"
+#include "env.hpp"
+
+namespace Natalie {
+
+ValuePtr find_constant(Env *env, Vector<String> paths)
```
seven1m, line 11: I think we could use an `initializer_list` for this, like we do [here](https://github.com/seven1m/natalie/blob/d01faf8d593fda99592087fa431b18570c125024/src/encoding_value.cpp#L16), and get rid of the macro?

TheGrizzlyDev: Good call :) I don't really care about the amount of typing to be honest, as long as it is something clear and straightforward.

TheGrizzlyDev: Addressed this comment and added just a lil small operator to quickly convert strings to symbols, what do you think?

seven1m: I'm not sure! Haha. On the one hand, I like how quick it is to type. On the other, not being a fantastic C++ programmer nor having a lot of experience reading other C++ projects, will this make it harder for new people to understand what's going on?

Going to ponder this one for a bit...

include/natalie/dsl.hpp
```diff
@@ -0,0 +1,27 @@
...
+        if (c == nullptr) {
+            c = find_const(env, path);
+            continue;
+        }
+        c = c->const_find(env, path);
+    }
+    return c;
+}
+
+}
```
seven1m, line 27: What are your thoughts on this `dsl.hpp` header? On first glance, I think these two new functions should go in our junk drawer known as `natalie.hpp`, but maybe you have thoughts about cleaning that up?

TheGrizzlyDev: Nah, it was in this header because the initial idea was a DSL to simplify the Interaction from c++ to the Natalie runtime, but that wasn't needed so I'd rather move it to the Natalie header and if we ever think it needs to be cleaned up we will :)

include/natalie/dsl.hpp
TheGrizzlyDev: As a c++ noob I felt that was intuitive enough when I saw it in serenity, enough that I didn't even question it initially, only to then ask myself "hey, how does this voodoo trick even work???" And find out it's just an operator :) I wouldn't use it extensively but for concepts that are so close like symbols and strings, it kind of makes sense I think :)

seven1m: OK that makes sense. Let's do it.

TheGrizzlyDev:
Neat, thanks @seven1m !

seven1m:
No, thank you!!

PR #288, opened 2021-12-12
Array fix ignored specs

TheGrizzlyDev:
Went through some of the specs we skipped while developing Array and re-enabled the most trivial ones, which mostly had to do with bignum or open ranges.


src/array_value.cpp
```diff
@@ -40,6 +40,10 @@ ValuePtr ArrayValue::initialize(Env *env, ValuePtr size, ValuePtr value, Block *
     auto to_int = SymbolValue::intern("to_int");
     if (!size.is_integer() && size->respond_to(env, to_int))
         size = size->send(env, to_int);
+
+    if (size.is_bignum())
+        env->raise("RangeError", "bignum too big to convert into `long'");
```
richardboehme, line 45: Maybe not in scope of this PR but should we add a small helper for this? We already have IntegerValue::assert_fixnum but this requires hydration (because we need to call it on IntegerValue)...

richardboehme (commented):
That's pretty cool! 👍

src/array_value.cpp
TheGrizzlyDev: To be honest, I think this is a great idea! Since this method is only used in Array currently maybe we should just move it to ValuePtr instead, what do you think? It'd help us getting a bit more dehydrated :)

richardboehme: Yeah I guess that would be cool. Do you think `assert_fixnum` would automatically assert that it's an integer as well? This would mean we could remove the `size->assert_type` as well when we assert fixnum.

TheGrizzlyDev: Oh, that's how I am doing it to be honest!

Issue #291, opened 2021-12-13
Use const more consistently


seven1m:
As a C/C++ noob, and as a lazy human being, I have not been great about using `const` everywhere I should have. This issue will track progress toward a full "const-aware" Natalie.

I will be adding [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) configuration and tooling to help us find missing `const` qualifiers and to add them.

seven1m:
I don't believe this is complete, by any means, but I have made enough `const` changes that clang-tidy no longer yells at me. There are certainly other types we can mark const, but that will just have to be a discipline we embrace going forward.

PR #292, opened 2021-12-13
Add clang-tidy config and start working on adding `const` everywhere

seven1m:
This is some progress toward #291.


seven1m:
Oh, I forgot strlcpy isn't available everywhere.

PR #296, opened 2021-12-15
Unionise Value

Hendiadyoin1:
This decreases the size of Value by unionising the Object* and the fast_integer.
The test-performance does not seem to be affected significantly by this, but it
is still nice to decrease the memory footprint a bit.

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -31,8 +31,9 @@ void Value::hydrate() {
         // debugging VERY confusing. Maybe someday we can remove this GC stuff...
         bool was_gc_enabled = Heap::the().gc_enabled();
         Heap::the().gc_disable();
+        auto i = m_integer;
+        m_object = new IntegerObject { i };
         m_type = Type::Pointer;
-        m_object = new IntegerObject { m_integer };
         if (was_gc_enabled) Heap::the().gc_enable();
         break;
     }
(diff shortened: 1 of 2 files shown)


seven1m:
Thanks!

seven1m:
:thinking:

src/value.cpp
```diff
@@ -29,7 +29,6 @@ void Value::hydrate() {
     case Type::Integer: {
         m_type = Type::Pointer;
         m_object = new IntegerObject { m_integer };
```
TheGrizzlyDev, line 31: c++ noob here, but isn't this implying that Value has both m_object and m_integer at the same time? Could it be that when building not in release that is possible and somehow does not complain but when running the release build the union gets compacted and thus you'd end up with some kind of undefined behaviour?

seven1m: It's a good thought. We did try different ordering here, a temporary variable, etc., but it didn't seem to matter. But we may have missed something.

seven1m:
This made a wee-bit of a speed up on the boardslam example code, so that's neat.

Screenshot from 2021-12-21 20-31-26

seven1m:
Here's a slightly better speed-up from the [pemutation_spec.rb](https://github.com/seven1m/natalie/blob/master/spec/core/array/permutation_spec.rb):

Screenshot from 2021-12-21 21-10-35

Issue #217, opened 2021-10-04
Implement methods of String


seven1m:
Our main approach is to work on one spec from [ruby/spec](https://github.com/ruby/spec/tree/master/core/string) at a time, implementing as much of the spec as possible, then check it off the list:

- [ ] allocate_spec.rb
- [ ] append_spec.rb
- [ ] ascii_only_spec.rb
- [ ] b_spec.rb
- [ ] bytesize_spec.rb
- [ ] byteslice_spec.rb
- [ ] bytes_spec.rb
- [ ] capitalize_spec.rb
- [ ] casecmp_spec.rb
- [ ] case_compare_spec.rb
- [ ] center_spec.rb
- [ ] chars_spec.rb
- [ ] chomp_spec.rb
- [ ] chop_spec.rb
- [ ] chr_spec.rb
- [ ] clear_spec.rb
- [ ] clone_spec.rb
- [ ] codepoints_spec.rb
- [ ] comparison_spec.rb
- [ ] concat_spec.rb
- [ ] count_spec.rb
- [ ] crypt_spec.rb
- [ ] delete_prefix_spec.rb
- [ ] delete_spec.rb
- [ ] delete_suffix_spec.rb
- [ ] downcase_spec.rb
- [ ] dump_spec.rb
- [ ] dup_spec.rb
- [ ] each_byte_spec.rb
- [ ] each_char_spec.rb
- [ ] each_codepoint_spec.rb
- [ ] each_grapheme_cluster_spec.rb
- [ ] each_line_spec.rb
- [ ] element_reference_spec.rb
- [ ] element_set_spec.rb
- [ ] empty_spec.rb
- [ ] encode_spec.rb
- [ ] encoding_spec.rb
- [ ] end_with_spec.rb
- [ ] eql_spec.rb
- [ ] equal_value_spec.rb
- [ ] force_encoding_spec.rb
- [ ] freeze_spec.rb
- [ ] getbyte_spec.rb
- [ ] grapheme_clusters_spec.rb
- [ ] gsub_spec.rb
- [ ] hash_spec.rb
- [ ] hex_spec.rb
- [ ] include_spec.rb
- [ ] index_spec.rb
- [ ] initialize_spec.rb
- [ ] insert_spec.rb
- [ ] inspect_spec.rb
- [ ] intern_spec.rb
- [ ] length_spec.rb
- [ ] lines_spec.rb
- [ ] ljust_spec.rb
- [ ] lstrip_spec.rb
[...]

PR #298, opened 2021-12-16
String strip! family

RobertBendun:
Another 3 for #217


src/string_object.cpp
```diff
@@ -671,6 +679,32 @@ Value StringObject::lstrip(Env *env) const {
...
+    for (first_char = 0; first_char < len; first_char++) {
+        char c = c_str()[first_char];
+        if (!is_strippable_whitespace(c))
+            break;
+    }
+
+    if (first_char == 0)
+        return NilObject::the();
+
+    memcpy(&m_string[0], &m_string[0] + first_char, len - first_char);
```
seven1m, line 703: You cannot memcpy from one part of a string onto an overlapping part. From the man page for `memcpy`:

>If dst and src overlap, behavior is undefined.

So we'll need a temporary buffer.

seven1m (changes requested):

src/string_object.cpp
RobertBendun: Oh yes, wierldy I have been sure that is defined for this kind of copy, but we can use [memmove](https://en.cppreference.com/w/cpp/string/byte/memmove) without temporary buffer

seven1m: OK, I think I see. It's not actually overlapping, since it's copying characters to the "left". Ok, carry on.

seven1m (approved):

seven1m:
Thanks!

PR #300, opened 2021-12-18
Simplify first build steps for new folks :-)

seven1m:


TheGrizzlyDev (approved):
Nice stream and lovely non-passive-aggressive message :^)

PR #301, opened 2021-12-19
Make Integer#times spec-compliant

richardboehme:
Please take a close look at the compiler change as I'm not sure if this is correct. The process for this fix was more of a "trial and error"-thing than understanding what's going on.

[#201]

### DIFF
--- a/lib/natalie/compiler/pass1b.rb
+++ b/lib/natalie/compiler/pass1b.rb
@@ -27,12 +27,13 @@ module Natalie
           break_context[:raises_local_jump_error] = true
           exp.new(:raise_local_jump_error, :env, process(value), s(:l, 'LocalJumpErrorType::Break'))
 
-        when :loop, :rescue_do
+        when :loop
+          exp.new(:block, s(:set, var_name, process(value)), s(:c_break))
+        when :rescue_do
           exp.new(:block,
             s(:set, var_name, process(value)),
             (containing_context && containing_context[:type] == :fn ? s(:block) : s(:add_break_flag, var_name)),
             s(:c_break))
-
         else
           raise "unknown break context: #{type}"
         end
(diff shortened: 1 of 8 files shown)


test/natalie/integer_test.rb
```diff
@@ -184,6 +184,24 @@
     end
   end
 
+  describe '#times' do
+    fit 'handles bignums correctly' do
```
seven1m, line 188: focused test

test/natalie/break_test.rb
```diff
@@ -176,4 +176,16 @@ def break_in_begin_in_while
     result = break_in_begin_in_while { |x| break :ok }
     result.should == :ok
   end
+
+  fit 'should not leak break flag' do
```
seven1m, line 180: focused test :joy:

lib/natalie/compiler/pass1b.rb
```diff
@@ -27,12 +27,13 @@ def process_break(exp)
           break_context[:raises_local_jump_error] = true
           exp.new(:raise_local_jump_error, :env, process(value), s(:l, 'LocalJumpErrorType::Break'))
 
-        when :loop, :rescue_do
+        when :loop
+          exp.new(:block, s(:set, var_name, process(value)), s(:c_break))
+        when :rescue_do
```
seven1m, line 32: This is great! :1st_place_medal:

test/natalie/break_test.rb
richardboehme: Well... Seems like my self-protection measure did not work 😅 Will fix those tomorrow!

richardboehme: I think we need to pass the CI env variable to the docker container stuff somehow to make the "focus"-detection work?

seven1m: Yep! I fixed that in 81027a4bf73755256078de21556627678ef7034b just now. :hugs:

PR #302, opened 2021-12-19
Use Hashmap to store Allocator blocks instead of Vector

OhadRau:
Heap::collect() previously spent a lot of its time checking
whether a candidate block was a member of an allocator (when
collecting roots). This was an O(n) operation, where n is the
number of blocks in the allocator. To speed this up, we can
use a Hashmap as a set to quickly check in O(1). This brings
Heap::collect() down from 27.71% of dynamic instrs to 10.47%
in examples/boardslam.rb and probably benefits other code as
well.

The new code is a little funky and could benefit from a
Hashset class instead of a Hashmap in a future change.

### DIFF
--- a/include/natalie/gc/allocator.hpp
+++ b/include/natalie/gc/allocator.hpp
@@ -8,6 +8,7 @@
 
 #include "natalie/gc/heap_block.hpp"
 #include "natalie/macros.hpp"
+#include "tm/hashmap.hpp"
 #include "tm/vector.hpp"
 
 namespace Natalie {
@@ -23,8 +24,8 @@ public:
 
     ~Allocator() {
         for (auto block : m_blocks) {
-            block->~HeapBlock();
-            free(block);
+            block.first->~HeapBlock();
+            free(block.first);
         }
     }
 
@@ -69,11 +70,7 @@ public:
     }
 
     bool is_my_block(HeapBlock *candidate_block) {
-        for (auto block : m_blocks) {
-            if (candidate_block == block)
-                return true;
-        }
-        return false;
+        return m_blocks.get(candidate_block) != nullptr;
     }
 
     auto begin() {
@@ -86,7 +83,7 @@ public:
 
     void unmark_all_cells_in_all_blocks() {
         for (auto block : *this) {
-            block->unmark_all_cells();
+            block.first->unmark_all_cells();
         }
     }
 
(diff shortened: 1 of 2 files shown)


include/natalie/gc/allocator.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include "natalie/gc/heap_block.hpp"
 #include "natalie/macros.hpp"
+#include "tm/hashmap.hpp"
 #include "tm/vector.hpp"
```
Hendiadyoin1, line 12: is the vector include still needed then?

include/natalie/gc/allocator.hpp
```diff
@@ -104,15 +101,15 @@ class Allocator {
     HeapBlock *add_heap_block() {
         auto *block = reinterpret_cast<HeapBlock *>(aligned_alloc(HEAP_BLOCK_SIZE, HEAP_BLOCK_SIZE));
         new (block) HeapBlock(m_cell_size);
-        m_blocks.push(block);
+        m_blocks.put(block, block);
```
seven1m, line 104: You can treat the hashmap as a hashset by using `set()` instead of `put()`.

include/natalie/gc/allocator.hpp
```diff
@@ -104,15 +101,15 @@ class Allocator {
...
         add_free_block(block);
         m_free_cells += cell_count_per_block();
         return block;
     }
 
     size_t m_cell_size;
     size_t m_free_cells { 0 };
-    Vector<HeapBlock *> m_blocks;
+    Hashmap<HeapBlock *, HeapBlock *> m_blocks;
```
seven1m, line 112:
```suggestion
    Hashmap<HeapBlock *> m_blocks;
```

You don't need to specify the value type since we don't use it.

include/natalie/gc/allocator.hpp
```diff
@@ -86,7 +83,7 @@ class Allocator {
 
     void unmark_all_cells_in_all_blocks() {
         for (auto block : *this) {
```
seven1m, line 85: This makes me wish we had a `Hashmap<T>::keys()` method... :thinking:

seven1m:
This is great! Just a few small tweaks that I think would make it a bit nicer. The speedup is impressive!

include/natalie/gc/allocator.hpp
OhadRau: Yep, this is still used for `m_free_blocks`. That might also make sense as a different data structure, but I haven't looked into it yet

include/natalie/gc/allocator.hpp
OhadRau: Ahh that makes a lot of sense! Will fix that

include/natalie/gc/allocator.hpp
OhadRau: Happy to add something along those lines in this PR, haven't included it yet since I wasn't sure if that's out of scope?

seven1m: Let's do that in another PR. We can merge your PR and clean it up later. Thanks!

PR #304, opened 2021-12-19
Optimize Range's include? method for numeric values

OhadRau:
This is a small optimization for Range#include? when operating on numeric values. Previously, the method would iterate over all values from begin to end and compare them to the argument. Thankfully, the [Ruby spec](https://ruby-doc.org/core-3.0.3/Range.html#include-3F-method) allows us to implement this as equivalent to Range#cover? for numbers. In other words, we're able to simply do `arg >= begin && arg <= end`, which is a lot faster :) This PR also goes a little further and optimizes for fast integers whenever possible

### DIFF
--- a/src/range_object.cpp
+++ b/src/range_object.cpp
@@ -142,6 +142,13 @@ Value RangeObject::include(Env *env, Value arg) {
         if (begin <= i && i <= end)
             return TrueObject::the();
         return FalseObject::the();
+    } else if (m_begin->is_numeric() && m_end->is_numeric()) {
+        const auto leq = "<="_s, geq = ">="_s;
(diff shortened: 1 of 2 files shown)


seven1m (changes requested):
This is really nice!

One thing: this doesn't handle the case of the triple dots, e.g. `1...10`, which exclusive of the `10`. You'll need to check `m_exclude_end` to handle that.

OhadRau:
> This is really nice!
> 
> One thing: this doesn't handle the case of the triple dots, e.g. `1...10`, which is exclusive of the `10`. You'll need to check `m_exclude_end` to handle that.
> 
> Oh, and we'll need a test to prove that was broken. 😉

Good catch! Should be updated now (with the accompanying test)

seven1m (approved):

PR #305, opened 2021-12-20
Don't allocate copies of compile-time c-strings in Symbol

Hendiadyoin1:
I am not yet satisfied with the semantics of the constructor, but as a first draft this should suffice.

The impact on performance of the tests seems to be small.

### DIFF
--- a/include/natalie/symbol_object.hpp
+++ b/include/natalie/symbol_object.hpp
@@ -17,9 +17,10 @@ class SymbolObject : public Object {
 public:
     static SymbolObject *intern(const char *);
     static SymbolObject *intern(const String *);
+    static SymbolObject *intern_static_string(const char *, size_t);
 
     virtual ~SymbolObject() {
-        free(const_cast<char *>(m_name));
+        if (m_owned_string) free(const_cast<char *>(m_name));
     }
 
     const char *c_str() { return m_name; }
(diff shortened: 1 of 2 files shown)


include/natalie/symbol_object.hpp
```diff
@@ -68,21 +69,32 @@ class SymbolObject : public Object {
...
+    // FIXME: find a better semantic for this,
+    //        maybe some sort of badge
+    SymbolObject(const char *name, size_t)
+        : Object { Object::Type::Symbol, GlobalEnv::the()->Symbol() }
+        , m_should_free { false }
+        , m_name { name } {
+        assert(m_name);
+    }
 
+    bool m_should_free { false };
```
seven1m, line 92: Let's go with something like this:

```suggestion
    bool m_owned_string { false };
```

include/natalie/symbol_object.hpp
```diff
@@ -68,21 +69,32 @@ class SymbolObject : public Object {
 
     SymbolObject(Env *env, const char *name)
         : Object { Object::Type::Symbol, GlobalEnv::the()->Symbol() }
+        , m_should_free { true }
         , m_name { strdup(name) } {
         assert(m_name);
     }
```
seven1m, line 75: I actually think this constructor can go away? We shouldn't need this any more.

include/natalie/symbol_object.hpp
```diff
@@ -68,21 +69,32 @@ class SymbolObject : public Object {
...
+    // FIXME: find a better semantic for this,
+    //        maybe some sort of badge
+    SymbolObject(const char *name, size_t)
+        : Object { Object::Type::Symbol, GlobalEnv::the()->Symbol() }
+        , m_should_free { false }
+        , m_name { name } {
+        assert(m_name);
+    }
```
seven1m, line 90: Instead of using the length to signal the semantics, let's be explicit by passing a bool `owned_string`. Then we can have a single constructor:

```c++
    SymbolObject(const char *name, bool owned_string = false)
        : Object { Object::Type::Symbol, GlobalEnv::the()->Symbol() }
        , m_owned_string { owned_string }
        , m_name { owned_string ? name : strdup(name) } {
        assert(m_name);
    }
``

seven1m (changes requested):
This is really cool! Just a few tweaks...

seven1m:
:tada: This has a significant speed up when the script has a lot of symbols. Here is a synthetic benchmark with 10,000 symbols: https://gist.github.com/seven1m/537bde0626023953133649c01f2d9eed

Result:

Screenshot from 2021-12-20 19-56-05

seven1m (approved):

PR #307, opened 2021-12-20
Create install.sh

Mizan-Al-Sabbih:
So anyone can install it using one line command in terminal. I'm going to edit README.md file's installing section code, too (for saying people how to install it with one line command in terminal.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,8 @@
+echo "\e[0;34m [Installing Natalie, Please wait...]"
+
+git clone https://github.com/seven1m/natalie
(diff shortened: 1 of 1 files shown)


TheGrizzlyDev:
Good idea, can we follow the pattern described in https://github.com/github/scripts-to-rule-them-all ?

seven1m:
I'm sorry, but I don't think this is the direction I want to go with the install script. When we're ready (not right now), we'll consider adding a `rake install` task.

Please hop on Discord if you want to discuss!

PR #313, opened 2021-12-21
Initialize capacity of Arrays more often

richardboehme:
Following the PR #299 I sat down and tried to review all places where we initialize a new Array and set it's capacity if possible. This should speed up large array operations a bit.


seven1m:
Nice! :tada:

PR #317, opened 2021-12-23
Implement Integer#denominator

richardboehme:
[#201]


TheGrizzlyDev (approved):
Looks good to me, but I really don't get the point of this method :/

richardboehme:
> Looks good to me, but I really don't get the point of this method :/

I guess it's a common method for Rational and Integer so that both are interoperable. You should be able to just use denominator without having to care about whether it's an integer or a rational you're working with.

PR #318, opened 2021-12-23
Make Integer#<=> spec compliant

richardboehme:
I do not like the way I have to use the same if-condition twice in IntegerObject::cmp. Does somebody have an idea how this code could be done in a nicer way? 🤔 

[#201]

### DIFF
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -106,7 +106,12 @@ ArrayObject *block_args_to_array(Env *env, size_t signature_size, size_t argc, V
 
 void arg_spread(Env *env, size_t argc, Value *args, const char *arrangement, ...);
 
-std::pair<Value, Value> coerce(Env *, Value, Value);
+enum class CoerceInvalidReturnValueMode {
+    Raise,
+    Allow,
+};
+
+std::pair<Value, Value> coerce(Env *, Value, Value, CoerceInvalidReturnValueMode = CoerceInvalidReturnValueMode::Raise);
 
 char *zero_string(int);
 
--- a/src/integer_object.cpp
+++ b/src/integer_object.cpp
@@ -226,11 +226,13 @@ Value IntegerObject::pow(Env *env, Value arg) const {
 }
 
 Value IntegerObject::cmp(Env *env, Value arg) {
-    if (!arg.is_fast_integer()) {
-        arg.unguard();
-        if (!arg->is_integer() && !arg->is_float())
-            return NilObject::the();
-    }
+    // Check if we might want to coerce the value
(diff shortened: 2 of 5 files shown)


include/natalie.hpp
```diff
@@ -106,7 +106,7 @@ ArrayObject *block_args_to_array(Env *env, size_t signature_size, size_t argc, V
 
 void arg_spread(Env *env, size_t argc, Value *args, const char *arrangement, ...);
 
-std::pair<Value, Value> coerce(Env *, Value, Value);
+std::pair<Value, Value> coerce(Env *, Value, Value, bool raise = true);
```
seven1m, line 109: Can we make this an enum like we did with [ConstLookupFailureMode](https://github.com/seven1m/natalie/blob/master/include/natalie/object.hpp#L43-L46)?

src/integer_object.cpp
```diff
@@ -226,11 +226,13 @@ Value IntegerObject::pow(Env *env, Value arg) const {
...
-        if (!arg->is_integer() && !arg->is_float())
-            return NilObject::the();
-    }
+    // Check if we might want to coerce the value
+    if (!arg.is_fast_integer() && !arg->is_integer() && !arg->is_float())
+        arg = Natalie::coerce(env, arg, this, false).second;
+
+    // Check if compareable
+    if (!arg.is_fast_integer() && !arg->is_integer() && !arg->is_float())
+        return NilObject::the();
```
seven1m, line 235: Sadly I cannot think of a technique to eliminate this code duplication that itself would not make the code harder to read. 😞 I considered a lambda, but I don't think the added complexity is worth eliminating one duplicated line. 🤷

include/natalie.hpp
richardboehme: Yeah of course! That's a great idea 😄

src/integer_object.cpp
richardboehme: Yeah I thought about a lambda as well... But what do you think about a short method? Something like `is_comparable_with`?

TheGrizzlyDev: I think a lambda would bring less overhead and the compiler is very likely to inline it, so I don't think it'd be too bad :)

richardboehme: I added a lambda for it! @seven1m If you think it's too complex you can just remove the last commit. I'll be away for the next few days so I just did that in advance 😄 🎄

include/natalie.hpp
richardboehme: I added it! The name I chose is a bit mouthful to me but I couldn't think of a better one...

PR #320, opened 2021-12-24
Add Integer#div spec

ryangjchandler:
Already spec-compliant, was just missing from the spec sheet.


ryangjchandler:
Hm, tests are failing in CI but not locally.

ryangjchandler:
Will re-open after I've figured it out.

PR #321, opened 2021-12-24
Make `Integer#%` spec compliant

ryangjchandler:


seven1m:
@ryangjchandler you may already know this, but this modulo_spec is failing due to a mod zero. Whenever the program crashes like that, you can get some help from gdb or lldb by running `-d gdb` or `-d lldb`, like this:

Screenshot from 2021-12-28 08-24-39

>Program received signal SIGFPE, Arithmetic exception.

I added a printf just above that code and got:

```
m_integer = 13, arg = 0
```

ryangjchandler:
Yeah, thanks @seven1m. Going to take a look at this PR later on today, probably just needs some `0` specific handling :)

ryangjchandler:
Trying to `x % 0` now throws a `ZeroDivisonError` which covers a lot of the failing test cases. The only outstanding ones now are related to handling the actual `%` operator correctly.

ryangjchandler:
Got the main math for the `mod` operator working (finally), was just a case of casting from `long long` to a `float` when dividing (always something simple!)

Need to support bignums now too, since they're not behaving (as per).

seven1m:
@ryangjchandler I had a bit of time and worked on this. I'm not 100% confident about my change to BigInt, but it seems to work and act like Ruby now. @richardboehme if you have any feedback about my change to BigInt, please chime in! :hugs:

richardboehme:
> @richardboehme if you have any feedback about my change to BigInt, please chime in! 🤗

Looks good to me 👍

ryangjchandler:
@seven1m Thanks! Didn't get any time to come back to this and didn't _really_ want to touch `BigInt` since I'm not super familiar with internals etc.

PR #323, opened 2021-12-27
Regex spec compliance

TheGrizzlyDev:
I tried to push the spec-compliant of Regexp as much as possible, but I've encountered a few issues with Encoding (missing lots of encodings :) ) and how we handle that for a regex during compilation (on a quick analysis it looks like we lose most encodings at compile-time)

Specs covered here:

- case_compare_spec.rb
- casefold_spec.rb
- eql_spec.rb
- equal_value_spec.rb
- hash_spec.rb
- initialize_spec.rb
- inspect_spec.rb
- last_match_spec.rb
- match_spec.rb
- options_spec.rb
- to_s_spec.rb

I've had to drop a few TODOs but they're either related to encodings or the use of Class.allocate (to allocate an object without calling initialize) which felt a bit too much out of scope :^)

PS: you might be wondering why I did so much stuff in a single PR. Well, funny story I was trying to make compile spec-compliant, but as you might notice it isn't even in this list, so what happened? Things are quite intertwined around here and compile basically needs you to implement almost everything before you are even able to test it. Halfway through and a few methods later, I've decided to cut this short or otherwise it'd have been a mess to review :^)


seven1m:
You're nuts! 😅

TheGrizzlyDev:
> You're nuts! sweat_smile

Should be good to review now, have fun :D

src/regexp_object.cpp
```diff
@@ -79,8 +127,150 @@ Value RegexpObject::match(Env *env, Value other, size_t start_index) {
...
+                break;
+            case 'i':
+                will_be_i = active;
+                break;
+            case 'x':
+                will_be_x = active;
+                break;
+            case '-':
+                if (!active) // this means we've already encountered a '-' which is illegal, so we just to append_options;
+                    goto append_options;
```
seven1m, line 208: :scream: I can't do the `goto`, sorry. Can we achieve the same result with a boolean `append_options = true` then falling through to subsequent code?

src/regexp_object.cpp
```diff
@@ -79,8 +127,150 @@ Value RegexpObject::match(Env *env, Value other, size_t start_index) {
...
+
+    if (is_m) out->append_char('m');
+    if (is_i) out->append_char('i');
+    if (is_x) out->append_char('x');
+
+    if (!(is_m && is_i && is_x)) out->append_char('-');
+
+    if (!is_m) out->append_char('m');
+    if (!is_i) out->append_char('i');
+    if (!is_x) out->append_char('x');
```
seven1m, line 246: Can you explain what's going on here? I see we append 'mix' both if they values are true **and** if they are false. Confused!

seven1m (changes requested):
This is really great and a ton of work. I just can't go for the `goto`. :grimacing:

src/regexp_object.cpp
TheGrizzlyDev: Let me try cleaning this up, I might have an idea :)

src/regexp_object.cpp
TheGrizzlyDev: Oh yeah, this is definitely a confusing behaviour, but nothing too crazy :^) Basically if ANY of the options is missing, then it will be appended after a '-' like in the examples below:

`ruby
irb(main):001:0> //mix.to_s
=> "(?mix:)"
irb(main):002:0> //ix.to_s
=> "(?ix-m:)"
irb(main):003:0> //x.to_s
=> "(?x-mi:)"
`

src/regexp_object.cpp
TheGrizzlyDev: does this look any better now? 
Anyhow, the idea here was to stop parsing the options in the top-level group if the format was incorrect or the top level group wasn't the only top level group nor completely wrapping the regex :)

src/regexp_object.cpp
seven1m: Ahh ok!

seven1m (approved):

PR #328, opened 2021-12-30
Implement a 'canary' pass that optimizes well-know patterns in ruby code

TheGrizzlyDev:
This adds a concept of pass-0, which doesn't 'lower' the AST to the c++ one but rather transforms a ruby-ast into a 'simplified' ruby-ast.
By simplified we do not mean during compilation time, but rather at runtime, so we can avoid unnecessary overhead (like allocating an array to get the min/max out of a few numbers).

Inspired by: https://ruby-compilers.com/examples/#example-program-canaryrb


seven1m:
Whoa cool! 😎

TheGrizzlyDev:
> Whoa cool! sunglasses

this first version is still a bit naive but I guess it illustrates the process :) I wonder though just how many methods we can improve this way, should I create an issue and start collecting them?

TheGrizzlyDev:
Thanks for the fixes @seven1m !

PR #329, opened 2021-12-30
Implement `::allocate` and `nil` methods

FnControlOption:
Skipped `NilClass#to_c`, `#to_r`, and `#rationalize` because they require `Complex` and `Rational`

Ported mspec's [`SpecVersion`](https://github.com/ruby/mspec/blob/master/lib/mspec/utils/version.rb)

### DIFF
--- a/include/natalie/nil_object.hpp
+++ b/include/natalie/nil_object.hpp
@@ -22,9 +22,14 @@ public:
         return s_instance;
     }
 
+    bool and_method(Env *, Value);
+    bool or_method(Env *, Value);
+    bool eq(Env *, Value);
     Value eqtilde(Env *, Value);
     Value to_s(Env *);
     Value to_a(Env *);
+    Value to_h(Env *);
+    Value to_f(Env *);
     Value to_i(Env *);
     Value inspect(Env *);
 
--- a/include/natalie/string_object.hpp
+++ b/include/natalie/string_object.hpp
@@ -19,7 +19,8 @@ public:
     const int STRING_GROW_FACTOR = 2;
 
     StringObject(ClassObject *klass)
-        : Object { Object::Type::String, klass } { }
+        : Object { Object::Type::String, klass }
+        , m_encoding { Encoding::ASCII_8BIT } { }
 
     StringObject()
         : StringObject { "" } { }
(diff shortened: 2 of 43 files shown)


test/support/version.rb
```diff
@@ -0,0 +1,55 @@
...
+  def <=>(other)
+    if other.respond_to? :to_int
+      other = other.to_int
+    else
+      other = SpecVersion.new(other.to_s).to_i
+    end
+
+    self.to_i <=> other
+  end
+end
```
seven1m, line 55: Since this is a direct copy of the file from the mspec library, we will need to make a note in the README about using part of mspec, which is MIT. Up until now, we had re-implemented everything that we needed, not directly looking at mspec code. This is the first file where it was just copied.

seven1m: Oh, and we should add the copyright to the top of the file just to be clear we didn't write this...

include/natalie/string_object.hpp
```diff
@@ -189,6 +189,6 @@ class StringObject : public Object {
     void raise_encoding_invalid_byte_sequence_error(Env *, size_t) const;
 
     String m_string {};
-    Encoding m_encoding { Encoding::UTF_8 };
+    Encoding m_encoding { Encoding::ASCII_8BIT };
```
seven1m, line 192: Can you explain why this change was necessary?

include/natalie/nil_object.hpp
```diff
@@ -22,9 +22,14 @@ class NilObject : public Object {
         return s_instance;
     }
 
+    bool _and(Env *, Value);
+    bool _or(Env *, Value);
```
seven1m, line 26: Our convention is to call these `and_method` and `or_method`

seven1m:
This is great work. The PR itself is kinda all over the place... in the future, it would help with PR review if you could keep it to a single logical change...

The change of default encoding on StringObject has me really scratching my head, so I'm hoping we can put that back to UTF8.

include/natalie/string_object.hpp
FnControlOption: `String::allocate` (and `String::new` without an `encoding:` argument) default to ASCII-8BIT; I've found a better way to do this

seven1m:
I rebased this against master and merged each 'SQUASHME' commit with the commit it accompanied. That was a bit of work, so I'll let you do that next time @FnControlOption. :stuck_out_tongue:

PR #330, opened 2021-12-31
Apply the same Value optimizations we do for integers for floats as well

Hendiadyoin1:
In the hope I did not miss anything...

I did not yet do a performance test

(this also adds .vscode to .gitignore)

### DIFF
--- a/include/natalie/float_object.hpp
+++ b/include/natalie/float_object.hpp
@@ -57,20 +57,6 @@ public:
         return m_double;
     }
 
-    Value to_int_no_truncation(Env *env) {
-        if (is_nan() || is_infinity()) return this;
-        if (m_double == ::floor(m_double)) {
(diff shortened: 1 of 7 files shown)


ryangjchandler:
@Hendiadyoin1 You could always add the vscode folder to a global gitignore instead of every project, just a small tip :)

include/natalie/float_object.hpp
```diff
@@ -143,19 +129,19 @@ class FloatObject : public Object {
     Value to_f() { return this; }
     Value to_i(Env *) const;
     Value to_s(Env *) const;
-    Value truncate(Env *, Value) const;
+    Value truncate(Env *, Value);
```
seven1m, line 132: I hate that we lost these `const` qualifiers, but I know why it happened. Hoping someday I can figure out how to hydrate a Value without needing to remove `const` everywhere.... :thinking:

seven1m:
Sweet!!

PR #332, opened 2022-01-01
Parse method names with `[]` and `[]=`

FnControlOption:
For example:

```rb
class Matrix
  def Matrix.[](*rows)
    rows
  end
end

Matrix[1,2,3] # => [1,2,3]
# is the same as
Matrix.[](1,2,3) # => [1,2,3]
```


seven1m:
Nice!

PR #333, opened 2022-01-01
Teaching the compiler about internal Floats and fixing them up

Hendiadyoin1:
I temporarily removed the previous implementation of it, to narrow down on some errors.

Current errors include:
* [X] `Kernel.Float: returns the identical Float for numeric Floats   1.12 should be equal to 1.12 (line 9)`
* [x] `Kernel.Float: returns the identical NaN if to_f is called and it returns NaN   NaN should be equal to NaN (line 281)`
* [x] `Float#divmod: returns an [quotient, modulus] from dividing self by other  -Infinity should be within 3.0e-05 of 2.8284 (line 10)`

### DIFF
--- a/include/natalie/value.hpp
+++ b/include/natalie/value.hpp
@@ -62,13 +62,9 @@ public:
         if (is_fast_integer()) {
             if (other.is_fast_integer())
                 return get_fast_integer() == other.get_fast_integer();
-            if (other.is_fast_float())
-                return get_fast_integer() == other.get_fast_float();
             return false;
         }
         if (is_fast_float()) {
-            if (other.is_fast_integer())
-                return get_fast_integer() == other.get_fast_integer();
             if (other.is_fast_float())
                 return get_fast_float() == other.get_fast_float();
             return false;
@@ -77,21 +73,7 @@ public:
     }
 
     bool operator!=(Value other) const {
-        if (is_fast_integer()) {
-            if (other.is_fast_integer())
-                return get_fast_integer() != other.get_fast_integer();
-            if (other.is_fast_float())
-                return get_fast_integer() != other.get_fast_float();
-            return true;
-        }
-        if (is_fast_float()) {
-            if (other.is_fast_integer())
-                return get_fast_integer() != other.get_fast_integer();
-            if (other.is_fast_float())
-                return get_fast_float() != other.get_fast_float();
-            return true;
-        }
-        return m_object != other.object();
+        return !(*this == other);
     }
 
     bool is_null() const { return m_type == Type::Pointer && !m_object; }
--- a/include/natalie/value.hpp
+++ b/include/natalie/value.hpp
@@ -28,22 +28,13 @@ public:
     explicit Value(nat_int_t integer)
         : m_type { Type::Integer }
         , m_integer { integer } { }
-    explicit Value(double value)
-        : m_type { Type::Double }
-        , m_double { value } { }
 
     static Value integer(nat_int_t integer) {
         // This is required, beacause initialization by a literal is often
         // ambiguous.
         return Value { integer };
     }
-    static Value floatingpoint(double value) {
-        // This is used in the code gen
-        // auto v =  Value { value };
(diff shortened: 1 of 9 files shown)


include/natalie/value.hpp
```diff
@@ -80,7 +86,21 @@ class Value {
     }
 
     bool operator!=(Value other) const {
-        return !(*this == other);
+        if (is_fast_integer()) {
+            if (other.is_fast_integer())
+                return get_fast_integer() != other.get_fast_integer();
+            if (other.is_fast_float())
+                return get_fast_integer() != other.get_fast_float();
+            return false;
```
TheGrizzlyDev, line 94: Neither this nor the other non-hydrating return false at line 101 is true, what if you are comparing with nulltpr? Or what if you are comparing with an hydrated int/float? I'd definitely suggest adding an explicit check for nullptr but also removing return false here (and the other way around for ==) and letting it fall-through a check with the hydrated objects at the end (or just simply fall-through the way it was before). Having said that I think it's a really good idea checking fast-non-hydrated-values :^)

Hendiadyoin1:
The source of the problem might be that we are not calling to_f directly.
We are calling `main_object.object#Float`, which does not go through the optimization check
This might cause a hydration after copy, which is not what we want.
Just my speculation though

That check is weird though, and I think that nothing in the real world should depend on that behavior....

Another way to fix this is to check for NaN after every maybe NaN-producing function and force hydrate them.
This would include *all* Math functions and Float#div, for example.

include/natalie/value.hpp
```diff
@@ -80,7 +86,21 @@ class Value {
...
+            return true;
+        }
+        if (is_fast_float()) {
+            if (other.is_fast_integer())
+                return get_fast_integer() != other.get_fast_integer();
+            if (other.is_fast_float())
+                return get_fast_float() != other.get_fast_float();
+            return true;
+        }
+        return m_object != other.object();
```
seven1m, line 103: I don't think this duplication is neccesary. This *should* be equivalent to boolean NOT of `==`.

lib/natalie/compiler/pass4.rb
```diff
@@ -674,6 +674,7 @@ def process_sexp(exp, name = nil, type = 'Value ')
         fn, *args = exp
 
         return "Value::integer(#{args.map { |a| process_atom(a) }.join(', ')})" if fn == :'Value::integer'
+        return "Value::floatingpoint(#{args.map { |a| process_atom(a) }.join(', ')})" if fn == :'Value::integer'
```
seven1m, line 677: I think this is supposed to check `Value::floatingpoint`? Honestly don't remember what this code is doing... :grimacing:

src/float_object.cpp
```diff
@@ -30,6 +30,7 @@ bool FloatObject::eq(Env *env, Value other) {
         return m_double == other.get_fast_integer();
     if (other.is_fast_float())
         return m_double == other.get_fast_float();
+    other.unguard();
```
seven1m, line 33: We lose a bunch of benefit of this being an "optimized" method if we call unguard here.

seven1m:
We still need to replace all `new FloatObject {...}` with `Value::floatingpoint(...)`. I can work on that later, or if someone wants to lend a hand, please do! <3

seven1m:
Welp, I broke it. 😆 I will work on fixing this later...

include/natalie/value.hpp
```diff
@@ -143,6 +136,18 @@ class Value {
             hydrate();
     }
 
+    bool fast_equal(Value other) const {
+        if (is_fast_integer() && other.is_fast_integer())
+            return get_fast_integer() == other.get_fast_integer();
+        else if (is_fast_integer() && other.is_fast_float())
+            return get_fast_integer() == other.get_fast_float();
```
seven1m, line 143: I just realized this is wrong. The `operator==` check is to check if the objects are the _same pointer_ -- not logically equivalent. A float and an int would never _be the same pointer_.

PR #334, opened 2022-01-01
Make `String#concat` (mostly) spec compliant

ryangjchandler:
There's quite a few encodings being used here that we don't support at the moment so I'm skipping those and adding a `NATFIXME` comment above.

The rest of the spec is passing though. Not entirely sure what it takes to add support for other encodings.


seven1m:
Works for me! I'm not looking forward to all the work we need to do with encodings actually. :grimacing:

PR #339, opened 2022-01-01
Implement `Module#private_constant`, `#public_constant`

FnControlOption:


seven1m:
I didn't know about private constants in Ruby. Interesting!

PR #343, opened 2022-01-02
Some more canary features [basic constant evaluation in p0c]

Hendiadyoin1:
This adds the ability to coalesce constant arithmetic and if-statements to pass0c
With this we should save some send calls on primitive data types

Here follows a quick demonstration

<details>
<summary>input</summary>

```ruby
x = 1 unless true

1 & 3
2 | 1
3 ^ 2

1 << 4
32 >> 3

[1, 2, 3].max
[-1,2,3].min

1 < 2
-1 < 1

defined?(1 + 1)

1 + 1 + 1
1 * 2
3 * 4
6 / 3
1.0 + 3
3.0 + 1.0
3.0 * 6
1.0 / 2
1.0 / 0
1.0 % 0.5
5 % 3
16 % 9
12.349 % 1

2 <=> 3
1 === 0.0

!false

if (1+1 == 0)
    do_something
elsif (1+1 == 2)
    do_something_else
else
    do_nothing
end
```

</details>

<details>
<summary>ast</summary>

```ruby
s(:block,
 s(:if, s(:true), nil, s(:lasgn, :x, s(:lit, 1))),
 s(:call, s(:lit, 1), :&, s(:lit, 3)),
 s(:call, s(:lit, 2), :|, s(:lit, 1)),
 s(:call, s(:lit, 3), :^, s(:lit, 2)),
 s(:call, s(:lit, 1), :<<, s(:lit, 4)),
 s(:call, s(:lit, 32), :>>, s(:lit, 3)),
 s(:call, s(:array, s(:lit, 1), s(:lit, 2), s(:lit, 3)), :max),
 s(:call, s(:array, s(:lit, -1), s(:lit, 2), s(:lit, 3)), :min),
 s(:call, s(:lit, 1), :<, s(:lit, 2)),
 s(:call, s(:lit, -1), :<, s(:lit, 1)),
 s(:defined, s(:call, s(:lit, 1), :+, s(:lit, 1))),
 s(:call, s(:call, s(:lit, 1), :+, s(:lit, 1)), :+, s(:lit, 1)),
 s(:call, s(:lit, 1), :*, s(:lit, 2)),
 s(:call, s(:lit, 3), :*, s(:lit, 4)),
 s(:call, s(:lit, 6), :/, s(:lit, 3)),
 s(:call, s(:lit, 1.0), :+, s(:lit, 3)),
 s(:call, s(:lit, 3.0), :+, s(:lit, 1.0)),
 s(:call, s(:lit, 3.0), :*, s(:lit, 6)),
 s(:call, s(:lit, 1.0), :/, s(:lit, 2)),
 s(:call, s(:lit, 1.0), :/, s(:lit, 0)),
 s(:call, s(:lit, 1.0), :%, s(:lit, 0.5)),
 s(:call, s(:lit, 5), :%, s(:lit, 3)),
 s(:call, s(:lit, 16), :%, s(:lit, 9)),
 s(:call, s(:lit, 12.349), :%, s(:lit, 1)),
 s(:call, s(:lit, 2), :<=>, s(:lit, 3)),
 s(:call, s(:lit, 1), :===, s(:lit, 0.0)),
 s(:call, s(:false), :!),
 s(:if,
  s(:call, s(:call, s(:lit, 1), :+, s(:lit, 1)), :==, s(:lit, 0)),
  s(:call, nil, :do_something),
  s(:if,
   s(:call, s(:call, s(:lit, 1), :+, s(:lit, 1)), :==, s(:lit, 2)),
   s(:call, nil, :do_something_else),
   s(:call, nil, :do_nothing))))
```

</details>

<details>
<summary>output</summary>

```ruby
s(:block,
 s(:if, s(:true), nil, s(:lasgn, :x, s(:lit, 1))),
 s(:lit, 1),
 s(:lit, 3),
 s(:lit, 1),
 s(:lit, 16),
 s(:lit, 4),
 s(:lit, 3),
 s(:lit, -1),
 s(:true),
 s(:true),
 s(:defined, s(:call, s(:lit, 1), :+, s(:lit, 1))),
 s(:lit, 3),
 s(:lit, 2),
 s(:lit, 12),
 s(:lit, 2),
 s(:lit, 4.0),
 s(:lit, 4.0),
 s(:lit, 18.0),
 s(:lit, 0.5),
 s(:call, s(:lit, 1.0), :/, s(:lit, 0)),
 s(:lit, 0.0),
 s(:lit, 2),
 s(:lit, 7),
 s(:lit, 0.3490000000000002),
 s(:lit, -1),
 s(:false),
 s(:true),
 s(:call, nil, :do_something_else))
```

</details>

Some intermediaries seem to be weirdly rounded
This optimization might also be applied to String arithmetic

In the future the ability to lookup global constants or freshly stored variables would make this optimization pass even more powerful

Please check if the ruby code is up to the standard, these are pretty much my first 100 lines ruby

### DIFF
--- a/lib/natalie/compiler/pass0c.rb
+++ b/lib/natalie/compiler/pass0c.rb
@@ -40,9 +40,6 @@ module Natalie
       def type_is_always_falsey(arg)
         return [:false, :nil].include?(arg)
       end
-      def type_is_const_evaluable(arg)
-        return [:true, :str, :lit, :array, :false, :nil].include?(arg)
-      end
 
       def process_if(exp, block_with_args=nil)
         exp = exp.new(*exp.map { |e| process_atom(e) })
@@ -98,46 +95,18 @@ module Natalie
             class_arg2 = args[0][1].class
             if [Integer, Float].include?(class_arg1) && [Integer, Float].include?(class_arg2)
               case method
-              when :+
-                return canary_const_addition(exp, values)
-              when :-
-                return canary_const_subtraction(exp, values)
-              when :*
-                return canary_const_multiplication(exp, values)
-              when :/
-                # FIXME: We could propagate NaNs and INFINITIES in the float case
-                return canary_const_division(exp, values) unless args[0][1] == 0
-              when :%
-                return canary_const_modulo(exp, values) unless args[0][1] == 0
+              when :+, :-, :*, :<=>
+                return canary_const_op(exp, values)
+              when :/, :%
+                # FIXME: We could propagate NaNs and INFINITIES in the float-division case
+                return canary_const_op(exp, values) unless args[0][1] == 0
               # FIXME: This one makes String#* timeout
               # when :**
-              #   return canary_const_exp(exp, values) unless values[0] < 0
-              when :&
-                return canary_const_and(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
-              when :|
-                return canary_const_or(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
-              when :^
-                return canary_const_xor(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
-              when :<<
-                return canary_const_shl(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
-              when :>>
-                return canary_const_shr(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
-              when :==
-                return canary_const_eq(exp, values)
-              when :!=
-                return canary_const_neq(exp,values)
-              when :>
-                return canary_const_gt(exp, values)
-              when :>=
-                return canary_const_geq(exp, values)
-              when :<
-                return canary_const_lt(exp, values)
-              when :<=
-                return canary_const_leq(exp, values)
-              when :===
-                return canary_const_eqeq(exp, values)
-              when :<=>
-                return canary_const_cmp(exp, values)
+              #   return canary_const_op(exp, values) unless values[0] < 0
+              when :&, :|, :^, :<<, :>>
+                return canary_const_op(exp, values) if [class_arg1, class_arg2].all? {|x| x == Integer}
+              when :==, :!=, :>, :>=, :<, :<=, :===
+                return canary_const_bool_op(exp, values)
               end
             end
           elsif args.length == 0
@@ -164,89 +133,14 @@ module Natalie
         return exp.new(:if, exp.new(:call, first, :>, second), canary_array_max(exp, [first, *rest]), canary_array_max(exp, [second, *rest]))
       end
 
-      # FIXME: These are pretty repetitive, maybe do these with meta programming
-      def canary_const_addition(exp, values)
-        _, (_, summand1), op, (_, summand2) = exp
-        return s(:lit, summand1 + summand2)
-      end
-      def canary_const_subtraction(exp, values)
-        _, (_, minuend), op, (_, subtrahend) = exp
-        return s(:lit, minuend - subtrahend)
-      end
-      def canary_const_multiplication(exp, values)
-        _, (_, factor1), op, (_, factor2) = exp
-        return s(:lit, factor1 * factor2)
-      end
-      def canary_const_division(exp, values)
-        _, (_, numerator), op, (_, denominator) = exp
-        return s(:lit, numerator / denominator)
-      end
-      def canary_const_modulo(exp, values)
-        _, (_, numerator), op, (_, denominator) = exp
-        return s(:lit, numerator % denominator)
-      end
-      def canary_const_exp(exp, values)
-        _, (_, base), op, (_, power) = exp
-        return s(:lit, base ** power)
-      end
-      def canary_const_and(exp, values)
+      def canary_const_op(exp, values)
         _, (_, arg1), op, (_, arg2) = exp
-        return s(:lit, arg1 & arg2)
+        return exp.new(:lit, arg1.send(op, arg2))
       end
-      def canary_const_or(exp, values)
+      def canary_const_bool_op(exp,values)
         _, (_, arg1), op, (_, arg2) = exp
-        return s(:lit, arg1 | arg2)
-      end
-      def canary_const_xor(exp, values)
-        _, (_, arg1), op, (_, arg2) = exp
-        return s(:lit, arg1 ^ arg2)
-      end
-      def canary_const_shl(exp, values)
-        _, (_, arg1), op, (_, arg2) = exp
-        return s(:lit, arg1 << arg2)
-      end
-      def canary_const_shr(exp, values)
-        _, (_, arg1), op, (_, arg2) = exp
-        return s(:lit, arg1 >> arg2)
-      end
-      def canary_const_eq(exp,values)
-        _, (_, arg1), leq, (_, arg2) = exp
-        return s(:true) if arg1 == arg2
-        return s(:false)
-      end
-      def canary_const_neq(exp,values)
-        _, (_, arg1), leq, (_, arg2) = exp
-        return s(:true) if arg1 != arg2
-        return s(:false)
-      end
-      def canary_const_gt(exp,values)
-        _, (_, arg1), leq, (_, arg2) = exp
-        return s(:true) if arg1 > arg2
-        return s(:false)
-      end
-      def canary_const_geq(exp,values)
-        _, (_, arg1), leq, (_, arg2) = exp
-        return s(:true) if arg1 >= arg2
-        return s(:false)
-      end
-      def canary_const_lt(exp,values)
-        _, (_,arg1), leq, (_,arg2) = exp
-        return s(:true) if arg1 < arg2
-        return s(:false)
-      end
-      def canary_const_leq(exp,values)
-        _, (_, arg1), leq, (_, arg2) = exp
-        return s(:true) if arg1 <= arg2
-        return s(:false)
-      end
-      def canary_const_eqeq(exp,values)
-        _, (_, arg1), eqeq, (_, arg2) = exp
-        return s(:true) if arg1 === arg2
-        return s(:false)
-      end
-      def canary_const_cmp(exp,values)
-        _, (_, arg1), cmp, (_, arg2) = exp
-        return s(:lit, arg1 <=> arg2)
+        return exp.new(:true) if arg1.send(op, arg2)
+        return exp.new(:false)
       end
 
       def canary_const_not(exp, values)
--- a/lib/natalie/compiler/pass0c.rb
+++ b/lib/natalie/compiler/pass0c.rb
@@ -2,7 +2,7 @@ require_relative './base_pass'
 
 module Natalie
   class Compiler
-    # A 'canary' pass that tries to simplify well-known ruby pattern which would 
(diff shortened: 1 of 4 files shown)


lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
         )
       end
 
+      def process_defined(exp, block_with_args=nil)
+        @in_defined_statement = true
+        exp = exp.new(*exp.map {|sub_exp| sub_exp.is_a?(Sexp) ? process(sub_exp) : sub_exp})
+        @in_defined_statement = false
+        return exp
+      end
```
Hendiadyoin1, line 33: This might fail on nested `defined?` calls, but who would do such a thing

lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
         )
       end
 
+      def process_defined(exp, block_with_args=nil)
+        @in_defined_statement = true
+        exp = exp.new(*exp.map {|sub_exp| sub_exp.is_a?(Sexp) ? process(sub_exp) : sub_exp})
```
seven1m, line 30: The BasePass has `process_atom` which should work:

```suggestion
        exp = exp.new(*exp.map { |e| process_atom(e) })
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
...
+        @in_defined_statement = false
+        return exp
+      end
+
+      def type_is_always_truthy(arg)
+        # FIXME: Are there any other types that are always truthy?
+        return [:true,:str, :lit, :array].include?(arg)
```
seven1m, line 37: missing space nitpick:

```suggestion
        return [:true, :str, :lit, :array].include?(arg)
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
...
+      end
+      def type_is_always_falsy(arg)
+        return [:false, :nil].include?(arg)
+      end
+      def type_is_const_evaluatable(arg)
+        return [:true,:str, :lit, :array, :false, :nil].include?(arg)
```
seven1m, line 43:
```suggestion
        return [:true, :str, :lit, :array, :false, :nil].include?(arg)
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
...
+      end
+
+      def process_if(exp, block_with_args=nil)
+        exp = exp.new(*exp.map {|sub_exp| sub_exp.is_a?(Sexp) ? process(sub_exp) : sub_exp})
+        if exp.flatten.include?(:lasgn)
+          # This is required for now beacause
```
seven1m, line 49:
```suggestion
          # This is required for now because
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -24,7 +25,62 @@ def process_iter(exp)
...
+
+        return exp
+      end
+
+      def process_and(exp, block_with_args=nil)
+        exp = exp.new(*exp.map {|sub_exp| sub_exp.is_a?(Sexp) ? process(sub_exp) : sub_exp})
+        _, (t1, *arg1), (t2, *arg2) = exp
+        return s(t1, *arg1) if type_is_always_falsy(t1)
+        return exp          if !type_is_always_truthy(t1) # t1 is not const-evaluatable
+        # t2 will be returned even if it is falsy, so we donot have to check it
```
seven1m, line 78:
```suggestion
        # t2 will be returned even if it is falsey, so we don't have to check it
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -34,9 +90,63 @@ def process_call(exp, block_with_args=nil)
             elsif method == :max && args.empty? && !block_with_args
               return process(canary_array_max(exp, values))
             end
+          elsif t == :lit && args.length == 1 && args[0][0] == :lit && !@in_defined_statement
+            # we cannot collapse these in defined? statements, because it cahnges the type
```
seven1m, line 94:
```suggestion
            # we cannot collapse these in defined? statements, because it changes the type
```

lib/natalie/compiler/pass0c.rb
```diff
@@ -34,9 +90,63 @@ def process_call(exp, block_with_args=nil)
...
+            if [Integer, Float].include?(class_arg1) && [Integer, Float].include?(class_arg2)
+              case method
+              when :+
+                return canary_const_addition(exp, values)
+              when :-
+                return canary_const_subtraction(exp, values)
+              when :*
+                return canary_const_multiplication(exp, values)
+              when :/
+                # FIXME: We could propagade NaNs and INFINITIES in the float case
```
seven1m, line 107:
```suggestion
                # FIXME: We could propagate NaNs and INFINITIES in the float case
```

seven1m (changes requested):
Pretty cool! Just a few small things, mostly typos to fix.

lib/natalie/compiler/pass0c.rb
seven1m: This pattern is repeated in a few places btw.

seven1m (approved):
Nice! I'm going to let @TheGrizzlyDev take a look also, since he has his head in this area already.

lib/natalie/compiler/pass0c.rb
```diff
@@ -20,11 +21,66 @@ def process_iter(exp)
...
       end
 
+      def process_defined(exp, block_with_args=nil)
+        @in_defined_statement = true
+        exp = exp.new(*exp.map { |e| process_atom(e) })
+        @in_defined_statement = false
+        return exp
+      end
+
+      def type_is_always_truthy(arg)
```
TheGrizzlyDev, line 35: could this not be `!type_is_always_falsey` instead?

lib/natalie/compiler/pass0c.rb
```diff
@@ -20,11 +21,66 @@ def process_iter(exp)
...
+        # FIXME: Are there any other types that are always truthy?
+        return [:true, :str, :lit, :array].include?(arg)
+      end
+      def type_is_always_falsey(arg)
+        return [:false, :nil].include?(arg)
+      end
+      def type_is_const_evaluable(arg)
```
TheGrizzlyDev, line 42: is this used? If so, what does it mean?

lib/natalie/compiler/pass0c.rb
```diff
@@ -53,6 +163,97 @@ def canary_array_max(exp, values)
         return exp.new(:if, exp.new(:call, first, :>, second), canary_array_max(exp, [first, *rest]), canary_array_max(exp, [second, *rest]))
       end
 
+      # FIXME: These are pretty repetitive, maybe do these with meta programming
```
TheGrizzlyDev, line 166: How about `canary_const_op(exp, values, op)` where op is a symbol and you use send on the first operand?

It'd look something 

```ruby
def canary_const_op(exp, values)
    _, (_, arg1), op, (_, arg2) = exp
   arg1.send op arg2
end

```

TheGrizzlyDev (commented):
Gave it a quick look and left a few comments, I'll give it a more thorough review tomorrow :)

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: So similar reasoning to the next question
there are types we can say, that they are always truthy or falsey,
but we cannot say the same about variables, call results and the like (yet)
so I just made these helpers with types I know will work fine

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: sure I will try that

lib/natalie/compiler/pass1.rb
```diff
@@ -139,10 +139,14 @@ def process_case(exp)
...
             when_body = [s(:nil)] if when_body == [nil]
-            matchers.each do |matcher|
-              cond << s(:is_truthy, s(:public_send, process(matcher), s(:intern, '==='), s(:args, value_name), 'nullptr'))
-              cond << s(:block, *when_body)
-            end
+            full_condition = s(
+              :is_truthy,
+              # FIXME: Get rid of the :nil here
+              # FIXME: This is still a bit hacky
+              matchers.reduce(s(:nil)) {|new_cond, matcher| process_or(s(:or, new_cond, s(:public_send, process(matcher), s(:intern, '==='), s(:args, value_name), 'nullptr'))) }
```
Hendiadyoin1, line 146: A way to represent a just boolean `or` might be nice here

lib/natalie/compiler/pass0c.rb
seven1m: Yeah, you cannot just do `!type_is_always_falsey` for something like `:lvar`.

lib/natalie/compiler/pass1.rb
seven1m: What's the motivation for this change? I think it's harder to read what's going on, but if there's a performance boost with it, I could see how it could be worth the change...

seven1m: Nevermind, I see in Discord you are fixing a bug where we compile lambdas inside the `when_body` twice. Carry on!

seven1m: Please add a test to prove this is fixed. I would hate to regress on that.

seven1m: I added a test for it. :hugs:

Hendiadyoin1: Thanks!

Hendiadyoin1:
Just fixed one small oversight of mine in the compiler self-test up
are there any more things to be done?

lib/natalie/compiler/pass0c.rb
```diff
@@ -53,6 +132,22 @@ def canary_array_max(exp, values)
...
+      end
+      def canary_const_bool_op(exp,values)
+        _, (_, arg1), op, (_, arg2) = exp
+        return s(:true) if arg1.send(op, arg2)
+        return s(:false)
+      end
+
+      def canary_const_not(exp, values)
+        _, (type,*), _not = exp
+        return s(:false) if type_is_always_truthy(type)
```
TheGrizzlyDev, line 147: When using 's(...)' you may want to use instead 'exp.new(...)' as that will propagate the original source metadata (file, line)

TheGrizzlyDev:
> Just fixed one small oversight of mine in the compiler self-test up are there any more things to be done?

Just a single thing I've just noticed, the rest looks pretty lit :+1:

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: ahh ok didn't know that
will change that accordingly

lib/natalie/compiler/pass0c.rb
TheGrizzlyDev: yeah, sounds unlikely, but just in case, it might be worth adding a comment :D

seven1m:
What a fun addition!

PR #345, opened 2022-01-02
Make Integer#** spec compliant

richardboehme:
This was some fun 😅 

There are some points to take a closer look here:
* Ruby does not allow pow-ing that results in *really* large numbers. There seems to be some threshold related to the size the resulting bigint would need. I tried to replicate the behavior but the boundary is different between natalie and MRI for now.
* I'm not sure if the constant definition in natalie/constants.hpp is ok? I wanted to use constexpr but I can not because sqrt is not a constexpr... Please correct me if we should use something else.
* BigInt's division algorithm seems to be doing something different that what MRI does. I did not want to investigate further for this PR but we might want to look at this (basically `-100000000000000000000000000000000000000000000000000 / 10000000000000000000000000000000000000001` returns `-9999999999` but it should be `-10000000000`)

[#201]

### DIFF
--- a/include/natalie/constants.hpp
+++ b/include/natalie/constants.hpp
@@ -1,5 +1,8 @@
 // This is used at build time as well in order to generate numbers.rb
 #include <limits.h>
+#include <math.h>
 
 #define NAT_MAX_FIXNUM LLONG_MAX
 #define NAT_MIN_FIXNUM LLONG_MIN + 1
+
+const long long NAT_MAX_FIXNUM_SQRT = (long long)sqrt((double)NAT_MAX_FIXNUM);
--- a/include/natalie/constants.hpp
+++ b/include/natalie/constants.hpp
@@ -1,8 +1,26 @@
 // This is used at build time as well in order to generate numbers.rb
+#pragma once
+#include "natalie/types.hpp"
 #include <limits.h>
-#include <math.h>
 
 #define NAT_MAX_FIXNUM LLONG_MAX
 #define NAT_MIN_FIXNUM LLONG_MIN + 1
 
-const long long NAT_MAX_FIXNUM_SQRT = (long long)sqrt((double)NAT_MAX_FIXNUM);
+namespace Natalie {
+
+constexpr nat_int_t lowest_squarable_fixnum() {
+    nat_int_t mask = (nat_int_t)1 << (sizeof(nat_int_t) * 8 - 2);
+    nat_int_t max_sqrt = 0;
+    while (mask > 0) {
+        nat_int_t current_sum = max_sqrt + mask;
+        mask >>= 1;
+        if ((NAT_MAX_FIXNUM / current_sum) < current_sum) continue; // overflow
+        max_sqrt = current_sum;
+    }
+
+    return max_sqrt;
+}
+
+constexpr nat_int_t NAT_MAX_FIXNUM_SQRT = lowest_squarable_fixnum();
+
+}
--- a/include/natalie/integer_object.hpp
+++ b/include/natalie/integer_object.hpp
@@ -54,7 +54,7 @@ public:
     virtual Value mul(Env *, Value);
     virtual Value div(Env *, Value);
     Value mod(Env *, Value) const;
-    Value pow(Env *, Value) const;
+    virtual Value pow(Env *, Value);
     Value cmp(Env *, Value);
     virtual Value times(Env *, Block *);
     Value bitwise_and(Env *, Value) const;
--- a/include/natalie/integer_object.hpp
+++ b/include/natalie/integer_object.hpp
@@ -5,6 +5,7 @@
 #include <inttypes.h>
 
 #include "natalie/class_object.hpp"
+#include "natalie/constants.hpp"
 #include "natalie/forward.hpp"
 #include "natalie/global_env.hpp"
 #include "natalie/macros.hpp"
--- a/src/integer_object.cpp
+++ b/src/integer_object.cpp
@@ -129,17 +129,25 @@ Value IntegerObject::sub(Env *env, Value arg) {
     return sub_fast(m_integer, other->to_nat_int_t());
 }
 
-Value mul_fast(nat_int_t a, nat_int_t b) {
-    if (a == 0 || b == 0)
-        return Value::integer(0);
+bool will_multiplication_overflow(nat_int_t a, nat_int_t b) {
+    if (a == b) {
+        return a > NAT_MAX_FIXNUM_SQRT || a < -NAT_MAX_FIXNUM_SQRT;
+    }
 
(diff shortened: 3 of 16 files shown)


ryangjchandler:
This is awesome, I've been working on `Integer#pow` so should help me a tad.

include/natalie/constants.hpp
```diff
@@ -1,5 +1,8 @@
 // This is used at build time as well in order to generate numbers.rb
 #include <limits.h>
+#include <math.h>
 
 #define NAT_MAX_FIXNUM LLONG_MAX
-#define NAT_MIN_FIXNUM LLONG_MIN + 1
\ No newline at end of file
+#define NAT_MIN_FIXNUM LLONG_MIN + 1
+
+const long long NAT_MAX_FIXNUM_SQRT = (long long)sqrt((double)NAT_MAX_FIXNUM);
```
richardboehme, line 8: Hm clang does not like this (me neither). There does not seem to be a way to define a compile time sqrt without having a custom sqrt function?

Hendiadyoin1: GCC allows this,
Clang might allow it with `-fno-math-errno` set, but not sure
also the arguments to sqrt are always* doubles, so that cast seems redundant

*: In some libcs it can also do float->float and Integral-> double implicitly without the any casting or suffix'
[see here](https://en.cppreference.com/w/cpp/numeric/math/sqrt)

richardboehme: I've just added the cast after the first test run failed, where clang complained about NAT_MAX_FIXNUM not being a double but a long long. Not sure if we want to add this flag.. Maybe we can find another way around this...

richardboehme: @seven1m What do you think we should do with this? I guess we have the following options:

* Try this `-fno-math-errno` flag for clang that @Hendiadyoin1 suggested
* Remove this constant (included the optimization I added in Integer multiplication overflow checks)
* Implement a custom sqrt function that is a defined as constexpr
* Calculate the sqrt ourselves (3037000499) and put a constant value into the macro (this would mean we lose the portability of using LLONG_MAX)

Maybe there are even other ways around it? I'm not really a C++ dev so I have no clue if there are other options around.

TheGrizzlyDev: what if we take a slightly different approach? 

---- Before you continue reading this, I might be saying something extremely dumb, in which case just let me know :D 

What we need is not the sqrt of the max value, but rather the last number that when elevated to the pow of 2 will not overflow. This can be prechecked and runs reasonably fast, in fact you only need sizeof(nat_int_t) (or whatever our fixnum type is) iterations to calculate that.

If you start from the highest bits you can check if setting that bit to 1 will make the pow^2 of it overflow, if not then set it to 1, otherwise leave it to 0, step onto the next lower bit and repeat until you reach the first digit. Something like that would be:
- easy to implement
- fast
- easy to read 

and thus we could run it as a constexpr without making the compilation any slower (not that I would mind anyway, I mostly like the fact that it is somewhat simple to read/write in this case).

What do you think?

richardboehme: Just for my understanding: The calculation you described would result in the sqrt of LLONG_MAX right? It's the same number because the sqrt is the last number that can be squared without overflowing.

I guess this would work though. How would you do the overflow check? Just using the overflow check we use normally?

TheGrizzlyDev: >  The calculation you described would result in the sqrt of LLONG_MAX right?
it'd get you the closest integer number there is to it yeah :) 

> How would you do the overflow check? 
I'd write it something like this:

```c++
nat_int_t mask = 1 << (sizeof(nat_int_t) - 1); 
nat_int_t max_sqrt = 0;

while (mask > 0) {
   nat_int_t current_sum = max_sqrt + mask;
   if ((NAT_MAN_FIXNUM / current_sum) >= current_sum) continue; // overflow
   max_sqrt = current_sum;
   mask >>= 1;
}

return max_sqrt;

```

wdyt @richardboehme ?

richardboehme: Looks great! Could you have an eye on the small changes I needed to do to your code snippet?

TheGrizzlyDev: Looks good!

include/natalie/integer_object.hpp
```diff
@@ -13,6 +14,21 @@
...
+        nat_int_t current_sum = max_sqrt + mask;
+        mask >>= 1;
+        if ((NAT_MAX_FIXNUM / current_sum) < current_sum) continue; // overflow
+        max_sqrt = current_sum;
+    }
+
+    return max_sqrt;
+}
+
+constexpr long long NAT_MAX_FIXNUM_SQRT = lowest_squarable_fixnum();
```
TheGrizzlyDev, line 30: Why not adding this in constants?

src/integer_object.cpp
```diff
@@ -223,14 +230,84 @@ Value IntegerObject::mod(Env *env, Value arg) const {
...
+
+    nat_int_t result = 1;
+
+    bool negative = a < 0;
+    if (negative) a = -a;
+    negative = negative && (b % 2);
+
+    while (b != 0) {
+        while (b % 2 == 0) {
+            if (will_multiplication_overflow(a, a)) {
```
TheGrizzlyDev, line 261: can you just check instead if a > NAT_MAX_FIXNUM_SQRT?

TheGrizzlyDev (approved):
very cool! Left a few comments, let me know if you need some clarification ;)

include/natalie/integer_object.hpp
richardboehme: I did not want to include all the stuff necessary to use nat_int_t in constants.hpp. Also it is not needed for numbers.rb so I thought it might be better to move it. Do you think we should embrace constants.hpp including all the natalie stuff?

(I see I should make NAT_MAX_FIXNUM_SQRT a nat_int_t as well)

TheGrizzlyDev: I am not that strongly for it, I just think it's kind of nice to keep everything in one place, but I am totally cool if we separate them. I also don't really see any use for it outside of nat's runtime, which otherwise would have been a big point for putting it into constants.hpp. I'll leave it to you :D

src/integer_object.cpp
richardboehme: Yeah I guess we could do that... It would probably be more performant, right? 

My intentation was to make all overflow checks use the same method which is why I integrated this check to will_multiplication_overflow instead.

TheGrizzlyDev: yeah, that'd be way faster, but I still like that you separated that overflow check like that :+1:

richardboehme: I changed both things you commented on. Thanks for the suggested changes 🤗

seven1m:
Nice work friends.

PR #346, opened 2022-01-02
Partial `String#length` & `String#size` spec compliance

borisromanov:
Adds spec files for `String#length` and `String#size`, and tweaks `StringObject::next_char`.

`StringObject::next_char` was being called by `StringObject::size`, after adding the size and length specs, I noticed the UTF-8 spec was failing. Using lldb, I noticed the while loop in `::size` was exiting early prior to completely processing the string due to index being incremented when processing UTF-8 strings.

The shared `length.rb` specs make calls to, `String#encode`, with encodings the project currently does not support. As well as, `String#bytesize`, which currently is unmapped in `binding_gen.rb`. I avoided implementing the various encodings being called in the shared spec, I feel like I'm way to naive to tackle that at the moment. However, I hope to look into implementing a spec compliant `#bytesize` next. Therefore,  I `x'd` out those specs to avoid errors, but it's been a while since I've contributed and do not remember if that's best practice. Please let me know what y'all prefer here.

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -32,10 +32,10 @@ char *StringObject::next_char(Env *env, char *buffer, size_t *index) {
         buffer[0] = str[i];
         if (((unsigned char)buffer[0] >> 3) == 30) { // 11110xxx, 4 bytes
             if (i + 3 >= len) raise_encoding_invalid_byte_sequence_error(env, i);
-            buffer[1] = str[++i];
-            buffer[2] = str[++i];
-            buffer[3] = str[++i];
-            buffer[4] = 0;
+            for (size_t j = 1; j < 5; j++)
+            {
+                buffer[j] = str[j];
+            }
         } else if (((unsigned char)buffer[0] >> 4) == 14) { // 1110xxxx, 3 bytes
             if (i + 2 >= len) raise_encoding_invalid_byte_sequence_error(env, i);
             buffer[1] = str[++i];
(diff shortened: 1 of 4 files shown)


src/string_object.cpp
```diff
@@ -32,10 +32,10 @@ char *StringObject::next_char(Env *env, char *buffer, size_t *index) {
...
         if (((unsigned char)buffer[0] >> 3) == 30) { // 11110xxx, 4 bytes
             if (i + 3 >= len) raise_encoding_invalid_byte_sequence_error(env, i);
-            buffer[1] = str[++i];
-            buffer[2] = str[++i];
-            buffer[3] = str[++i];
-            buffer[4] = 0;
+            for (size_t j = 1; j < 5; j++)
+            {
+                buffer[j] = str[j];
+            }
```
seven1m, line 38: The point of next_char() is to (1) fill the `buffer` with the bytes of the character, and (2) increment the `index` (pointer to) integer to the next byte position. Your change here makes it so that no longer happens.

I'm sure there are bugs here -- we haven't exposed this code to specs yet. But the code change here doesn't seem right to me.

seven1m:
@borisromanov is fine to skip specs in order to make progress -- that's fine!

seven1m:
Sorry @borisromanov! Didn't mean to be discouraging. 😬 We can discuss in Discord later if you want to tackle this again.

borisromanov:
@seven1m, no hard feelings. I'll try picking this up some other time. Didn't want to leave a draft like this lingering in the queue.

PR #348, opened 2022-01-02
Fix error messages that may reference anonymous classes

FnControlOption:
Error using `class_name_or_blank`:

```rb
anonymous_class = Class.new
'string'.force_encoding anonymous_class.new # => TypeError (no implicit conversion of  into String)
```

Error using `inspect_str`:

```rb
anonymous_class = Class.new
'string'.force_encoding anonymous_class.new # => TypeError (no implicit conversion of #<Class:0x7fcf44827020> into String)
```


seven1m:
Good fix!

PR #356, opened 2022-01-04
Make Fixnum and Bignum deprecated constants

richardboehme:
This PR also adds support for deprecated constants by implementing `Module#deprecated_constant` and improves our `warn` method on the Env class to match ruby's behavior with warnings.

### DIFF
--- a/include/natalie/constant.hpp
+++ b/include/natalie/constant.hpp
@@ -1,0 +1,29 @@
+#pragma once
+#include "natalie/forward.hpp"
+#include "natalie/value.hpp"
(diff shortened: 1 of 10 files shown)


richardboehme:
> @richardboehme would it be possible to implement this same functionality without adding a new hashmap? Could we make a new wrapper for constants that stores an `m_deprecated` member? That way, we only have to do a single hashmap lookup rather than two for each constant.

I tried implementing this. What do you think?

seven1m:
Nice!

PR #357, opened 2022-01-04
Fix variable names in IncludeExpectation methods

timcraft:
These lines should be referencing the local variable `value` not the instance variable `@value`


seven1m:
Whoops, good catch!

PR #358, opened 2022-01-04
Implement Kernel#instance_variable_defined?

timcraft:


src/kernel_module.cpp
```diff
@@ -239,6 +239,14 @@ Value KernelModule::main_obj_inspect(Env *env) {
     return new StringObject { "main" };
 }
 
+Value KernelModule::instance_variable_defined(Env *env, Value name_val) {
+    if (is_nil() || is_boolean() || is_integer() || is_float() || is_symbol()) {
```
seven1m, line 243: I assume we'll need this same string of checks for `instance_variable_get` et al too, yes? If you work on that, we should extract this to a common `Value::can_have_instance_variables()` or similar.

timcraft: I've looked into that today whilst making instance_variable_get spec compliant and I think `is_frozen()` should be sufficient for both methods—I don't know if it's the correct implementation of "can have instance variables", but it more closely matches the `assert_not_frozen` call in `instance_variable_set` and passes both specs

seven1m: Sweet.

PR #359, opened 2022-01-04
Make `Module#private` spec-compliant

FnControlOption:
For example:

```rb
class A
  def foo
    "bar"
  end
end

class B < A
  private :foo
end

A.new.foo # => "bar"
B.new.foo # => NoMethodError
B.remove_method :foo
B.new.foo # => "bar"
```


seven1m:
This kinda blew my mind. Good work. I don't love that we need yet another hashmap, but I don't have a better solution either.

PR #364, opened 2022-01-05
Prevent temporary "new" value creation in variable declaration in Pass4

Hendiadyoin1:
Just something small to get rid of some variables that are immediately reassigned to another one.

There are probably more instances of this behavior,
like when creating or assigning a local-variable.


seven1m:
Hey cool!

PR #367, opened 2022-01-06
Make `Kernel#send` spec-compliant

FnControlOption:
From [Calling Methods ¶ Method Lookup](https://ruby-doc.org/core-3.0.0/doc/syntax/calling_methods_rdoc.html#label-Method+Lookup):

> Here is the order of method lookup for the receiver's class or module `R`:
> 
> - The prepended modules of `R` in reverse order
> - For a matching method in `R`
> - The included modules of `R` in reverse order
> 
> If `R` is a class with a superclass, this is repeated with `R`'s superclass until a method is found.

On Natalie's `master` branch, included modules are actually looked up in _forward_ order. This can be fixed by adding them to `m_included_modules` at the index that is right after the class into which they are included (instead of adding them at the end of the list).

Additionally, the `after_method` parameter of `find_method` only keeps its change to `nullptr` in the current method scope (or recursive `find_method` calls), but not across method scopes that are at the same level in the class hierarchy. This can be fixed by using a double pointer:

`Method *find_method(..., Method *after_method)` | `Method *find_method(..., Method **after_method)`
--- | ---
graphviz (0) | graphviz (1)

### DIFF
--- a/include/natalie/module_object.hpp
+++ b/include/natalie/module_object.hpp
@@ -90,7 +90,8 @@ public:
     void set_method_visibility(Env *, SymbolObject *, MethodVisibility);
     MethodVisibility get_method_visibility(Env *, SymbolObject *);
     MethodInfo *find_method_info(Env *, SymbolObject *);
-    Method *find_method(Env *, SymbolObject *, ModuleObject ** = nullptr, Method * = nullptr) const;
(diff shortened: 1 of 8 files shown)


seven1m:
Sick! 🥰

PR #369, opened 2022-01-06
Implement Symbol#empty?

timcraft:


seven1m:
We can save a GC allocation by using `strlen(m_name) == 0`.

PR #374, opened 2022-01-06
Implement Range#size

FnControlOption:


seven1m:
Nice use of ruby here. 😃

PR #376, opened 2022-01-07
Make boolean methods return bool instead of Value

FnControlOption:


seven1m:
Oh boy I've been wanting to do this for a long time. I'll merge once the tests are passing. 🤗

PR #378, opened 2022-01-07
Implement String#include?

FnControlOption:
[#217]

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -667,6 +667,14 @@ Value StringObject::split(Env *env, Value splitter, Value max_count_value) {
     }
 }
 
+bool StringObject::include(Env *env, Value arg) {
+    auto to_str = "to_str"_s;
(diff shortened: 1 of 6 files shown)


src/string_object.cpp
```diff
@@ -657,6 +657,14 @@ Value StringObject::split(Env *env, Value splitter, Value max_count_value) {
     }
 }
 
+bool StringObject::include(Env *env, Value arg) {
+    auto to_str = "to_str"_s;
+    if (!arg->is_string() && arg->respond_to(env, to_str))
+        arg = arg->send(env, to_str);
+    arg->assert_type(env, Object::Type::String, "String");
+    return ::strstr(c_str(), arg->as_string()->c_str()) != nullptr;
```
seven1m, line 665: I don't think we can use `strstr` (or even `strnstr`) here because it's possible to generate a Ruby string containing a null character in the middle:

```ruby
"foo\x00bar".include?("bar") # => true
````

PR #385, opened 2022-01-07
Make Range#inspect spec compliant

timcraft:
Currently it doesn't handle a nil begin value:

    $ bin/natalie
    nat> (..3).inspect
    "nil..3"

    $ irb
    irb(main):001:0> (..3).inspect
    => "..3"

I've re-implemented `RangeObject::inspect` to use append instead of `StringObject::format` in order to avoid nesting the ifs and eliminate the duplication between them.

### DIFF
--- a/src/range_object.cpp
+++ b/src/range_object.cpp
@@ -92,19 +92,19 @@ Value RangeObject::first(Env *env, Value n) {
 }
 
 Value RangeObject::inspect(Env *env) {
-    if (m_exclude_end) {
-        if (m_end->is_nil()) {
-            return StringObject::format(env, "{}...", m_begin->inspect_str(env));
-        } else {
-            return StringObject::format(env, "{}...{}", m_begin->inspect_str(env), m_end->inspect_str(env));
-        }
-    } else {
-        if (m_end->is_nil()) {
-            return StringObject::format(env, "{}..", m_begin->inspect_str(env));
-        } else {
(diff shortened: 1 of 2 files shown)


src/range_object.cpp
```diff
@@ -92,19 +92,19 @@ Value RangeObject::first(Env *env, Value n) {
...
-            return StringObject::format(env, "{}..", m_begin->inspect_str(env));
-        } else {
-            return StringObject::format(env, "{}..{}", m_begin->inspect_str(env), m_end->inspect_str(env));
-        }
+        string->append(env, "..");
+    }
+    if (!m_end->is_nil() || m_begin->is_nil()) {
+        string->append(env, m_end->inspect_str(env));
     }
+    return string;
```
seven1m, line 107: I don't know if it's just me, but I'm having an incredibly hard time reading this new code. I agree it's technically less code, but the beginning and ending conditionals read funny to me.

The old code had the virtue of being very clear, at least IMO.

Further, I have vague ideas about potentially optimizing `StringObject::format` so it doesn't need to do so many resizes, so it could potentially be faster than doing repeated `append()`s.

seven1m: I think I see why you made the changes you did, though, since you were in the position to add at least two more branches to the `if`s. I think I'll :shipit:.

timcraft: Yeah, I wasn't sure what the performance difference would be. Agree about the conditionals 👍

I've opened #390 which reverts back to using `StringObject::format`

PR #398, opened 2022-01-09
Add built-in constants

FnControlOption:


Rakefile
```diff
@@ -233,14 +240,17 @@ file 'bin/nat' => OBJECT_FILES + ['bin/natalie'] do
...
   sh "#{cc} -g -fPIC -c -o #{t.name} #{t.source}"
 end
 
 rule '.cpp.o' => ['src/%n'] + HEADERS do |t|
+  rm_rf 'build/generated/platform.cpp'
   sh "#{cxx} #{cxx_flags.join(' ')} -std=#{STANDARD} -c -o #{t.name} #{t.source}"
 end
 
 rule '.rb.o' => ['.rb.cpp'] + HEADERS do |t|
+  rm_rf 'build/generated/platform.cpp'
```
seven1m, line 253: I'd like to figure out a way to regenerate this file on-demand, when needed, without having to add these `rm_rf` lines everywhere. I'm looking into how to do that with Rake. Let me know if you have ideas!

FnControlOption: I pushed a commit that regenerates the file in a much better way:

```rb
file 'build/generated/platform.cpp' => OBJECT_FILES - ['build/generated/platform.cpp.o'] do |t|
  # ...
end
```

seven1m: Cool thanks!

PR #400, opened 2022-01-09
Disallow overwriting optimized methods

richardboehme:
This PR prints a warning when trying to overwrite optimized methods like Integer#+ or Array#min. To allow those sorts of overwrites a new flag `--allow-overwrites` was introduced with disables the canary compiler step and the warning.

To test this I also implement the `ruby_exe` helper from mspec.

I'm not really happy with how I pass the new flag from the compiler to our code. If there are better options for this I'd be happy to change it 😄

### DIFF
--- a/include/natalie/module_object.hpp
+++ b/include/natalie/module_object.hpp
@@ -144,6 +144,15 @@ public:
     bool has_env() { return !!m_env; }
     Env *env() { return m_env; }
 
+    void add_non_overwritable_method(SymbolObject *method) {
+        m_non_overwritable_methods.set(method);
+    }
+
+    void add_non_overwritable_methods(const TM::Vector<SymbolObject *> &methods) {
+        for (auto method : methods)
+            add_non_overwritable_method(method);
+    }
+
     virtual void visit_children(Visitor &) override final;
 
     virtual void gc_inspect(char *buf, size_t len) const override {
@@ -164,6 +173,8 @@ protected:
     Vector<ModuleObject *> m_included_modules {};
     MethodVisibility m_method_visibility { MethodVisibility::Public };
     bool m_module_function { false };
+
+    TM::Hashmap<SymbolObject *> m_non_overwritable_methods {};
 };
 
 }
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -178,7 +178,9 @@ def ruby_exe(code, args: nil)
     file.write(code)
     file.rewind
 
-    output = `bin/natalie #{file.path} #{args}`
+    binary = ENV['NAT_BINARY'] || 'bin/natalie'
+
+    output = `#{binary} #{file.path} #{args}`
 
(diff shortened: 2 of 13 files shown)


richardboehme:
I have to do some work on Tempfile to make the new define_method tests work on Ruby and Natalie.

test/support/spec.rb
```diff
@@ -173,6 +173,19 @@ def min_long
   -((2**(8 * 8 - 2)) * 2)
 end
 
+def ruby_exe(code, args: nil)
+  Tempfile.create('ruby_exe.rb') do |file|
+    file.write(code)
+    file.rewind
+
+    output = `bin/natalie #{file.path} #{args}`
```
seven1m, line 181: This should check for the existence of `ENV['NAT_BINARY']` and use that if available:
```suggestion
    binary = ENV['NAT_BINARY'] || 'bin/natalie'
    output = `#{binary} #{file.path} #{args}`
```

include/natalie/module_object.hpp
```diff
@@ -164,6 +173,8 @@ class ModuleObject : public Object {
     Vector<ModuleObject *> m_included_modules {};
     MethodVisibility m_method_visibility { MethodVisibility::Public };
     bool m_module_function { false };
+
+    TM::Hashmap<SymbolObject *> m_non_overwritable_methods {};
```
seven1m, line 177: I don't like that we need another Hashmap on every module, but it seems to have negligible effect on performance from my testing locally.

Still, if I could think of a way to do this without a bunch of hashmaps, I would. Perhaps a single Hashmap on GlobalEnv that uses a `std::pair<ModuleObject*, SymbolObject*>` for its key... :thinking:

richardboehme: > Still, if I could think of a way to do this without a bunch of hashmaps, I would. Perhaps a single Hashmap on GlobalEnv that uses a `std::pair<ModuleObject*, SymbolObject*>` for its key... 🤔

I've tried implementing this but it seems like our HashMap implementation does not support hashing std::pair objects. I'm getting the following error:
```
natalie/include/tm/hashmap.hpp:152:31: error: cannot convert ‘std::pair<Natalie::ModuleObject*, Natalie::SymbolObject*>’ to ‘const void*’ in argument passing
  152 |         auto hash = m_hash_fn(key);
      |                               ^~~
      |                               |
      |                               std::pair<Natalie::ModuleObject*, Natalie::SymbolObject*>
```

As a C++ beginner I'm not quite sure why one HashMap for each ModuleObject would result in a big performance decrease. Is it because of the initialization? Maybe an alternative would be to set the initial capacity of those HashMaps to 0?

seven1m: I just checked the Hashmap code, and it seems we do not allocate the chunk of memory for storing items until the first item is added. In that case, I would feel ok with this as-is.

PR #406, opened 2022-01-10
Implement Math functions

timcraft:
This implements most of the functions in the Math module.

I've left out `Math.lgamma`, `Math.ldexp`, and the second argument for `Math.log` due to some oddities I can't work out right now. And there are a couple of issues I've added `NATFIXME` comments for.


src/math.rb
```diff
@@ -249,6 +249,20 @@ def log2(x)
...
+      raise TypeError, "can't convert String into Float" if x.is_a?(String)
+      begin
+        x = Float(x)
+      rescue ArgumentError
+        raise TypeError, "can't convert #{x.class.name} into Float"
+      end
+      return Float::NAN if x.nan?
+      raise DomainError, 'Numerical argument is out of domain' if x < 0
+      __call__('::log', x)
+    end
```
Hendiadyoin1, line 264: How about adding a variadic argument?
(Plus the checks I forgot)
```suggestion
    def log(*x)
      if x.length == 1
        raise TypeError, "can't convert String into Float" if x.is_a?(String)
        begin
          x = Float(x)
        rescue ArgumentError
          raise TypeError, "can't convert #{x.class.name} into Float"
        end
        return Float::NAN if x.nan?
        raise DomainError, 'Numerical argument is out of domain' if x < 0
        __call__('::log', x)
      elsif x.length==2
        return log(x[0]) / log(x[1])
      else
        raise ArgumentError, 'Wrong number of arguments: #{x.length} (1 or 2 expected)'
      end
    end
```

timcraft: Implemented, thanks!

src/math.rb
```diff
@@ -1,7 +1,213 @@
 require 'natalie/inline'
 
 module Math
+  class DomainError < StandardError
+  end
+
   class << self
+    __function__('::acos', ['double'], 'double')
```
seven1m, line 8: Nice! Glad we can use this so soon.

PR #408, opened 2022-01-10
Allow folding of simple conditional assignments

Hendiadyoin1:
### Optimize self-assignment out

This is just a neat way of ensuring a variable exists, so we can replace
it with just a declare_var statement.

### Allow folding of simple conditional assignments

This works, by replacing the non assigning path with a `var = var`
statement to ensure that the variable it is allocated.

---- 

This now allows us to fold
```rb
x = 1 unless false
```
to
```rb
x = 1
```

### DIFF
--- a/lib/natalie/compiler/pass0c.rb
+++ b/lib/natalie/compiler/pass0c.rb
@@ -38,18 +38,38 @@ module Natalie
 
       def process_if(exp, block_with_args = nil)
         exp = exp.new(*exp.map { |e| process_atom(e) })
-        if exp.flatten.include?(:lasgn)
-          # This is required for now because
-          # ```rb
-          # x = 1 unless true # s(:if, s(:true), nil, s(:lasgn, :x, s(:lit, 1)))
-          # ```
-          # gets converted to an if-statement, that does not allocate x in the
-          # taken branch, which makes x undefined
-          return exp
-        end
         _, (boolean, *), true_case, false_case = exp
-        return true_case if type_is_always_truthy(boolean)
-        return false_case if type_is_always_falsey(boolean)
+
+        # This transforms variations of
+        # ```rb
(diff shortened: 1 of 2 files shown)


TheGrizzlyDev (approved):
Feels a bit hacky but very interesting nonetheless. I am really surprised assigning undeclared vars to themselves does result in just a pure declaration in MRI itself, but this actually achieves both goals, compliance and optimisation, very cool :D

lib/natalie/compiler/pass0c.rb
```diff
@@ -38,18 +38,37 @@ def type_is_always_falsey(arg)
...
-        end
         _, (boolean, *), true_case, false_case = exp
-        return true_case if type_is_always_truthy(boolean)
-        return false_case if type_is_always_falsey(boolean)
+
+        # This transforms variations of
+        # ```rb
+        # x = 1 unless true # s(:if, s(:true), nil, s(:lasgn, :x, s(:lit, 1)))
+        # ```
+        # To use `x = x`, which will be optimized away in pass1
```
seven1m, line 47: This is cool, but the comment is confusing. When I played with this just now, I get:

```
bin/natalie -e "x = 1 unless false" -d p1
s(:block, s(:var_set, :env, s(:s, :x), s(:"Value::integer", 1)))
```

...and for the opposite case:

```
bin/natalie -e "x = 1 unless true" -d p1
s(:block, s(:var_declare, :env, s(:s, :x)))
```

So is this comment really correct?

BTW I think the optimization is cool and it makes sense to have it, but I just disagree with what the comment is saying...

Hendiadyoin1: I have adjusted the wording and the example to be a bit more precise.
Is it better now?

PR #410, opened 2022-01-11
Implement String prepend

TheGrizzlyDev:
Mostly based around String#concat with some very minor differences

References: #217

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -371,9 +371,9 @@ Value StringObject::prepend(Env *env, size_t argc, Value *args) {
         }
 
         str_obj->assert_type(env, Object::Type::String, "String");
-        appendable.append(str_obj->c_str());
+        appendable.append(&str_obj->m_string);
     }
-    m_string.prepend(appendable.c_str());
+    m_string.prepend(&appendable);
 
     return this;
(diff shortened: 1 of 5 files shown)


src/string_object.cpp
```diff
@@ -344,6 +344,40 @@ Value StringObject::ord(Env *env) {
...
+        } else if (arg->respond_to(env, to_str)) {
+            str_obj = arg.send(env, to_str)->as_string();
+        } else {
+            env->raise("TypeError", "cannot call to_str", arg->inspect_str(env));
+        }
+
+        str_obj->assert_type(env, Object::Type::String, "String");
+        appendable.append(str_obj->c_str());
+    }
+    m_string.prepend(appendable.c_str());
```
seven1m, line 376: One thing that @FnControlOption worked on was making our `String` class properly handle internal null characters `\0`. I think we need to avoid dropping down to cstrs as much as possible for that reason.

```ruby
s = "foo"
s.prepend("bar\0baz")
p s # => "bar\u0000bazfoo"
```

seven1m: You should be able to add a `String::prepend()` that accepts another String....

PR #423, opened 2022-01-13
Some small GC improvements

TheGrizzlyDev:
Gain: 1.05-1.10 faster across variously configured benchmarks of various scripts

Changes:
-  Automatically unmark cells during collection so that we completely remove the global unmarking from the marking phase
- Do not check again cells that have already been gathered from gather_conservative_roots()


seven1m:
Nice!! 🎉

PR #428, opened 2022-01-14
Add support for Float literals to VM

ryangjchandler:
Adds support for floats to the VM.

Let me know if I've missed anything, but it's basically identical to the `PushIntInstruction`.


seven1m:
Nice! 🎉

PR #430, opened 2022-01-15
More constant folding support

Hendiadyoin1:
This mainly includes more folding support for array literals.

### DIFF
--- a/lib/natalie/compiler/pass0c.rb
+++ b/lib/natalie/compiler/pass0c.rb
@@ -126,9 +126,12 @@ module Natalie
               end
             end
           elsif args.length == 0
-            if method == :!
-              # This can fail and fallback, so we should not try to process it again
-              return canary_const_not(exp, values)
+            # These can fail and fallback, so we should not try to process them again
+            case method
+            when :!
+              return canary_const_not(exp)
+            when :is_truthy
+              return canary_const_truthy(exp)
             end
           end
         end
@@ -173,10 +176,17 @@ module Natalie
         return exp.new(:false)
       end
 
-      def canary_const_not(exp, values)
+      def canary_const_not(exp)
         _, (type, *), _not = exp
         return exp.new(:false) if type_is_always_truthy(type)
         return exp.new(:true) if type_is_always_falsey(type)
+        return exp
+      end
+
+      def canary_const_truthy(exp)
+        _, (type, *), is_truthy = exp
+        return exp.new(:true) if type_is_always_truthy(type)
+        return exp.new(:false) if type_is_always_falsey(type)
         return exp
       end
     end
--- a/lib/natalie/compiler/pass0c.rb
+++ b/lib/natalie/compiler/pass0c.rb
@@ -124,16 +124,19 @@ module Natalie
         _, receiver, method, *args = exp
         if receiver && !receiver.empty?
           t, *values = receiver
-          if t == :array && args.empty? && block_with_args.nil?
+          if t == :array && block_with_args.nil?
+            # FIXME: Maybe throw argument errors when we detect anything wrong with the argument count
             case method
+            when :include?
+              return process(canary_array_include(exp, values, *args)) if args.length == 1
             when :min
-              return process(canary_array_min(exp, values))
+              return process(canary_array_min(exp, values)) if args.length == 0
             when :max
-              return process(canary_array_max(exp, values))
+              return process(canary_array_max(exp, values)) if args.length == 0
             when :any?
-              return process(canary_array_any(exp, values))
+              return process(canary_array_any(exp, values)) if args.length == 0
             when :all?
-              return process(canary_array_all(exp, values))
+              return process(canary_array_all(exp, values)) if args.length == 0
             end
           elsif t == :lit && args.length == 1 && args[0][0] == :lit && !@in_defined_statement
             # we cannot collapse these in defined? statements, because it changes the type
@@ -259,6 +262,14 @@ module Natalie
 
         return exp.new(:call, new_function, :is_truthy) if values.length == 1
         return exp.new(:and, exp.new(:call, new_function, :is_truthy), canary_array_all(exp, rest, args, function))
+      end
+
+      def canary_array_include(exp, values, arg)
+        return exp.new(:false) unless values&.length > 0
+
+        first, *rest = values
+        return exp.new(:call, first, :==, arg) if values.length == 1
+        return exp.new(:or, exp.new(:call, first, :==, arg), canary_array_include(exp, rest, arg))
       end
 
       def canary_const_op(exp, values)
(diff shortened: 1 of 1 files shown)


lib/natalie/compiler/pass0c.rb
```diff
@@ -16,7 +16,33 @@ def initialize(context)
...
+              return process(canary_array_min(exp, values, args, *body))
+            when :max
+              return process(canary_array_max(exp, values, args, *body))
+            when :any?
+              return process(canary_array_any(exp, values, args, *body))
+            when :all?
+              return process(canary_array_all(exp, values, args, *body))
+            end
+          end
+        end
```
seven1m, line 45: As this file grows, it's going to be really hard to understand what these various optimizations are doing. At the very least, we should move this code to a method called `optimize_single_value_array_operations` or some such, just so it's easy to talk about and see at-a-glance what the intention is. Otherwise it's just a big block of AST manipulation and my eyes glaze over.

lib/natalie/compiler/pass0c.rb
```diff
@@ -98,11 +124,19 @@ def process_call(exp, block_with_args = nil)
...
+            when :include?
+              return process(canary_array_include(exp, values, *args)) if args.length == 1
+            when :min
+              return process(canary_array_min(exp, values)) if args.length == 0
+            when :max
+              return process(canary_array_max(exp, values)) if args.length == 0
+            when :any?
+              return process(canary_array_any(exp, values)) if args.length == 0
+            when :all?
+              return process(canary_array_all(exp, values)) if args.length == 0
```
seven1m, line 139: I believe the `canary_array_whatever` methods should decide which exps to process... otherwise I have to jump back and forth a couple places to know if an optimization is running...

lib/natalie/compiler/pass0c.rb
```diff
@@ -126,43 +160,118 @@ def process_call(exp, block_with_args = nil)
...
+          return(
+            exp.new(
+              :if,
+              exp.new(:call, first, :<, second),
+              canary_array_min(exp, [first, *rest]),
+              canary_array_min(exp, [second, *rest]),
+            )
+          )
+        end
+        return exp if args.length != 3 # Wat
```
seven1m, line 189: This doesn't make sense. If it doesn't make sense to the one who wrote it, I guarantee it won't make sense to the reader. This either needs a comment or a special method to handle the edge case.

lib/natalie/compiler/pass0c.rb
```diff
@@ -126,43 +160,118 @@ def process_call(exp, block_with_args = nil)
...
+
+        new_function1 = function.map { |expr| ((expr == s(:lvar, args[1])) ? first : expr) }
+        new_function2 = function.map { |expr| ((expr == s(:lvar, args[2])) ? second : expr) }
```
seven1m, line 192: I have no idea what this is doing. Needs comments or a named method to describe it.

lib/natalie/compiler/pass0c.rb
```diff
@@ -126,43 +160,118 @@ def process_call(exp, block_with_args = nil)
...
+
         return(
           exp.new(
             :if,
-            exp.new(:call, first, :<, second),
+            exp.new(:call, new_function1, :<, new_function2),
```
seven1m, line 197: `new_function1` and `new_function2` confused me. Why do we need new functions?

lib/natalie/compiler/pass0c.rb
```diff
@@ -126,43 +160,118 @@ def process_call(exp, block_with_args = nil)
...
+              exp.new(:call, first, :>, second),
+              canary_array_max(exp, [first, *rest]),
+              canary_array_max(exp, [second, *rest]),
+            )
+          )
+        end
+        return exp if args.length != 3 # Wat
+
+        new_function1 = function.map { |expr| ((expr == s(:lvar, args[1])) ? first : expr) }
+        new_function2 = function.map { |expr| ((expr == s(:lvar, args[2])) ? second : expr) }
```
seven1m, line 222: Lots of duplication from the previous method? Can that be cleaned up?

lib/natalie/compiler/pass0c.rb
```diff
@@ -126,43 +160,118 @@ def process_call(exp, block_with_args = nil)
...
+        return exp.new(:false) unless values&.length > 0
+
+        first, *rest = values
+        if args.nil? && function.nil?
+          return exp.new(:call, first, :is_truthy) if values.length == 1
+          return exp.new(:or, exp.new(:call, first, :is_truthy), canary_array_any(exp, rest))
+        end
+
+        return exp if args.length != 2 # Wat
+        new_function = function.map { |expr| ((expr == s(:lvar, args[1])) ? first : expr) }
```
seven1m, line 244: Again with the 'wat'

seven1m (changes requested):
As a proof of concept, this PR is quite neat in that it shows what is possible for small arrays.

But in terms of the implementation, this PR could use some work. I fear there are so many FIXMEs and "wats" in it this will cause us weird hard-to-debug issues down the road. Not to mention it's dang-near impossible to read as someone who isn't familiar with what these Sexps actually look like.

If we want to do these sorts of optimizations, I think we have to think about how this pass0c is architected. It needs many many more named chunks of code (methods) and/or comments at the minimum.

Ideally, we would have separate classes handling each of these optimizations, I think.

Pass0c is growing fast, but it's in the awkward teenage years and we need to help it grow into adulthood.

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: I see that point,
but with the current architecture this would make it loop processing the same statement,
because we would try to reprocess the if that it should spit out, but it's still the call it was

In this case these checks should be an ArgumentErrors, I guess...
I will move them into the functions for now, but going forward separating this into multiple optimization passes would be the best way going forward, as discussed in the discord

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: As stated above this should probably be an ArgumentError

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: I will add a comment about this, maybe even outline it and rename the variables

But for now,
we replace every mention of the variable passed into the block (ie `s(:lvar, args[1])`) with the actual value we are currently iterating on

lib/natalie/compiler/pass0c.rb
Hendiadyoin1: as stated above these represent the passed block/function with the value inlined

The names arent really good, as I have to admit

seven1m:
I'm excited about all the potential compiler optimizations we can do with Natalie, but I'd like to wait until we have Compiler2 a little more fleshed out, and then I'd like to see what a small, simple optimization might look like against the IR, in a compiler pass that works against the IR (rather than the AST). I'd like to hope that it will be a bit easier to understand, but time will tell!

I appreciate the exploration you have done in this space, but I'm going to close this for now. Let's circle back and implement constant folding on the new compiler once things are looking pretty good there.

PR #432, opened 2022-01-15
Vector improvements - improve relocation of existing entries in insert and remove

TheGrizzlyDev:
This especially crafted script measures up to a 10% improvement in speed 
```ruby
a = []
100_000.times { a << 1 }
100_000.times { a.pop }
```

Having said that, we shouldn't expect the same speed-up anywhere else.

### DIFF
--- a/include/tm/vector.hpp
+++ b/include/tm/vector.hpp
@@ -14,25 +14,26 @@ const int VECTOR_MIN_CAPACITY = 10;
 template <typename T>
 class Vector {
 public:
+    #define ARRAY_OF_SIZE(size) static_cast<T *>(::operator new(size * sizeof(T)))
     Vector()
         : m_capacity { VECTOR_MIN_CAPACITY }
-        , m_data { new T[VECTOR_MIN_CAPACITY] {} } { }
+        , m_data { ARRAY_OF_SIZE(VECTOR_MIN_CAPACITY) } { }
 
     Vector(size_t initial_capacity)
         : m_size { initial_capacity }
         , m_capacity { initial_capacity }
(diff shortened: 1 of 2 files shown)


include/tm/vector.hpp
```diff
@@ -4,17 +4,16 @@
...
 
-#define TM_MIN(a, b) (a < b ? a : b)
-
 namespace TM {
 
 const int VECTOR_GROW_FACTOR = 2;
 const int VECTOR_MIN_CAPACITY = 10;
 
+#define ARRAY_OF_SIZE(size) reinterpret_cast<T *>(new unsigned char[size * sizeof(T)] {})
+
```
seven1m, line 13: I'm still a C++ noob, but I think you can do this without a macro using `constexpr`...

```
    constexpr T *array_of_size(size_t size) {
        return reinterpret_cast<T *>(new unsigned char[size * sizeof(T)] {});
    }
```

TheGrizzlyDev: I am not sure you can, wouldn't a constexpr be evaluated at compile-time (maybe we understood this feature differently)? I think you could do this with a function that is always inlined, but it's quite a minimal macro and unlikely to be changed :) Plus just to be safe I undefine it at the end because I don't want it to be used anywhere else :D

seven1m (approved):
This is great!

I was going to suggest prefixing your macro with `TM_` just because macros share the global namespace and I like to be a good citizen, but then I think I found a way to do it with a constexpr function? Your call on that.

include/tm/vector.hpp
seven1m: OK fair enough, carry on.

PR #435, opened 2022-01-16
Use realloc instead of reallocarray to support macOS

FnControlOption:


TheGrizzlyDev:
How about only using realloc then?

PR #438, opened 2022-01-17
Implement Integer::try_convert

richardboehme:
This also bumps our targeted ruby version to 3.1.0

### DIFF
--- a/spec/core/hash/transform_keys_spec.rb
+++ b/spec/core/hash/transform_keys_spec.rb
@@ -76,7 +76,8 @@ describe "Hash#transform_keys!" do
   end
 
   ruby_version_is "3.0.2" do
-    it "returns the processed keys and non evaluated keys if we break from the block" do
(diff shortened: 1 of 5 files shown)


spec/core/hash/transform_keys_spec.rb
```diff
@@ -76,7 +76,7 @@
   end
 
   ruby_version_is "3.0.2" do
-    it "returns the processed keys and non evaluated keys if we break from the block" do
+    xit "returns the processed keys and non evaluated keys if we break from the block" do
```
seven1m, line 79: I think we should put `NATFIXME` notes here so we don't forget to come back and fix these. (Unless of course, you already have fixes on the way... :wink:)

richardboehme: Thanks for catching that! We should definitely do that. I'm going to add them later today.

PR #439, opened 2022-01-17
Compiler2: Port, Devide and Expand pass0[c]

Hendiadyoin1:
A first stab at some of the optimizer things discussed in the discord, applied to the original compilers Pass0c
Plus a reworked, improved and fixed version of #430 for Compiler2

This PR uses a more Serenity-ish commit style starting with the second commit to differentiate the Compilers Pass0c and Compilerr2s Pass0, tell me if I need to adjust the wording/message-style in any way

This might accidentally increase the capabilities of the stack-based IR by allowing some Array functions with blocks on constant arrays, which we don't seem to support otherwise...

In the future we might want to re-implement these passes using the stack-based IR in the future, as it might allow us to see into some variable's values and to eliminate the dependence on `Natalie::Compiler::BasePass`

@TheGrizzlyDev
I have removed the `block_with_args`, I wasn't sure if it was needed
Can you check the re-check logic behind the canary-pass?

### DIFF
--- a/bin/natalie
+++ b/bin/natalie
@@ -13,7 +13,7 @@ require_relative '../lib/natalie'
 
 orig_args = ARGV.dup
 
-options = { load_path: [], require: [], profile: false }
+options = { load_path: [], require: [], profile: false , optimization_level: 0}
 OptionParser.new do |opts|
   opts.banner = 'Usage: natalie [options] file.rb'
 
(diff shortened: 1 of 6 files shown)


seven1m:
Same as my note on #430, I'd like to wait on working on compiler optimization stuff for now, at least until we're relatively sure Compiler2 is going to be the path forward. Then I'd like to see what an IR-based optimization pass looks like...

PR #441, opened 2022-01-18
Compiler2: optional args

seven1m:
Just pushing this up as I work on it, so everyone knows what's going on. I haven't figured out how to handle intentionally-passed `nil` values to methods yet, but I will...


seven1m:
I feel mostly ok with this work, but I'm open to suggestions on the new instructions. In earlier commits, I thought we could use two primitive instructions `array_pop` + `or`, but that did not allow the user to pass an intentional `nil` as an argument.

cc14144a78f731334cdc91bf1d037b39455c575a creates a couple new instructions `array_pop_with_default` and `array_shift_with_default`. I can't say that I love the names, but I hope their purpose is clear enough that one can see what is going on... (not quite as good as `pop` + `or`, but maybe good enough):

```
→ bin/natalie -i -e "def foo(x = :default, y); end" -d p1
0 define_method foo
1 push_args
2 array_pop
3 variable_set y
4 push_symbol :default
5 array_pop_with_default # if the array has a value in it (even if nil), that value is used
6 variable_set x
7 pop
8 push_nil
9 end define_method
```

I'm open to feedback on this. I want the new compiler to be as readable as possible!

PR #447, opened 2022-01-19
Compiler2: Compile range literals

FnControlOption:


seven1m:
Since `vm_test.rb` runs the VM on top of Natalie, it is using our own parser rather than the `ruby_parser` gem. Looks like we need to work on range parsing...

src/parser.cpp
```diff
@@ -1456,7 +1462,15 @@ Node *Parser::parse_proc_call_expression(Node *left, LocalsHashmap &locals) {
 Node *Parser::parse_range_expression(Node *left, LocalsHashmap &locals) {
     auto token = current_token();
     advance();
-    return new RangeNode { token, left, parse_expression(LOWEST, locals), token->type() == Token::Type::DotDotDot };
+    Node *right;
+    try {
+        right = parse_expression(LOWEST, locals);
+    } catch (SyntaxError &e) {
+        // NOTE: I'm not sure if this is the "right" way to handle an endless range,
+        // but it seems to be effective for the tests I threw at it. ¯\_(ツ)_/¯
```
FnControlOption, line 1470: CRuby prints a warning:

```sh-session
irb(main):001:0> 1..
=> 1..
irb(main):002:0> (1...)
=> 1...
irb(main):003:0> 1...
(irb):3: warning: ... at EOL, should be parenthesized?
=> 1...
```

seven1m: Why would `...` print a warning but not `..` 🤔

FnControlOption: Maybe it's because of the arguments forwarding operator `...` added in 2.7:

```
irb(main):001:0> RUBY_VERSION
=> "2.6.3"
irb(main):002:0> 1..
irb(main):003:0* 2
=> 1..2
irb(main):004:0> 1...
irb(main):005:0* 2
=> 1...2
```

```
irb(main):001:0> RUBY_VERSION
=> "2.7.5"
irb(main):002:0* 1..
irb(main):003:0> 2
=> 1..2
irb(main):004:0* 1...
irb(main):005:0> 2
(irb):4: warning: ... at EOL, should be parenthesized?
=> 1...2
```

```
irb(main):001:0> RUBY_VERSION
=> "3.0.3"
irb(main):002:0> 1..
=> 1..
irb(main):003:0> 1...
(irb):3: warning: ... at EOL, should be parenthesized?
=> 1...
```

The script below also gives the same output for all three versions (but only 2.6 doesn't output a warning):

```rb
r = 1..
3
p r # => 1..3
r = 1...
3
p r # => 1...3
```

seven1m: That's good to know. We can work on the parsing of ranges in another PR; at least it works well enough for this compiler functionality. 😄

PR #449, opened 2022-01-19
C2: Parse case-statements

Hendiadyoin1:
To do this we essentially do what C1 does:
Unpacking everything into an huge if-else-block

To achieve this two new instructions were needed;
* **Swap**
  Swaps the stack top with the one below
* **PushRel**(index)
  Pushes an element already on the stack to the stack
  `index` gives the index of the item to be pushed relative to the stack-top starting at 0


seven1m:
Our parser cannot yet parse an expressionless `case` expression, e.g.

```ruby
case
when foo
  # ..
when bar
  # ..
end
```

I'll see if I can work on that...

lib/natalie/compiler2/pass1.rb
```diff
@@ -78,6 +78,66 @@ def transform_call(exp, used:, with_block: false)
...
+        if var
+          instructions << transform_expression(var, used: true)
+          whens.each do |when_statement|
+            # case a
+            # when b, c, d
+            # =>
+            # if (a === b || a === c || a === d)
+            _, options, body = when_statement
+
+            options[1...].each do |option|
```
seven1m, line 94: I dislike using the `[1...]` syntax for this, as it makes it hard to read. Why skip the first one? (Don't answer that in a comment please -- make it clear in the code.)

In this case, should we destructure the `options` array so we can label the parts?

lib/natalie/compiler2/instructions/push_rel_instruction.rb
```diff
@@ -0,0 +1,24 @@
+require_relative './base_instruction'
+
+module Natalie
+  class Compiler2
+    class PushRelInstruction < BaseInstruction
+      def initialize(index)
+        @index = index
+      end
+      def to_s
+        "push st[#{@index}]"
```
seven1m, line 10: Let's call this `push_rel`:

```suggestion
        "push_rel #{@index}"
```

lib/natalie/compiler2/pass1.rb
seven1m, line 94: I dislike using the `[1...]` syntax for this, as it makes it hard to read. Why skip the first one? (Don't answer that in a comment please -- make it clear in the code.)

In this case, should we destructure the `options` array so we can label the parts?

```suggestion
            _, options_array, body = when_statement
            _, *options = options_array
            options.each do |option|
```

lib/natalie/compiler2/instructions/push_rel_instruction.rb
seven1m: On second thought, I wonder if we should call this instruction `dup_rel` 🤔

test/natalie/compiler2/bootstrap_test.rb
```diff
@@ -124,6 +124,40 @@ def test_or
...
+
+    case
+    when false
+      fail
+    when nil
+      fail
+    when 1 + 1 == 2
+      foo = :foo
+    end
+    assert_eq(foo, :foo)
```
seven1m, line 157: We should add a test for a `case` that has no `else` and does not match any of the branches:

```sh
bin/natalie -i -e "case 0; when 1; end"
natalie/lib/natalie/vm.rb:49:in `pop': out of stack (RuntimeError)
```

Hendiadyoin1: Seems to be an error in the interpreter,
The cpp backend seems to be able to compile this...

Hendiadyoin1: Found it
We should push nil in empty blocks

lib/natalie/compiler2/instructions/push_rel_instruction.rb
Hendiadyoin1: Maybe just make it `dup`?
current `dup` is just the `push st[0]` case
So
`dup` -> `dup 0`
`push st[N]` -> `dup N`

---
btw I just used a notation similar to x87 indexing, because that's what I knew

lib/natalie/compiler2/pass1.rb
Hendiadyoin1: 👍 
---
but
```rb
 # case
            # when a == b, b == c, c == d
            # =>
            # if (a == b || b == c || c == d)
            _, options, body = when_statement

            # s(:array, option1, option2, ...)
            # =>
            # s(:or, option1, s(:or, option2, ...))
            options = options[2..].reduce(options[1]) { |prev, option| s(:or, prev, option) }
 ```
 in the var-less case is fine/documented enough?
 just to confirm

lib/natalie/compiler2/instructions/push_rel_instruction.rb
seven1m: I don't know why, but `dup 0` feels weird to me. I'd be ok with having `dup` and `dup_rel <index>`.

lib/natalie/compiler2/pass1.rb
seven1m: Yeah, the comments serve us well there, and I think it makes sense. Thanks!

seven1m:
I pushed a fix (again) for the expression-less case parsing. Try not to force push over it again :laughing:.

Hendiadyoin1:
Reordered, Rebased and addressed the naming issue.

seven1m:
🎉 nice!

PR #455, opened 2022-01-20
Return enumerators with sizes

timcraft:
This implements the `:enumeratorized_with_unknown_size` and `:enumeratorized_with_origin_size` behavior in `spec/core/enumerable/shared/enumeratorized.rb` for Array, Hash, Enumerable, and Enumerator methods


seven1m:
Nice!

PR #460, opened 2022-01-21
Compiler2: Add `memoize` instruction

FnControlOption:


Hendiadyoin1:
As @seven1m already said, maybe we could fix that up in the other instructions, as it is only a problem with the c++ backend

other than that maybe telling the reasoning in the commit message would be a good idea

seven1m:
I'm hoping we can make the CppBackend smarter rather than make instructions that kind of leak the implementation of the C++ generation out to the IR.

PR #463, opened 2022-01-23
Remove GC-managed objects from the Parser

seven1m:
I'm working to extract the Parser from Natalie, and one step toward that is to have it not use the GC.


seven1m:
Wrong branch. Whoops.

PR #465, opened 2022-01-24
C2: Use c++ stack for known variables of a context

Hendiadyoin1:
For this we introduce the declare-instruction, which marks the declared
variable as not captured by the environment and declares it as a `Value`
in the cpp-backend.

This instruction is generated in the second pass of the new compiler.

---

This was tested on
```rb
def f
  x = 1
  p x
end

x = 2

f

p x
```
and produced 
```cpp
        Value x1;
        self->define_method(env, "f"_s, defn_f2, 0);
        x1 = Value::integer(2);
        auto send_f5 = self->send(env, "f"_s, {  }, nullptr);
        auto send_p6 = self->send(env, "p"_s, { x1 }, nullptr);
        send_p6;
```
for the main block
and
```cpp
    Value x3;
    x3 = Value::integer(1);
    auto send_p4 = self->send(env, "p"_s, { x3 }, nullptr);
    return send_p4;
```
for f.

The Instruction-array manipulation is very weird and should probably be improved upon by someone who knows more about ruby than me.

I am also not quite sure about the `define_class` and `define_block` aliases


seven1m:
This is a good try, but it's a bit early for this optimization. We don't yet have lambda/procs/blocks implemented yet, so it's impossible to know if this is the right way to do it without a proper closure to test it with.

And just like I said on #460, I'm not convinced we need to invent IR instructions that are only useful to the CppBackend. I'd like to explore the possibility of making the backend smart enough to know when it needs to set a local variable vs one on the Env.

PR #473, opened 2022-01-28
Fix Rakefile for Mac

FnControlOption:


seven1m:
Whoops, thanks! (Hadn't gotten back to the Mac to test yet.)

PR #476, opened 2022-01-28
Implement Integer#divmod and Integer#digits

richardboehme:
This also fixes a bug with divisions of BigInts and a bug in Integer#% when passed a float.

### DIFF
--- a/src/float_object.cpp
+++ b/src/float_object.cpp
@@ -360,7 +360,17 @@ Value FloatObject::mod(Env *env, Value rhs) {
 
     if (divisor == 0) env->raise("ZeroDivisionError", "divided by 0");
 
-    return Value { fmod(dividend, divisor) };
+    bool negative = false;
+    if (divisor < 0) {
+        negative = true;
+    }
+
+    auto result = fmod(dividend, divisor);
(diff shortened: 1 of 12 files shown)


src/float_object.cpp
```diff
@@ -360,7 +360,12 @@ Value FloatObject::mod(Env *env, Value rhs) {
 
     if (divisor == 0) env->raise("ZeroDivisionError", "divided by 0");
 
-    return Value { fmod(dividend, divisor) };
+    auto result = fmod(dividend, divisor);
+    if (signbit(dividend) != signbit(divisor)) {
+        result += divisor;
+    }
```
seven1m, line 366: Ha! I believe this is the same change I had to make to BigInt here: https://github.com/seven1m/natalie/blob/31e7e82e756e0e6a5b2b797fd5fd778e912c07ed/src/big_int.cpp#L1349-L1350

I didn't really understand it at the time, but glad to see you came to the same solution.

richardboehme: Too be honest I pretty much copied that from BigInt because I knew you had done some changes there to make it ruby-like 😅

PR #477, opened 2022-01-28
Parser: fix some issues with assignment expressions

FnControlOption:
Assignment expression as call argument:

```rb
foo(a = b, c)
```

Multiple assignment with comma-separated values:

```rb
x, y = 1, 2
```

### DIFF
--- a/include/natalie_parser/node.hpp
+++ b/include/natalie_parser/node.hpp
@@ -67,10 +67,12 @@ public:
         Shell,
         Splat,
         SplatAssignment,
+        SplatValue,
         StabbyProc,
         String,
         Super,
         Symbol,
+        ToArray,
         True,
         Until,
         While,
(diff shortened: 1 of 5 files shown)


include/natalie_parser/node.hpp
```diff
@@ -1329,6 +1331,20 @@ class SplatNode : public Node {
     Node *m_node { nullptr };
 };
 
+class SplatValueNode : public Node {
+public:
+    SplatValueNode(const Token &token, Node *value)
+        : Node { token }
+        , m_value { value } { }
```
seven1m, line 1338: You also need a destructor like this one: https://github.com/seven1m/natalie/blob/eafe3045f7bafc096d77bd8de12216dfdf2533db/include/natalie_parser/node.hpp#L173-L175

include/natalie_parser/node.hpp
```diff
@@ -1366,6 +1382,20 @@ class SymbolNode : public Node {
     SharedPtr<String> m_name {};
 };
 
+class ToArrayNode : public Node {
+public:
+    ToArrayNode(const Token &token, Node *value)
+        : Node { token }
+        , m_value { value } { }
```
seven1m, line 1389: Destructor

src/natalie_parser/parser.cpp
```diff
@@ -1193,6 +1197,41 @@ Node *Parser::parse_assignment_expression(Node *left, LocalsHashmap &locals) {
     }
 };
 
+Node *Parser::parse_assignment_expression(bool to_array, LocalsHashmap &locals, bool allow_multiple) {
```
seven1m, line 1200: Let's rename this something like `parse_assignment_expression_value`. If I'm reading this right, it's just giving you back the right hand side of an assignment...

seven1m (changes requested):
This is great work, fantastic. I just had a few small issues and a nitpick with naming as I was reading through the code -- it would help with understanding I think.

seven1m (approved):

PR #482, opened 2022-01-30
Add spec for String#size

TheGrizzlyDev:
#217

### DIFF
--- a/spec/core/string/shared/length.rb
+++ b/spec/core/string/shared/length.rb
@@ -1,0 +1,40 @@
+# encoding: utf-8
+
+describe :string_length, shared: true do
+  it "returns the length of self" do
+    "".send(@method).should == 0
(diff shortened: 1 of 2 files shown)


spec/core/string/shared/length.rb
```diff
@@ -0,0 +1,40 @@
+# encoding: utf-8
+
+describe :string_length, shared: true do
+  it "returns the length of self" do
+    "".send(@method).should == 0
+    "\x00".send(@method).should == 1
+    "one".send(@method).should == 3
+    "two".send(@method).should == 3
+    "three".send(@method).should == 5
+    "four".send(@method).should == 4
```
seven1m, line 10: lol @ these specs 😆

TheGrizzlyDev: yeah, creativity at its finest :^)

PR #485, opened 2022-02-02
Parser: fix operator symbols

FnControlOption:


seven1m:
Nice!

We have some tests in `lexer_test.rb` for symbols -- we should move the tests there: https://github.com/seven1m/natalie/blob/9180dbaf939669978090a283c3dbd0805c0252f2/test/natalie/lexer_test.rb#L336-L369

PR #488, opened 2022-02-04
Implement Rational#to_f

timcraft:


seven1m:
I just noticed you added a commit for each file you added. I think next time this could easily be one atomic commit. 🤗

PR #492, opened 2022-02-05
Implement Integer#round and some new Integer helpers

richardboehme:
This implements `Integer#round`. However the PR got a bit larger than anticipated. I added new helper methods to `IntegerObject` to convert a `Object` to a `nat_int_t` using the `to_int` method. I think I could remove a lot of redundancy there and also all places now can benefit from fast integers and have correct unified error handling. I'd love to have feedback on how to improve the helpers (for example some cases need specific logic for handling bignums, see `IntegerObject::left_shift` for example)!

[#201]


seven1m (approved):
This is fantastic! I love how much code you were able to remove with your helpers. :smiling_face_with_three_hearts:

PR #494, opened 2022-02-05
Parser: move lexer code to .cpp file

FnControlOption:
Are there any particular reasons for keeping most lexer code in the header file?


src/natalie_parser/lexer.cpp
```diff
@@ -103,4 +1406,16 @@ void InterpolatedStringLexer::tokenize_interpolation(SharedPtr<Vector<Token>> to
...
+    if (m_index >= m_size)
+        return 0;
+    return m_input->at(m_index);
+}
+
+char InterpolatedStringLexer::peek() {
+    if (m_index + 1 >= m_size)
+        return 0;
+    return m_input->at(m_index + 1);
+}
```
seven1m, line 1419: Let's move current_char() and peek() back to the header just to ensure they get inlined.

seven1m:
No particular reason...

PR #495, opened 2022-02-06
Try to reduce object allocations

richardboehme:
I wanted to profile Natalie a bit and set my goal to reduce object allocations. 

Therefore I created a flag to print all allocated objects and the total number of allocation when executing a script with Natalie (just pass the `--print-objects` flag). 

First thing I noticed while looking at the boardslam example was that we were creating `StringObject`s every time a call to an `attr_reader` happened. I changed it to only use Symbols there. (same thing for `attr_writer`)

Running the boardslam example with arguments 3 5 1:
* Allocations before: 443475
* Allocations after: 416208

Another thing I noticed is that we create a new `ArrayObject` for each method call. After spending a lot of time changing this (see the last commit) I noticed that this issue is not present in the new compiler. Should've checked that before 😅... I don't know if we want to keep it (otherwise I'll just drop the commit).

Running the fib example:
* Allocations before: 490516
* Allocations after: 247714
(the runtime was reduced by around 25% on my machine)

Let me know how we should progress with this. Also expect the tests to fail maybe. I have to go to bed for now and will fix them later!


src/module_object.cpp
```diff
@@ -477,45 +477,44 @@ Value ModuleObject::name(Env *env) {
 Value ModuleObject::attr_reader(Env *env, size_t argc, Value *args) {
     for (size_t i = 0; i < argc; i++) {
         Value name_obj = args[i];
-        if (name_obj->type() == Object::Type::String) {
+        if (name_obj->type() == Object::Type::Symbol) {
             // we're good!
-        } else if (name_obj->type() == Object::Type::Symbol) {
-            name_obj = name_obj->as_symbol()->to_s(env);
+        } else if (name_obj->type() == Object::Type::String) {
+            name_obj = SymbolObject::intern(name_obj->as_string()->c_str());
```
seven1m, line 483: We should avoid using `c_str()` here and elsewhere because it causes Natalie to not properly handle the null character (`\0`), plus it requires a call to `strlen()`.

You can create a Symbol using the TM::String directly: https://github.com/richardboehme/natalie/blob/d71ab5e75f5b8f7f585bc3a5a2e53282cfdf0b73/include/natalie/symbol_object.hpp#L25

richardboehme: I added a method to `StringObject` that returns the `TM::String` reference we have there and updated `attr_reader/writer` to use this method instead of `c_str()`. We save the `strlen` for now and later if `SymbolObject` starts supporting null bytes we do not have to change those lines.

seven1m:
@richardboehme this is really cool, great work. It does seem to make a pretty dramatic effect on performance:

```
→ hyperfine -w 10 -r 30 ./fib1 ./fib2
Benchmark 1: ./fib1
  Time (mean ± σ):     233.7 ms ±   4.9 ms    [User: 201.2 ms, System: 32.0 ms]
  Range (min … max):   219.8 ms … 241.7 ms    30 runs
 
Benchmark 2: ./fib2
  Time (mean ± σ):     150.5 ms ±   5.0 ms    [User: 138.5 ms, System: 11.8 ms]
  Range (min … max):   141.0 ms … 159.3 ms    30 runs
 
Summary
  './fib2' ran
    1.55 ± 0.06 times faster than './fib1'


→ hyperfine -w 10 -r 30 './bs1 3 5 1' './bs2 3 5 1'
Benchmark 1: ./bs1 3 5 1
  Time (mean ± σ):     245.3 ms ±   8.2 ms    [User: 210.2 ms, System: 34.8 ms]
  Range (min … max):   236.5 ms … 279.4 ms    30 runs
 
Benchmark 2: ./bs2 3 5 1
  Time (mean ± σ):     208.1 ms ±   8.4 ms    [User: 181.9 ms, System: 25.9 ms]
  Range (min … max):   194.9 ms … 239.0 ms    30 runs
 
Summary
  './bs2 3 5 1' ran
    1.18 ± 0.06 times faster than './bs1 3 5 1'
```

richardboehme:
Thanks for benchmarking @seven1m!

I wanted to fix another thing that caught my eyes while watching at allocations in the fib program. For every method call we allocate a `ManagedVector`. In case of the fib program those are over 200000 objects the GC has to handle. While trying to come up with a way of fixing this, I came across the commit you did about it (b64f58a7d8f408b08b2eda9521adeafaa23faf74). I wonder if we can do anything about this... Maybe those short lived `Env` objects we create for method calls could cleanup the vars manually (using Defer maybe)?

seven1m:
>I wonder if we can do anything about this... Maybe those short lived Env objects we create for method calls could cleanup the vars manually (using Defer maybe)?

I wouldn't mind revisiting this, but I don't currently know of a way to make that work. When an Env is copied (as part of a closure, like a block/proc), then that copied Env needs to retain a pointer to the shared variable storage.

As you can see from that commit you linked, I originally was using a SharedPtr, and that worked fine, except the GC could not "see" that pointer. If we had GC-aware SharedPtr, it might work. 🤔 

I think for now, though, we should experiment with that in a separate PR... sound ok?

richardboehme:
> I think for now, though, we should experiment with that in a separate PR... sound ok?

Sounds good to me!

PR #503, opened 2022-02-09
Fix integer shift operations

richardboehme:
This also adds the `ruby_bug` method to the spec library. I figured we just want to skip over specs for ruby bugs. What do you think? @seven1m


seven1m:
Yes, I agree. We don't need to replicate Ruby bugs. :laughing:

PR #506, opened 2022-02-13
Trying to rework how we handle integers

richardboehme:
Hey! This is a super drafty pull request. Let's call it a proposal 😄 

### Motivation

While working on `Integer` I was constantly annoyed by the way we handle Bignums. Especially that we often implement logic twice. Mostly the logic for extracting/converting arguments or edge cases/exceptions. After some time I started implementing the fixnum and bignum case in one method in `IntegerObject` (for example `IntegerObject::bitwise_or`). However, this gets super complex for methods that are more complicated. For example I wanted to implement `Integer#pow`. This function takes up to two arguments which can both be bignums but do not have to be. The methods gets complex pretty fast because of all the different cases we would have to handle (self is fixnum vs bignum, arg1 is fixnum vs bignum, arg2 is fixnum vs bignum). This is why I would like to split `IntegerObject` and the handling of fixnums/bignums.

### Solution

My first idea was to implement an interface (or abstract base class) `Integer` that has two implementations `Fixnum` and `Bignum` that would implement all possible operations we need on `Integers`. We could use this as a private member on `IntegerObject` instead of the `nat_int_t` and remove `BignumObject`. I failed implementing this because operators like `operator+` have to return a copy of an object and returning an instance of an abstract base class does not seem to be allowed.  Also I think promoting/demoting to or from Bignums may be difficult with this solution. 

What I try in this PR is to just implement one `Integer` class which can hold a `BigInt` or a `nat_int_t`.  I'm not happy with the implementation yet but this should be a demonstration of what I would like achieve. Also I'm not sure what the implications on performance would be, considering that we definitely add some overhead to simple fixnum operations (instead of directly doing them on nat_int_ts we proxy them through another class).

I would appreciate feedback on the `Integer` class I created (especially on the union stuff I did there; it seems kinda wrong to me?). Maybe we can create a better abstraction than what I came up with. My main goal is to separate and abstract away the handling of fixnums vs. bignums from the `IntegerObect` class.

### Progress

For now I implemented the new Integer class for `Integer#+` and `Integer#**`. As you can see I could remove the methods from `BignumObject` and simplify the methods in `IntegerObject`.

### Conclusion

As I said this is an idea I came up with, that I just wanted to try out. I think it would be beneficial if we want to use Integers in our code (also in other classes than `IntegerObject`). Also we could simplify a lot of code in `IntegerObject` and reduce duplication of some code by deleting `BignumObject`. I'm not happy yet with the approach, so let me know if there are any other ideas on how we could handle the problem. If you don't like the approach we can also just wipe this PR 😄

If we agree on the general idea and the implementation I would go ahead and update all methods in `IntegerObject` and make the specs pass again!


seven1m (commented):
@richardboehme this looks like a good idea to me. Assuming we can still handle "fast integers" (Values) for certain "optimized" methods without instantiating GC-managed objects as we currently do, then this should work just fine. I don't foresee a bunch of overhead doing it this way, and the code to implement various methods should be much simpler, which I love.

The failing tests are a bit puzzling to me, though; I haven't dug into what's going on there...

richardboehme:
Thanks for your feedback on this @seven1m! I was thinking about the  "fast integer" concept that you're mentioning. Do you think it would be possible to store an instance of the new Integer class I introduce instead of a nat_int_t at the Value class? When calculating with the new Integer class those nat_int_ts will be converted to it anyway and it could handle bignums. However, I think it would increase the memory footprint of Value objects...

seven1m:
It's worth a try. I'm not sure of the perf penalty for doing that...

richardboehme:
I think this is ready for review now 👀

PR #511, opened 2022-02-24
Implement `observer` module

ryangjchandler:
This pull request introduces the `observer` module into Natalie and makes it spec-compliant too.

I'm not really sure what our takes are on implementing gems as first-party libraries - I know we already have some basic impl of the `strscan` gem too, so figured this would be okay.


seven1m:
Cool!

PR #513, opened 2022-02-24
Start implementing the `Base64` module

ryangjchandler:
This pull requests lays the groundwork for the `Base64` module.

Both the `Base64#encode_64` and `Base64#strict_encode64` methods are spec-compliant (minus encoding tests). The `Base64#urlsafe_encode64` method needs some work still.

I've also started work on `Array.pack("m")` which made implementing this module very simple. At the time of writing this PR description, 11 tests in the `m_spec.rb` file are now passing!

I'll definitely come back to implement the `decode_*` methods too, but this is a good start (I hope).

> NOTE: This code could definitely be optimised a little too, i.e. global arrays etc.

### DIFF
--- a/include/natalie/array_packer/packer.hpp
+++ b/include/natalie/array_packer/packer.hpp
@@ -37,6 +37,7 @@ namespace ArrayPacker {
                 case 'Z':
                 case 'h':
                 case 'H':
+                case 'm':
                 case 'u': {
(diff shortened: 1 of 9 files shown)


ryangjchandler:
Just realised that this was the wrong branch! 🤦‍♂️

PR #515, opened 2022-02-25
Parser: fix multi-assign that includes attribute assignment

FnControlOption:
For example:

```rb
a.b, c = 1, 2
```


seven1m:
Nice! You usually find a way to simplify the Parser node structure and I love it... :heart_eyes:

PR #518, opened 2022-02-28
Make `String#center` spec-compliant

ryangjchandler:
Feels like the code for this could be squashed a little - feels quite verbose... maybe it's okay since there's a little bit of math required to calculate the correct padding.

Related to #217.


ryangjchandler:
Ah, `to_int()` exists now. Nice!

PR #521, opened 2022-03-03
Make `String#ljust` spec-compliant

ryangjchandler:
As per, some encoding specs are commented out.

I think once this is merged in, I'll open a separate PR that adds a `to_str` method on `Object` so that we can abstract some of the conversion code (used in `String#center` too and afterwards `String#rjust`).

#217


ryangjchandler:
@seven1m Have updated this to use `Object#to_str` now.

PR #527, opened 2022-03-11
Implement Matchdata#to_a and #captures

timcraft:


src/match_data_object.cpp
```diff
@@ -2,6 +2,21 @@
 
 namespace Natalie {
 
+Value MatchDataObject::array(int start) {
+    auto size = (size_t)(m_region->num_regs - start);
```
seven1m, line 6: Should we throw a check in here (either an assertion or a raise) if the start is too big?

timcraft:
@seven1m The problem with adding that assertion is it'll fail when there are no groups in the regexp. For example:

    $ bin/natalie
    nat> /..\d+\d/.match("THX1138.").captures
    Assertion failed: (start < m_region->num_regs), function array, file match_data_object.cpp, line 6.

That case doesn't seem to be covered by ruby/spec but MRI behavior is to return an empty array:

    $ irb
    irb(main):001:0> /..\d+\d/.match("THX1138.").captures
    => []

The `MatchDataObject::array()` function is only called internally with either `0` or `1` as the start... an alternative could be to make it private to make it harder to accidentally misuse?

seven1m:
Ahh, very good. I misread the function, thinking it would segfault if the `start` was too high. I see now it just returns an empty array, which is great. I added a test and removed the assertion in cf8b4582cc5144f1f253f764324469e1f8cea34c. <3

PR #530, opened 2022-03-17
Implement Kernel#initialize_copy

timcraft:


seven1m:
I can't find much about this method -- is it just there as a hook for user programs to override and add functionality? (It doesn't seem to do anything otherwise.)

timcraft:
@seven1m It's an internal method that gets called by both #initialize_dup and #initialize_clone, which are hooks called by #dup and #clone. I don't think it's something you'd override very often in user programs, there are some stdlib classes which override it in MRI.

My intention is to next implement #initialize_dup and #initialize_clone, then we need to hook those into #dup and #clone to make it useful—I would have included more in this pull request but I think `Object::dup` needs re-implementing to make it spec compliant, which is a non trivial change.

seven1m:
Cool, sounds good. Thanks for the explanation. <3

PR #534, opened 2022-03-17
Implement Symbol#length and Symbol#size

timcraft:


src/symbol_object.cpp
```diff
@@ -85,6 +85,10 @@ bool SymbolObject::start_with(Env *env, Value needle) {
     return to_s(env)->internal_start_with(env, needle);
 }
 
+Value SymbolObject::length(Env *env) {
+    return to_s(env)->size(env);
```
seven1m, line 89: We can avoid an allocation of a StringObject on the heap...

```suggestion
    return Value::integer(m_name.length());
```

PR #535, opened 2022-03-17
Try to improve Integer#chr a bit

richardboehme:
There is still a lot to do here regarding encoding and strings. As soon as we have most of String implemented we can come back to this.

[#201]


seven1m:
Cool! :sunglasses:

PR #537, opened 2022-03-20
Implement String#[]= and MatchData#offset

richardboehme:
[#217]


richardboehme:
Hey @seven1m, can you figure out why those specs are failing? There seems to be a compilation error in one batch of specs related to symbols. I could fix this by adding `.force_encoding(Encoding::UTF_8)` in `lib/natalie/compiler/pass4.rb`.... I'm not sure how this is related to my changes. The only thing I can think of is that the combinations of specs that run in batch changed maybe?

seven1m:
Yeah I'll take a look. I will be removing the spec batching mechanism from the suite because I ran into a similar problem yesterday -- specs producing weird behavior when combined.

richardboehme:
> Looks like ungrouping the specs fixed it. Merge at will!

Thank you!

PR #538, opened 2022-03-22
Implement Symbol#name

timcraft:


seven1m:
This adds a bit of heft to every symbol and I'm not sure of the value. I understand you are making our Symbols "spec-compliant", but I wonder if the spec, in this case, is worth implementing. I would be inclined to skip the "returns same string instance" spec for now (and removing the extra, mostly-unused pointer on SymbolObject), and let's see, down the road, if that bites us somehow.

I have been writing Ruby for a number of years and I cannot think of a scenario where I needed `symbol.name` to return the same String object repeatedly.

Thoughts?

richardboehme:
@seven1m Do you really think adding a pointer to the object would be expensive (memory/performance-wise)? I'm not that C++-affine so I'd be happy to know why you consider this might be problematic 😅

timcraft:
> I have been writing Ruby for a number of years and I cannot think of a scenario where I needed symbol.name to return the same String object repeatedly.

I think it's more of a general optimization to reduce the number of String objects in memory than something that's of significant use to the caller. Symbols are immutable so their String representation won't change.

But yes, I think it's good to question these cases as they come up—less overhead is generally better, and we can always add it back in later if it becomes necessary/useful.

I'll update the PR to remove the pointer and return a new StringObject each time.

seven1m:
>I think it's more of a general optimization to reduce the number of String objects in memory than something that's of significant use to the caller. Symbols are immutable so their String representation won't change.

@timcraft makes perfect sense!

>Do you really think adding a pointer to the object would be expensive (memory/performance-wise)?

@richardboehme I did think so, because it seems wasteful to have a pointer that we 99.9% of the time never use. I did a little benchmark locally with `boardslam.rb` earlier and it was a teensy bit slower. But I just did it again with more warmup time and 50 runs each, and now I'm seeing no difference. I think I'm just going to merge as-is. Thanks for questioning my logic. :hugs:

PR #542, opened 2022-03-22
Make Integer#/ respect Rational objects

richardboehme:
This required some changes in RationalObject. I implemented `Rational#/`, `Rational#floor` and simplified the way we reduce fractions in RationalObject.


richardboehme:
The specs seem to fail for the exact same reason as in #537

PR #544, opened 2022-03-23
Start implementing the StringIO library

richardboehme:
This also includes some bug fixes and implementations of `IO#each`, `String#initialize_copy` and `Kernel#instance_of?`

Let me know if we should squash the commits related to StringIO!


seven1m:
This is awesome! Kinda think you're going for that commit count crown though... :joy:

PR #550, opened 2022-04-05
Start work on the CSV library 

richardboehme:
I started working on the CSV library for Natalie (implemented in pure Ruby). Sadly, the specs for this library are not that complete. Maybe we should start to create specs for it and contribute those to ruby/spec?

Currently the `CSV::Parser` class is not compatible with Ruby's `CSV::Parser` class. This is because it's really undocumented and I didn't want to reverse engineer it or look at the source code too much. It also has a note, that this class should only be used as an internal class. We can work on making it compatible later if we want to.

### DIFF
--- a/lib/csv.rb
+++ b/lib/csv.rb
@@ -1,0 +1,136 @@
+require 'stringio'
+require 'csv/parser'
+
+class CSV
+  DEFAULT_OPTIONS = {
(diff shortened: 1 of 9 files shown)


seven1m:
This is cool! 😎

Issue #551, opened 2022-04-15
Write CSV specs


richardboehme:
Due to [ruby/spec](https://github.com/ruby/spec) lacking specs for most of the CSV library, we want to provide our own specs (with the goal of upstreaming those to ruby/spec). Currently the following methods of CSV are implemented in Natalie and need specs:

- [ ] initialize
- [x] self.generate
- [x] self.generate_line
- [ ] self.open
- [x] self.parse
- [x] <<
- [x] add_row
- [x] col_sep
- [ ] each
- [ ] headers
- [x] liberal_parsing?
- [ ] lineno
- [ ] parser
- [ ] read
- [ ] readline
- [x] readlines
- [ ] rewind
- [x] row_sep
- [ ] shift

seven1m:
I think I might close this issue. It's a big broad and feels like it's not going to happen in my lifetime. I'm happy with where our CSV library is for now, and I'd rather just fix bugs as they come up. @richardboehme do you have thoughts?

richardboehme:
> I think I might close this issue. It's a bit broad and feels like it's not going to happen in my lifetime. I'm happy with where our CSV library is for now, and I'd rather just fix bugs as they come up. @richardboehme do you have thoughts?

I originally opened the issue because I felt the CSV implementation I built was pretty barebones and I wanted us to investigate more edge cases and eventually contribute those specs to ruby/spec. However I totally understand where you are coming from and I'd be fine with the issue being closed! :heart:

seven1m:
Thanks for the quick feedback. By all means, if you (or anyone else reading) feels ambitious and wants to add more specs for the csv library, please do! We can open smaller issues and PRs to address bits of it as we feel inspiration. Thanks @richardboehme!

herwinw:
I just ran the full upstream specs with Ruby 3.3.0-preview2 and noticed this warning:
```
warning: csv which will be not part of the default gems since Ruby 3.4.0
```
So I guess the CSV lib will be removed from the specs soon-ish. I don't think we should write new specs for it (which we weren't planning to do, since this case is closed)

seven1m:
Oh wow! Good to know.

PR #552, opened 2022-04-15
Start work on a CSV#<< (append) spec and fix some bugs in our implementation

seven1m:
[#551]

@richardboehme let me know if this is the sort of thing you had in mind...

These were my steps:

- Write the spec while running it through MRI, until I'm satisfied the edge cases are documented.
- Update our implementation until the spec passes with Natalie.


richardboehme:
That's exactly how I though about it as well. Actually I just started with the append spec myself, but you found some edge cases that I did not see yet 😅 Should we somehow "merge" our different specs we wrote?

I think the best approach is to write specifications for MRI and if we do not pass this spec we can either fix it directly or mark it as NATFIXME.

I'm not sure if we want to keep the new spec files in the `spec` directory or keep it separate for now (I thought about `test/lib/csv`)? This way we can keep it separate from the actual ruby/spec, but maybe it doesn't matter too much.

seven1m:
> Should we somehow "merge" our different specs we wrote?

Yeah, sure. Do you just want to push commits to this PR?

>I think the best approach is to write specifications for MRI and if we do not pass this spec we can either fix it directly or mark it as NATFIXME.

Cool :+1: 

>I'm not sure if we want to keep the new spec files in the spec directory or keep it separate for now (I thought about test/lib/csv)? This way we can keep it separate from the actual ruby/spec, but maybe it doesn't matter too much.

Yeah I'm good with either. Actually `test/lib/csv` might be better.

richardboehme:
I integrated the things I was working on. I hope the re-arrangements of the spec file are fine! There is some work to be done with encoding and field converters. I did not touch those yet because I have no idea how they work with CSV. Also we need to implement header support for CSVs.

richardboehme:
Also writing to the CSV *should* happen in the CSV::Writer class but I did not implement this yet.

seven1m (commented):
Nice! Merge whenever you want.

richardboehme:
I'll merge this now, leaving Encoding as a fixme for later.

Issue #555, opened 2022-05-13
Implement methods of Range


richardboehme:
Our main approach is to work on one spec from [ruby/spec](https://github.com/ruby/spec/tree/master/core/range) at a time, implementing as much of the spec as possible, then check it off the list:

- [x] begin_spec.rb
- [x] bsearch_spec.rb
- [x] case_compare_spec.rb
- [x] clone_spec.rb
- [x] count_spec.rb
- [x] cover_spec.rb
- [x] dup_spec.rb
- [x] each_spec.rb
- [x] end_spec.rb
- [x] eql_spec.rb
- [x] equal_value_spec.rb
- [x] exclude_end_spec.rb
- [x] first_spec.rb
- [x] frozen_spec.rb
- [x] hash_spec.rb
- [x] include_spec.rb
- [x] initialize_spec.rb
- [x] inspect_spec.rb
- [x] last_spec.rb
- [x] max_spec.rb
- [x] member_spec.rb
- [x] min_spec.rb
- [x] minmax_spec.rb
- [x] new_spec.rb
- [x] percent_spec.rb
- [x] range_spec.rb
- [x] size_spec.rb
- [x] step_spec.rb
- [x] to_a_spec.rb
- [x] to_s_spec.rb

richardboehme:
I think we have completed this so far. There are some issues with the Time Object left, and also there is a bug with our eval implementation but otherwise we are pretty green on this!

![image](https://user-images.githubusercontent.com/61323346/192139388-11eb3747-1434-4c59-bb58-b4daced8928f.png)

seven1m:
Nice! 🎉

timcraft:
Awesome! I've got Time#+ working which will help with those remaining spec failures, looks like we'll also need to implement Float#to_r for min_spec.rb and max_spec.rb

PR #558, opened 2022-05-28
Make Range#include? spec compliant

richardboehme:
This revealed a small bug in our `String#succ` method that I was not able to fix. Maybe somebody has an idea:
```ruby
"\u09b9".succ
# => MRI: "\u09b6\u09b6"
# => Natalie: "\u09ba"
```

[#555]


seven1m:
Ok, well we have some work to do! 😅

PR #559, opened 2022-05-28
Make Range#=== spec compliant by aliasing it to #cover?

richardboehme:
[#555]


seven1m (commented):
We should remove the binding for Range#=== in binding_gen.rb, yes?

seven1m (approved):
We should remove the binding for Range#=== in binding_gen.rb, yes?

richardboehme:
> We should remove the binding for Range#=== in binding_gen.rb, yes?

Yeah of course!

PR #561, opened 2022-06-03
Delay building an exception backtrace array

richardboehme:
Previously, we were building a backtrace array when initializing an
exception object. Sometimes though we do not want to print the backtrace
which makes generating it useless.

This patch tries to delay building the backtrace array, until it will
actually be used.

This should impact a lot of code slightly, because exceptions are
sometimes used for control-flow in the language itself (LocalJumpError
for example).

Benchmark:

```ruby
10000.times do |n|
  begin
    raise Exception, 'message'
  rescue Exception
  end
end
```

On MRI: 
```
real    0m0.100s
user    0m0.080s
sys     0m0.020s
```

On Natalie (master):
```
real    0m0.247s
user    0m0.237s
sys     0m0.010s
```

On Natalie (with this patch):
```
real    0m0.190s
user    0m0.189s
sys     0m0.000s
```

### DIFF
--- a/include/natalie/backtrace.hpp
+++ b/include/natalie/backtrace.hpp
@@ -1,0 +1,36 @@
+#pragma once
+
+#include "natalie/forward.hpp"
+#include "natalie/gc.hpp"
+#include "tm/hashmap.hpp"
+
+namespace Natalie {
+class Backtrace : public Cell {
+public:
+    struct Item {
+        const ManagedString *source_location;
+        const char *file;
+        size_t line;
+    };
+
+    void add_item(Env *env, const char *file, size_t line) {
+        // NATFIXME: Ideally we could store the env and delay building the code location name
+        m_items.push(Item { env->build_code_location_name(), file, line });
+    }
+    ArrayObject *to_ruby_array();
+
+    virtual void visit_children(Visitor &visitor) override final {
+        for (auto item : m_items) {
+            visitor.visit(item.source_location);
+        }
+    }
+
+    virtual void gc_inspect(char *buf, size_t len) const override {
+        snprintf(buf, len, "<Backtrace %p size=%ld>", this, m_items.size());
+    }
+
+private:
+    Vector<Item> m_items;
+};
+
+}
--- a/src/kernel_module.cpp
+++ b/src/kernel_module.cpp
@@ -54,7 +54,8 @@ Value KernelModule::binding(Env *env) {
 }
 
 Value KernelModule::caller(Env *env) {
-    auto ary = env->backtrace();
+    auto backtrace = env->backtrace();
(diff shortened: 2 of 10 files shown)


include/natalie/backtrace.hpp
```diff
@@ -0,0 +1,36 @@
...
+class Backtrace : public Cell {
+public:
+    struct Item {
+        const ManagedString *source_location;
+        const char *file;
+        size_t line;
+    };
+
+    void add_item(Env *env, const char *file, size_t line) {
+        // NATFIXME: Ideally we could store the env and delay building the code location name
```
richardboehme, line 17: I tried doing exactly this and I notices some really weird behavior with it:

The Env object seems to be fine, but the Method object that is references on the Env ends up being 0x10 which will crash the program. I suspect the GC or something like this breaks it?

src/kernel_module.cpp
```diff
@@ -54,7 +54,9 @@ Value KernelModule::binding(Env *env) {
 }
 
 Value KernelModule::caller(Env *env) {
-    auto ary = env->backtrace();
+    auto backtrace = env->backtrace();
+    auto ary = backtrace->to_ruby_array();
+    backtrace->mark();
```
richardboehme, line 59: I'm not sure if this is a great way to do this. Basically I want to tell the GC that this object can be sweeped at any time from there on. Is this the way to do this? (Maybe we should rethink allocating a GC-managed Backtrace object in this case?)

include/natalie/backtrace.hpp
seven1m: Well that's not cool. :cry:

src/kernel_module.cpp
seven1m: Actually, that has the opposite effect, but only for the next mark/sweep cycle. "Mark" in the GC means "this object is in use" (during this cycle). Once the GC mark/sweep cycle is over, all objects get unmarked. And a subsequent cycle should clean it up.

Probably best to just remove the `mark()` line here and let the GC do its job. :smile:

seven1m:
This is pretty cool! :tada:

src/kernel_module.cpp
richardboehme: Oops, I'm going to remove it 😅

include/natalie/backtrace.hpp
seven1m: Some of our Envs are stack-allocated and short-lived. Others are heap-allocated and get cleaned up by the GC. You can make a copy of a single Env, but that wouldn't fix the issue of other Envs in the call stack getting overwritten/cleaned up. I think this is the best we can do.

PR #562, opened 2022-06-05
Run tests in parallel on CI

seven1m:
I want to see if this has any effect...


seven1m:
Well that didn't help any. :laughing:

PR #563, opened 2022-06-05
Experiment with building shared object instead of static archive

seven1m:
The idea would be to use a shared object when running a script directly.

It seems to speed up compilation when you don't care about the temporary
binary.

TODO: When using `-c out` we will still need to use the static-compilation
method so that the binary is fully self-contained.


seven1m:
I will come back to this at a later date. :smile:

PR #565, opened 2022-06-06
Implement Range#min, #max and #minmax

richardboehme:
This also fixes some issues with our spec library and a bug with implicit block passing for `super`.

[#555]

### DIFF
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -833,7 +833,12 @@ class Stub
 
   def validate!
     unless @count_restriction == nil || @count_restriction === @count
-      raise SpecFailedException, "#{@subject.inspect} should have received #{@message}"
+      message = "#{@subject.inspect} should have received ##{@message}"
+      if @count_restriction != nil
+        message << " #{@count_restriction} time(s) but received #{@count} time(s)"
+      end
+
+      raise SpecFailedException, message
     end
   end
 end
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -1139,8 +1139,9 @@ def run_specs
           end
         end
         $stub_registry.reset
-        context.each { |con| con.before_each.each { |b| b.call } }
         $expectations = []
+        context.each { |con| con.before_each.each { |b| b.call } }
+
         fn.call
 
         $expectations.each { |expectation| expectation.validate! }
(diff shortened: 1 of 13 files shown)


test/support/spec.rb
```diff
@@ -763,11 +791,11 @@ def at_least(n)
     # FIXME: support endless ranges
```
richardboehme, line 791: Reminder: Remove this line

test/support/spec.rb
```diff
@@ -763,11 +791,11 @@ def at_least(n)
...
     case n
     when :once
-      @count_restriction = 1..9_999_999
+      @count_restriction = 1..
     when :twice
-      @count_restriction = 2..9_999_999
+      @count_restriction = 2..
     else
-      @count_restriction = n..9_999_999
+      @count_restriction = n..
```
richardboehme, line 798: Those currently cause the CI to fail... I'm going to investigate the issue 😄

richardboehme: This seems to be related to NatalieParser. I opened natalie-lang/natalie_parser#4 with more information!

seven1m:
I pushed some fixes to NatalieParser... let's see if that helps!

richardboehme:
@seven1m I may need some help from you here... I tried debugging this one failing test when running with clang and I can't wrap my head around the issue. It seems to be GC related and I think it has something to do with ivars on the Float::CanCoerce object that gets used by the failing spec but I could not find the exact issue there. Do you have an idea?

seven1m:
Well somehow I broke it worse. 😅 I'll dig in some more later today...

seven1m:
I honestly don't know what I did. But at least I feel better that we have more guards in place to prevent capturing synthesized values in ivars. :man_shrugging:

seven1m:
@richardboehme this is good to merge now yes?

richardboehme:
> @richardboehme this is good to merge now yes?

Yeah I think it should be good. I just hadn't time to come back to this, because I'm on holiday right now.

seven1m:
Rebased against master. Once the tests pass, I'll merge it.

seven1m:
Ahhh, the error returned! (This is a good thing, maybe I'll figure it out now.)

PR #568, opened 2022-06-19
Implemented String#chomp

Duncan-Britt:
Created spec/core/string/chomp_spec.rb
Added binding for String#chomp to lib/natalie/compile/binding_gen.rb
Added function prototype to include/natalie/string_object.hpp
Added function implementation to src/string_object.cpp

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -138,6 +138,58 @@ Value StringObject::center(Env *env, Value length, Value padstr) {
     return new StringObject { result, m_encoding };
 }
 
+Value StringObject::chomp(Env *env, Value record_separator) {
+    const char *cstr = m_string.c_str();
+    size_t end_idx = m_string.length();
+
+    if (record_separator.is_null()) {
+        if (cstr[end_idx-1] == '\r') {
+            --end_idx;
+        } else if (cstr[end_idx-1] == '\n') {
+            if (cstr[end_idx-2] == '\r') {
+                --end_idx;
+            }
+            --end_idx;
+        }
+        
+        return new StringObject(m_string.substring(0, end_idx));
+    } 
+
+    record_separator->assert_type(env, Object::Type::String, "String");
+    StringObject *sep = record_separator->as_string();
+    const char *csep = sep->m_string.c_str();
+    size_t sep_len = sep->m_string.length();
+
+    if (sep_len == 0) {
+        while (end_idx > 0 && cstr[end_idx - 1] == '\n') {
+            if (end_idx > 1 && cstr[end_idx-2] == '\r') {
+                --end_idx;
+            }
+            --end_idx;
+        }
+
+        return new StringObject(m_string.substring(0, end_idx));
+    }
+
+    size_t last_idx = end_idx - 1;
+    size_t sep_idx = sep_len - 1;
+    while (cstr[last_idx] == csep[sep_idx]) {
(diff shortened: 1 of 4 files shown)


src/string_object.cpp
```diff
@@ -138,6 +138,58 @@ Value StringObject::center(Env *env, Value length, Value padstr) {
...
+    const char *csep = sep->m_string.c_str();
+    size_t sep_len = sep->m_string.length();
+
+    if (sep_len == 0) {
+        while (end_idx > 0 && cstr[end_idx - 1] == '\n') {
+            if (end_idx > 1 && cstr[end_idx-2] == '\r') {
+                --end_idx;
+            }
+            --end_idx;
+        }
```
seven1m, line 169: I noticed this block of code is similar to the one starting on 145, but not quite the same. If I'm reading it correctly, passing an empty string as the separator is the same as passing no record separator. There shouldn't be a reason the code has to repeat, so we can share the same code if we rearrange things. Using a lambda to share code is fine, or checking for a null/empty separator up top is fine too.

```sh
→ cat > wat.rb
s = "foo\n\r"
p s.chomp("")

→ ruby wat.rb
"foo\n\r"

→ bin/natalie wat.rb
"foo\n\u000D"
```

seven1m (changes requested):
Thanks for your work on this!

I spent quite a bit of time staring at and playing with this code, trying to get it to break. I was certain that using C strings throughout was going to be buggy since Ruby allows null character (`\0`) in the middle of strings. But alas, I only found the one bug with the empty separator, as mentioned in the other comment.

Just be careful in the future dealing with C strings, as they terminate on null character, while Ruby strings do not. We do have older code that deals with C strings, which is a mistake -- we have learned to avoid them in newer code. Try to use the TM::String class instead, if possible. (If not possible, we might need to augment the class.)

In this one case, using a C string seems to be fine because you were careful to only work from the right side, which is great.

src/string_object.cpp
Duncan-Britt: There are actually two differences between passing an empty string and  passing no record separator. With an empty string, \r by itself is not removed unless \n comes after. Also, with an empty string, all of the trailing separators are removed instead of just the last. 
i.e. `"hello\r\n\r\n".chomp => "hello\r\n"; "hello\r\n\r\n".chomp('') => "hello"`

Duncan-Britt: I'm glad the null character didn't produce any unexpected results! To be honest I hadn't thought of that, and was working with c strings only because I was unfamiliar with TM::String, but in the future I will be sure to use TM::String wherever I can.

Duncan-Britt:
I'm glad the null character didn't produce any unexpected results! To be honest I hadn't thought of that, and was working with c strings only because I was unfamiliar with TM::String, but in the future I will be sure to use TM::String wherever I can.

src/string_object.cpp
seven1m: OK, very good. Thanks for explaining that.

Where my confusion came in was seeing different output in MRI vs Natalie, but that's just a separate bug we can fix later, i.e. `"\r"` is printed as `"\u000D"`.

seven1m (approved):

Duncan-Britt:
Egg on my face, but at least now I know! Maybe adding the ruby/spec for String#chomp will be my next task.

PR #573, opened 2022-07-01
Change how we optimize for "fast" ints/floats

seven1m:
First, a few definitions:

- "fast" int/float: the int/float value is stored in the Value pointer itself, rather than in a heap-allocated, GC-managed object
- "synthesis": creating a temporary IntegerObject or FloatObject that lives on the stack so that normal Ruby Object functionality works as expected
- "optimized" method: a method that uses such synthesized Objects without allowing them to escape their call stack, i.e. they are not given back in the return value, not set in a variable of any kind, and not wrapped in a data structure such as ArrayObject or HashObject.

Changes:

1. By marking each Method object with a boolean `m_optimized` and moving the synthesis of arguments into Method::call(), we can eliminate the Hashmap `s_optimized_methods` and the needed lookup. We can just check Method::is_optimized().
2. Next, by adding assertions at every point where a synthesized object might be captured outside its call stack, i.e. variable setting, array object "push", etc., we can be confident that our methods use the synthesized Objects as expected.
3. Last, we can eliminate a bunch of code that dealt solely with fast ints/floats in an effort to keep them pure. That is no longer needed!


seven1m:
This is a bit slower than it should be. It seems that some of the slowdown is from creating all those Hashmaps for `m_ivars` on Object. You cannot even have instance variables on an int or float, so I'm going to see about changing how we allocate `m_ivars`.

seven1m:
In release mode, this new branch is just as fast when running `fib.rb`! :tada: 

I'm stoked about this because of all the code we get to remove. :hugs:

PR #574, opened 2022-07-06
Pass block in Args object

seven1m:


seven1m:
Now that I see the results, I don't think I like this. Closing for now.

PR #575, opened 2022-07-06
Implement String's empty method

AI-Mozi:
[#217]

### DIFF
--- a/include/natalie/string_object.hpp
+++ b/include/natalie/string_object.hpp
@@ -193,6 +193,7 @@ public:
     Value to_i(Env *, Value) const;
     Value upcase(Env *);
     Value uplus(Env *);
+    bool empty(Env *);
 
     Value convert_float();
 
--- a/lib/natalie/compiler/binding_gen.rb
+++ b/lib/natalie/compiler/binding_gen.rb
@@ -896,6 +896,7 @@ gen.binding('String', 'to_s', 'StringObject', 'to_s', argc: 0, pass_env: false, 
 gen.binding('String', 'to_str', 'StringObject', 'to_str', argc: 0, pass_env: false, pass_block: false, return_type: :Object)
 gen.binding('String', 'to_sym', 'StringObject', 'to_sym', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('String', 'upcase', 'StringObject', 'upcase', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
+gen.binding('String', 'empty', 'StringObject', 'empty', argc: 0, pass_env: true, pass_block: false, return_type: :bool)
 
 gen.undefine_singleton_method('Symbol', 'new')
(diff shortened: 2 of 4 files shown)


lib/natalie/compiler/binding_gen.rb
```diff
@@ -901,6 +901,7 @@ def generate_name
 gen.binding('String', 'to_str', 'StringObject', 'to_str', argc: 0, pass_env: false, pass_block: false, return_type: :Object)
 gen.binding('String', 'to_sym', 'StringObject', 'to_sym', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('String', 'upcase', 'StringObject', 'upcase', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
+gen.binding('String', 'empty', 'StringObject', 'empty', argc: 0, pass_env: true, pass_block: false, return_type: :bool)
```
seven1m, line 904:
```suggestion
gen.binding('String', 'empty?', 'StringObject', 'is_empty', argc: 0, pass_env: false, pass_block: false, return_type: :bool)
```

You can bind directly to the existing method instead of creating a new one.

src/string.rb
```diff
@@ -1,5 +1,6 @@
 class String
   alias replace initialize_copy
+  alias empty? empty
```
seven1m, line 3: Not needed!

src/string_object.cpp
```diff
@@ -1176,6 +1176,14 @@ Value StringObject::uplus(Env *env) {
...
 }
 
+bool StringObject::empty(Env *env) {
+    if (this->is_empty()) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
```
seven1m, line 1186: Not needed!

include/natalie/string_object.hpp
```diff
@@ -204,6 +204,7 @@ class StringObject : public Object {
     Value to_i(Env *, Value) const;
     Value upcase(Env *);
     Value uplus(Env *);
+    bool empty(Env *);
```
seven1m, line 207: Not needed!

seven1m (changes requested):
Well, you picked an easy one! You just need to add the one line in binding_gen.rb and let the binding do the rest.

Also, be sure to copy in the relevant spec from [ruby/spec](https://github.com/ruby/spec/blob/master/core/string/empty_spec.rb) and make sure it passes.

AI-Mozi:
Tests from ruby/spec were already copied :)

AI-Mozi (commented):
Ohhh.. now I noticed that empty? method is done at line 868..

seven1m:
>Ohhh.. now I noticed that binding to empty? method is done at line 868..

Indeed, it is! Sorry I didn't check that. Well, this was great practice. I have checked off the `empty_spec.rb` item from #217. Sorry about that.

I hope you will pick another to work on! <3

PR #577, opened 2022-07-07
Add String#delete_prefix and #delete_prefix! methods

AI-Mozi:
[#217]


seven1m:
The failing build here is not related (directly) to your change, but rather a bug exposed in our GC, which I am working on. Once I get that fixed, I'll update this PR and merge it!

PR #579, opened 2022-07-14
Implement String#ascii_only?

AI-Mozi:
[#217]


seven1m (approved):
I hesitated on this one because I worry there will be an encoding where the simple `> 127` check won't work. But since we don't have an encoding where that can be tested, I think this is fine for now.

PR #582, opened 2022-08-08
Promote Compiler2, Remove old compiler

seven1m:
This is the final push to replace our original AST-manipulating compiler with the new IR-based one.

Major benefits:
1. The Intermediate Representation (IR) produced by the compiler is flat and agnostic from the backend.
2. There is support for multiple backends: The C++ compiler backend, which is ready in this PR, and the interpeter (VM) backend, which is an early proof-of-concept prototype.
3. The IR is easier to optimize. We can add one or more compiler passes in the future to optimize instructions and produce faster programs.
4. Debugging the IR is easier than debugging generated C++ code.
5. The new compiler produces faster binaries than the old compiler!

fib

boardslam


seven1m:
When I started writing the `src/*.rb` files with the new compiler, some new bugs were exposed. :sweat_smile: I'll work on getting those fixed up in this PR and merge when all the tests pass...

PR #586, opened 2022-08-17
Cleanup of Value

davidot:
This is a first step toward smaller values, although we will need to compromise on u64 values somehow...

It has been a long while since I did anything on this project, so feel free to roast the code as much as you want

### DIFF
--- a/include/natalie/value.hpp
+++ b/include/natalie/value.hpp
@@ -65,9 +65,9 @@ public:
                 return get_fast_integer() == other.get_fast_integer();
             return false;
         }
-        if (is_fast_float()) {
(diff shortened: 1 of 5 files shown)


seven1m (approved):
@davidot I think this looks OK. Do you intend to make similar changes for fast integers? If so, do you want to do it in this PR or in a separate PR?

davidot:
> @davidot I think this looks OK. Do you intend to make similar changes for fast integers? If so, do you want to do it in this PR or in a separate PR?

Yeah I'll continue in another PR, just wanted to keep the PRs on the smaller side

PR #587, opened 2022-08-19
Make Range#to_a spec-compliant

richardboehme:
I don't know about the code-duplication in #iterate_over_string_range and #iterate_over_symbol_range. We could add some kind of macro or maybe there is another idea?

I also was not sure about the location of StringUptoIterator. I decided for a new class but am open to other suggestions. If we decided on the approach we can adjust String#upto to use the iterator as well.

[#555]

### DIFF
--- a/src/string_upto_iterator.cpp
+++ b/src/string_upto_iterator.cpp
@@ -10,6 +10,7 @@ const TM::String StringUptoIterator::peek() const {
     if (m_treat_like_integer)
         return TM::String((Integer(m_string) + 1).to_nat_int_t());
 
+    // special handling for single characters
     if (m_string == "9")
         return ":";
     else if (m_string == "Z")
--- a/src/string_upto_iterator.cpp
+++ b/src/string_upto_iterator.cpp
@@ -1,0 +1,22 @@
+#include "natalie/string_upto_iterator.hpp"
+#include "natalie/integer.hpp"
+
+namespace Natalie {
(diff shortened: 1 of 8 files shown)


src/string_upto_iterator.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace Natalie {
+const TM::String StringUptoIterator::next() {
+    return m_string = peek();
+}
+
+const TM::String StringUptoIterator::peek() const {
+    if (m_treat_like_integer)
+        return TM::String((Integer(m_string) + 1).to_nat_int_t());
+
+    if (m_string == "9")
```
seven1m, line 13: Let's add a comment above this so we remember this is special single-char logic.

src/string_upto_iterator.cpp
```diff
@@ -0,0 +1,22 @@
...
+        return ":";
+    else if (m_string == "Z")
+        return "[";
+    else if (m_string == "z")
+        return "{";
```
seven1m, line 18: This same `upto` logic is in [src/string.rb](https://github.com/natalie-lang/natalie/blob/master/src/string.rb) ... how could we consolidate?

seven1m: Nevermind! I see you mentioned that in the PR description. Let's merge this and plan to use it from StringObject somehow, later...

seven1m (approved):
Just one change. Thanks Richard!

Issue #598, opened 2022-09-01
Implement TCPServer and TCPSocket


seven1m:
I'd like to eventually host https://natalie-lang.org with Natalie, so we'll need some socket stuff!

Here is the initial program I'd like to get running: https://gist.github.com/seven1m/9d18479b4e0ddd5fa71b06a8b6bb166b From there I think we can serve up a static file rather than the 'hello world' string.

I'll add more notes to this issue as I learn more... :hugs:

Checklist (growing as we learn):
- [ ] Socket.tcp
  - [x] `spec/library/socket/socket/tcp_spec.rb`
- [ ] Addrinfo
  - [x] `spec/library/socket/addrinfo/initialize_spec.rb`
  - [x] `spec/library/socket/addrinfo/ip_address_spec.rb`
  - [x] `spec/library/socket/addrinfo/ip_port_spec.rb`
  - [x] `spec/library/socket/addrinfo/to_sockaddr_spec.rb`
- [ ] BasicSocket
  - [x] `spec/library/socket/basicsocket/setsockopt_spec.rb`
  - [x] `spec/library/socket/basicsocket/getsockopt_spec.rb`
  - [x] `spec/library/socket/basicsocket/getsockname_spec.rb`
  - [x] `spec/library/socket/basicsocket/recv_spec.rb`
- [ ] Socket
  - [x] `spec/library/socket/socket/getaddrinfo_spec.rb`
  - [x] `spec/library/socket/socket/accept_spec.rb`
- [ ] IPSocket
  - [x] `spec/library/socket/ipsocket/addr_spec.rb`
- [ ] TCPServer
  - [x] `spec/library/socket/tcpserver/new_spec.rb`
  - [x] `spec/library/socket/tcpserver/accept_spec.rb`
- [ ] TCPSocket
  - [x] `spec/library/socket/tcpsocket/initialize_spec.rb`

ryangjchandler:
Perhaps we could add a new compiler macro that lets you embed the contents of a file in an executable, similar to Rust's `include_str!()` macro or `#include` in CXX.

seven1m:
@ryangjchandler that's sick!

TheGrizzlyDev:
> That way it really is self-contained and there's no need for file I/O at runtime..
You'd still have the test results, right? Regardless, it's such a great idea!

ryangjchandler:
@TheGrizzlyDev Yeah, I'm mainly talking about the landing / home page of the site.

[...]

PR #599, opened 2022-09-01
Start work on Socket Addrinfo#initialize

seven1m:
This is in support of #598.

Anyone can help!


seven1m:
I'm going to merge this into master so we can do our normal incremental workflow. :hugs:

PR #601, opened 2022-09-02
Make String#upto use the StringUptoIterator

richardboehme:
This also changed the StringUptoIterator-API a bit. It is now capable of performing endless iterations (like before) and iterations to a specific `to`-String.


seven1m (approved):
Nice!

PR #602, opened 2022-09-02
Implement Addrinfo#protocol

richardboehme:
**Note**: I set the base branch to be @seven1m's addrinfo-init branch. Not sure if we want to manage it like that or if we want to merge `addrinfo-init` and merge this branch into main as well?

[#598]


richardboehme:
I'm not sure if it really just returns the protocol passed into the initialize method, but it seems like it works? I tried receiving it from the addrinfo but with this change the first spec (return 0 as default) does not work because it figures out the correct protocol.

PR #603, opened 2022-09-02
Implement Addrinfo#ip_address

richardboehme:
[#598]

### DIFF
--- a/lib/socket.rb
+++ b/lib/socket.rb
@@ -260,9 +260,29 @@ class Addrinfo
       return Value::integer(info_struct->ai_family);
   END
 
-  def ip_address
-    # TODO
-  end
+  __define_method__ :ip_address, <<-CPP
+      auto addrinfo = self->ivar_get(env, "@_addrinfo"_s);
+      if(addrinfo->is_nil())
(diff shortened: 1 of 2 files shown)


lib/socket.rb
```diff
@@ -263,9 +263,30 @@ def protocol
...
+            inet_ntop(AF_INET, &(sockaddr->sin_addr), address, INET_ADDRSTRLEN);
+            return new StringObject { address };
+          }
+          case AF_INET6: {
+            char address[INET6_ADDRSTRLEN];
+            auto *sockaddr = (struct sockaddr_in6 *)info_struct->ai_addr;
+            inet_ntop(AF_INET6, &(sockaddr->sin6_addr), address, INET6_ADDRSTRLEN);
+            return new StringObject { address };
+          }
+      }
```
richardboehme, line 286: Should we cache those values that we retrieve from the addrinfo object? They cannot change in the object's lifetime, can they?

seven1m: I don't think it's very costly to look them up, but maybe I'm wrong. We can add caching later if we find there's a bottleneck.

PR #604, opened 2022-09-02
Try to fix Array#bsearch

richardboehme:
I feel like (correct me if I'm wrong) that the previous implementation
wasn't a binary search. It just moved the window used to determine the
middle of the array by one. However a binary search should continue it's
search in the left/right half of the array. This change also improves
performance of the bsearch method:

```ruby
p (0..100_000).to_a.bsearch { |x| x >= 73 }
```

master:
```
$ time ./find-min
73

real    0m0.524s
user    0m0.513s
sys     0m0.011s
```

with fix:
```
$ time ./find-min-2
73

real    0m0.455s
user    0m0.414s
sys     0m0.041s
```

When measuring the iteration count in the while-loop we can see that it
dropped from 99856 using master to only 17.

### DIFF
--- a/src/array_object.cpp
+++ b/src/array_object.cpp
@@ -1450,27 +1450,29 @@ Value ArrayObject::bsearch_index(Env *env, Block *block) {
         }
 
         if (outcome->is_numeric()) {
+            // Find-any mode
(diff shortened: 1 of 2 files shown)


seven1m:
Screenshot from 2022-09-02 21-52-29

Nice! :tada:

PR #606, opened 2022-09-03
Make `String#clone` spec compliant

ryangjchandler:
Related to #217.

This will actually help pass some other specs too that also test the `#clone` method I imagine - I haven't identified them though.


ryangjchandler:
Seeing now that the `Array#clone` spec is failing - since `String#clone` has it's own extra logic, perhaps we should override this method on the `StringObject` instead.

ryangjchandler:
Replaced by #610.

PR #607, opened 2022-09-03
Implement Range#bsearch

richardboehme:
This also adds some binary search helpers that can be used with Array#bsearch as well.

[#555]

### DIFF
--- a/include/natalie/bsearch.hpp
+++ b/include/natalie/bsearch.hpp
@@ -1,0 +1,17 @@
+#pragma once
+
+#include "natalie.hpp"
+
+namespace Natalie {
+enum class BSearchCheckResult {
+    SMALLER,
+    BIGGER,
+    EQUAL
+};
+
+BSearchCheckResult binary_search_check(Env *, Value);
+Optional<nat_int_t> binary_search(Env *, nat_int_t, nat_int_t, std::function<Value(nat_int_t)>);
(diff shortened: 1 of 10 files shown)


richardboehme:
Also I notices binary searching in ruby is pretty weird...

I mean what's that?
```
3.0.4 :004 > (...10).bsearch { |x| x == 10 }
 => 10
3.0.4 :005 > (1...10).bsearch { |x| x == 10 }
 => nil
```

seven1m:
This is also weird to me:

```ruby
irb(main):008:0> (..20).bsearch { |x| x == 10 }
=> nil
irb(main):009:0> (1..20).bsearch { |x| x == 10 }
=> 10
```

seven1m (approved):

richardboehme:
> ```ruby
> (..20).bsearch { |x| x == 10 }
> ```

Yeah... It's confusing but it seems like it's wanted behavior... See the [binary search documentation](https://github.com/ruby/ruby/blob/master/doc/bsearch.rdoc). When using `x == 10` we will be in find-minimum mode.

With a beginless range it searches the first value that evaluates to false (in this case it will find 19), because according to the find-minimum documentation everything before the first match must return false. The first object it finds that evaluates to false will be the left limit of the binary search which is why it will search between 19 and 20 and there is no 10 there so it will not find it.

PR #619, opened 2022-09-16
Implement Enumerator::ArithmeticSequence

richardboehme:
It took me way to longer to do this than it should have 😅 

This is progress towards `Range#step` and also makes `Numeric#step` spec-compliant.

@seven1m Please have a look at the modifications I did to the `__call__` macro. I'm not sure if I used it correctly and if my changes are good to merge.

### DIFF
--- a/include/natalie/object_type.hpp
+++ b/include/natalie/object_type.hpp
@@ -9,6 +9,8 @@ enum class ObjectType {
     Class,
     Complex,
     Encoding,
+    Enumerator,
+    EnumeratorArithmeticSequence,
     Exception,
     False,
     Fiber,
--- a/lib/natalie/compiler/instructions/inline_cpp_instruction.rb
+++ b/lib/natalie/compiler/instructions/inline_cpp_instruction.rb
@@ -51,6 +51,10 @@ module Natalie
           case type
           when 'double'
             "#{value}.as_double()"
+          when 'bool'
+            "#{value}->is_truthy()"
+          when 'Value'
+            value
           else
             raise "I don't yet know how to cast arg type #{type}"
           end
@@ -60,6 +64,8 @@ module Natalie
           case type
           when 'double'
             "Value::floatingpoint(#{value})"
+          when 'Value'
+            value
           else
             raise "I don't yet know how to cast return type #{type}"
           end
--- a/lib/natalie/compiler/instructions/inline_cpp_instruction.rb
+++ b/lib/natalie/compiler/instructions/inline_cpp_instruction.rb
@@ -71,9 +71,21 @@ module Natalie
           end
         end
 
-        casted_args = args.each_with_index.map do |arg, index|
+        casted_args = []
+
+        if fn[:args][0] == 'Env *'
+          # Push the env directly. This allows us to ommit it from the __call__
+          # macro call.
+          casted_args << 'env'
+        end
+
+        args.each_with_index do |arg, index|
+          if fn[:args][0] == 'Env *'
+            index += 1
+          end
+
           type = fn[:args][index]
-          cast_value_to_cpp.(arg, type)
+          casted_args << cast_value_to_cpp.(arg, type)
         end
 
         transform.exec_and_push(
--- a/src/enumerator/arithmetic_sequence_object.cpp
+++ b/src/enumerator/arithmetic_sequence_object.cpp
@@ -12,8 +12,8 @@ Integer ArithmeticSequenceObject::calculate_step_count(Env *env) {
         step_count = n->send(env, "+"_s, { n->send(env, "*"_s, { new FloatObject { std::numeric_limits<double>::epsilon() } }) })->send(env, "floor"_s)->as_integer()->integer();
     }
 
-    if (exclude_end() && step_count > 0)
-        step_count -= 1;
+    if (!exclude_end())
+        step_count += 1;
 
     return step_count;
 }
@@ -91,5 +91,12 @@ Value ArithmeticSequenceObject::last(Env *env) {
     } else {
         return m_end;
     }
+}
+
+Value ArithmeticSequenceObject::size(Env *env) {
+    if (m_end->send(env, "infinite?"_s)->is_truthy())
+        return FloatObject::positive_infinity(env);
+
+    return IntegerObject::create(step_count(env));
 }
 };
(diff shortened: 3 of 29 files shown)


include/natalie/object_type.hpp
```diff
@@ -4,11 +4,13 @@ namespace Natalie {
 
 enum class ObjectType {
     Nil,
+    EnumeratorArithmeticSequence,
```
seven1m, line 7: Can this be in alphabetical order with the rest? (Nil is at the top so it has a zero value.)

lib/natalie/compiler/instructions/inline_cpp_instruction.rb
```diff
@@ -60,14 +64,28 @@ def generate_call(transform, fn_name, *args)
...
+        end
+
+        args.each_with_index do |arg, index|
+          if fn[:args][0] == 'Env *'
+            index += 1
+          end
+
           type = fn[:args][index]
-          cast_value_to_cpp.(arg, type)
+          casted_args << cast_value_to_cpp.(arg, type)
```
seven1m, line 88: The changes you made for `__call__` look good to me!

src/enumerator/arithmetic_sequence_object.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include "natalie.hpp"
+
+namespace Natalie::Enumerator {
+ArithmeticSequenceObject::ArithmeticSequenceObject(Env *env, Origin origin, Value begin, Value end, Value step, bool exclude_end)
+    : ArithmeticSequenceObject {} {
+    m_origin = origin;
+    m_begin = begin;
+    m_end = end;
+    m_exclude_end = exclude_end;
+    m_step = step;
```
seven1m, line 11: Can we use the member initializer syntax for these members, like we do in the other classes?

```suggestion
    : Object { fetch_nested_const({ "Enumerator"_s, "ArithmeticSequence"_s }) }
    , m_origin { origin }
    , m_begin { begin }
    , m_end { end }
    , m_exclude_end { exclude_end }
    , m_step { step } {
```

src/enumerator/arithmetic_sequence_object.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+Value ArithmeticSequenceObject::each(Env *env, Block *block) {
+    if (!block)
+        return this;
+
+    iterate(env, [&env, &block](Value value) -> Value {
+        // I'm not actually sure how breaking works here. I would have expected
+        // it to not work but it somehow does?
+        // In my mind this should return the breaked result to the lambda instead
+        // of the each function.
```
seven1m, line 66: Breaking works by matching a "break point" and that's handled by the `CppBackend`.

src/enumerator/arithmetic_sequence_object.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+        if (count < 0)
+            env->raise("ArgumentError", "negative array size");
+
+        if (n_as_int->integer() > steps)
+            count = steps;
+
+        n_as_int->assert_fixnum(env);
+
+        // NATFIXME: What if we exceed size_t?
```
seven1m, line 208: Actually, the real danger is underflowing a size_t. The `count` integer could be negative (unless there is a check for that I'm missing). When casted to a size_t, the number can wrap around.

Let's throw an assert right here to ensure the `count` is positive.

(Also, `size_t` is a 64-bit integer, which presumably is big enough to represent every bye of memory any computer has or will have for a long time. :smile:)

src/numeric.rb
```diff
@@ -158,71 +162,17 @@ def step(to_pos = nil, by_pos = nil, by: nil, to: nil)
...
-        ascending ? n : -n
-      else
-        n = (to - self) / by
-        (n + n * Float::EPSILON).floor
-      end
-
-    if step_count.negative?
-      return self
-    end
+    enumerator = __call__('Enumerator::ArithmeticSequenceObject::from_numeric', self, to, by)
```
seven1m, line 172: You might want to look at the new `__bind__` macro, which serves a similar purpose, but would allow you to call it like a normal Ruby method. The `__call__` macro was originally designed for calling C APIs.

seven1m (changes requested):
This is amazing work! I just had a few little things that I noticed, but overall this is really really cool.

src/enumerator/arithmetic_sequence_object.cpp
```diff
@@ -0,0 +1,261 @@
...
+    if (m_begin->send(env, cmp, { m_end })->is_truthy())
+        return 0;
+
+    if (_step->is_float() || _begin->is_float() || _end->is_float()) {
+        _step = _step->to_f(env);
+        _begin = _begin->to_f(env);
+        _end = _end->to_f(env);
+    }
+
+    auto n = _end->send(env, "-"_s, { _begin })->send(env, "/"_s, { _step->to_f(env) });
```
seven1m, line 40: Oh yeah, I noticed you're using `->send()` a lot here. Let's try to use `.send()` instead, just for consistency. I'm afraid if we get in the habit of bypassing the `Value::send()` machinery, we'll lose some of our fast int/float optimizations accidentally.

(We should probably rename Object::send so it's not so easy to use directly.)

src/enumerator/arithmetic_sequence_object.cpp
richardboehme: There is a check above to verify that count is positive so this should be fine, shouldn't it?

src/numeric.rb
richardboehme: You mean __bind_method__? This defines the method on the object though which is not want I want here, right? I just want to call the method and use the return value.

src/enumerator/arithmetic_sequence_object.cpp
richardboehme: Interesting... I should take a look at how this works...

src/numeric.rb
seven1m: Sorry, yes, `__bind_method__`. Well you can use `__bind_method__` instead of `__function__`, since it makes a binding to the Ruby object. Then you don't have to use `__call__` here.

src/enumerator/arithmetic_sequence_object.cpp
seven1m: OK, very good. Let's just remove the comment then.

src/numeric.rb
seven1m: :thinking: Though... the ergonomics of getting the arguments out are little bit worse with `__bind_method__`, so I'm not recommending you change anything here. I just wanted you to be aware of it. :hugs:

seven1m (approved):
:tada:

PR #620, opened 2022-09-20
Make a start on Time objects

timcraft:
It's Time! Plenty still to do in order to make this fully featured and spec compliant, but here's what I have working so far in case anyone else wants to make use of it and/or work on it. A lot of the specs are dependent on timezones which aren't implemented yet, so I've added some basic sanity tests in `test/natalie/time_test.rb`


include/natalie/time_object.hpp
```diff
@@ -0,0 +1,72 @@
...
+    Value m_min;
+    Mode m_mode;
+    Value m_month;
+    Value m_sec;
+    Value m_string;
+    Value m_subsec;
+    struct tm m_time;
+    Value m_wday;
+    Value m_yday;
+    Value m_year;
```
seven1m, line 69: Do we really need all these members? I had assumed Ruby stored a Time as some offset since unix epoch...?

timcraft: Most of them are were an optimization for caching the return values of accessors, which is admittedly premature. I've removed those in 75207d8

The offset from epoch is held in `m_integer` but that isn't sufficient by itself—it doesn't encode timezone information or fractional subseconds for example. The `m_time` member holds a `struct tm` value, which has year, month etc and is what gets passed to strftime for formatting. The optional `m_subsec` member stores the fractional subseconds as a `RationalObject` value, similar to `time_object->vtm->subsecx` in MRI. The `m_mode` member is for differentiating between UTC and Localtime (and likely fixed offset times, which aren't implemented yet), similar to `time_object->tzmode` in MRI.

seven1m: Nice, thanks!

PR #622, opened 2022-09-23
Implement Range#step and Range#%

richardboehme:
This also fixes a bug in Range construction with the (x..y)-syntax. We were not checking if constructing the Range should be possible.

[#555]

### DIFF
--- a/src/range_object.cpp
+++ b/src/range_object.cpp
@@ -336,7 +336,57 @@ Value RangeObject::bsearch(Env *env, Block *block) {
     }
 }
 
-Value RangeObject::step(Env *env, Value step) {
-    return Enumerator::ArithmeticSequenceObject::from_range(env, m_begin, m_end, step, m_exclude_end);
+Value RangeObject::step(Env *env, Value n, Block *block) {
+    if (!n)
+        n = NilObject::the();
+
+    if (!n->is_numeric() && !n->is_nil())
+        n = n->to_int(env);
+
+    if (n.send(env, "=="_s, { Value::integer(0) })->is_true())
+        env->raise("ArgumentError", "step can't be 0");
+
+    if (m_begin->is_numeric() || m_end->is_numeric()) {
+        auto begin = m_begin;
+        auto end = m_end;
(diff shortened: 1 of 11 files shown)


src/range_object.cpp
```diff
@@ -3,11 +3,22 @@
...
 namespace Natalie {
 
-Value RangeObject::initialize(Env *env, Value begin, Value end, Value exclude_end_value) {
-    assert_not_frozen(env);
+RangeObject *RangeObject::create(Env *env, Value begin, Value end, bool exclude_end) {
+    assert_no_bad_value(env, begin, end);
+    auto range = new RangeObject(begin, end, exclude_end);
+    range->freeze();
+    return range;
+}
```
seven1m, line 11: I'm curious why we need a static function to create a RangeObject. Presumably RangeObject::initialize() would catch the invalid parameters, yes?

seven1m (approved):

src/range_object.cpp
richardboehme: Well we do not call `ìnitialize` when instantiating a Range from it's literal form (for example (1..2)). I thought of several solutions:

* Passing the env to the RangeObject constructor and call initialize in there. I do not really like passing the env into the constructor but it would work. I'm also not sure if throwing an exception in a constructor is fine but I guess it would be.
* Calling initialize after constructing the RangeObject at the caller site. This would lead to an allocation of the RangeObject even if it would be "invalid". Also we would probably forget about this when using RangeObject inside of other C++ code.
* Having a `create` function that takes the env, constructs the Range and calls initialize. This might also be a clean solution but we would also allocate the range before doing the bad value check.
* The current solution.

Which way would you prefer? Do you have other ideas?

seven1m: I like that you thought of all those alternatives. 🥰

I think the way you did it is fine, and your reasoning makes sense to me.

We may want to change the compiler so it emits code that creates objects with the equivalent of Klass.new in Ruby vs C++ `new Klass`. 🤔 

Let's ship this as is and give it all some thought down the road.

seven1m (approved):

PR #625, opened 2022-09-24
Add possibility to profile Natalie's compiler using StackProf

richardboehme:
This patch adds two new options to bin/natalie:

`--transform-only` - Only transform the given ruby code to cpp code without writing it to a file. This is useful for measuring the time needed by the compiler to transform ruby code to cpp.

`-pc` - Profiles the compiler using the stackprof gem. When speedscope is installed, the profile will be opened using speedscope!

Let me know what you think. We could definitely add more here. I'm thinking profiling the different passes for example.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,10 @@ gem 'minitest-reporters'
 gem 'minitest-parallel_fork'
 gem 'rake'
 
+group :development do
+  gem 'stackprof'
+end
+
 group :run_all_specs, optional: true do
   gem 'concurrent-ruby'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,6 +13,7 @@ GEM
       ruby-progressbar
     rake (13.0.6)
     ruby-progressbar (1.11.0)
+    stackprof (0.2.21)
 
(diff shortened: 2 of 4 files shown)


seven1m:
This is awesome!

What do you think about adding these new options under the `-d` debug flag. Seems more tidy, yeah?

richardboehme:
> What do you think about adding these new options under the -d debug flag. Seems more tidy, yeah?

Hm... I feel like the `transform-only` option would be a perfect fit for the `debug` flag. However, I don't feel like the profiling would fit there? It would make it hard to add more customizability to the profiler (for example only profile Pass1). I initially thought about integrating it into the `-p` profiling flag so we would do something like `bin/natalie -p compiler spec/core/numeric/step_spec.rb`. What do you think about this?

seven1m:
>I initially thought about integrating it into the -p profiling flag

Ahh, yes, that sounds even better. Let's do that!

richardboehme:
Ok I changed the options. What do you think?

seven1m (approved):

seven1m:
Great!

PR #626, opened 2022-09-24
Memoize label of a compiler instruction

richardboehme:
This reduces transformation time by ~50%

Before:
image

After:
image

With label as class method:

image


richardboehme:
Now that I've took another look at the profile. Couldn't `BaseInstruction#label` be a class method? This would speed this up a bit more...

seven1m:
Yeah probably could be a class method. Try it and see!

richardboehme:
> Yeah probably could be a class method. Try it and see!

This got us another ~100ms when compiling the numeric step spec!

seven1m:
Wow it's amazing what one little change can do!

PR #629, opened 2022-10-02
Compile Rational lit

richardboehme:
This patch adds support for compiling a Rational lit like `2/3r`.

This also improves the error message for unknown lits and adds the official ruby debug gem to ruby/debug (let me know if we should change that).


lib/natalie/compiler/instructions/push_rational_instruction.rb
```diff
@@ -0,0 +1,23 @@
...
+  class Compiler
+    class PushRationalInstruction < BaseInstruction
+      def to_s
+        'push_rational'
+      end
+
+      def generate(transform)
+        denominator = transform.pop
+        numerator = transform.pop
+        transform.exec_and_push(:rational, "Value(RationalObject::create(env, #{denominator}->to_int(env), #{numerator}->to_int(env)))")
```
richardboehme, line 13: Not sure about the `to_int` here, but RationalObject only takes IntegerObject pointers so we kind of have to do that. However, I'm not sure if RationalObject should hold Values instead of IntegerObjects (for example subclasses of Integer?)

seven1m: This is fine for now. We can change it later if needed!

bin/natalie
```diff
@@ -1,5 +1,14 @@
...
 
+if RUBY_ENGINE != 'natalie'
+  require 'bundler/inline'
+
+  gemfile do
+    source 'https://rubygems.org'
+    gem 'debug', '1.6.2'
+  end
+end
+
```
seven1m, line 11: I think this should go in `Gemfile` with the other development dependencies?

seven1m: Actually, I just realized we removed the need for Bundler when we removed RubyParser... I'm hesitant to add it back in. Do you think this could be made optional somehow?

richardboehme: The Gemfile is not loaded when running bin/natalie... However the gem seems to make problems on CI anyway 😢

richardboehme: > Actually, I just realized we removed the need for Bundler when we removed RubyParser... I'm hesitant to add it back in. Do you think this could be made optional somehow?

We could hide it behind a DEBUG environment variable or something?

seven1m (approved):

PR #634, opened 2022-10-05
Add Float Infinity expectation & unskip related tests

stevegeek:
Hi all! Love what you are building here! Wanted to help get more specs green.

I noticed a couple of skipped / failing specs on `divide_spec` for Float. Fixed by adding the missing matchers (inspired by https://github.com/ruby/mspec/blob/master/lib/mspec/matchers/infinity.rb) and updated the spec to the latest from `ruby/spec` (https://github.com/ruby/spec/blob/master/core/float/divide_spec.rb)

### DIFF
--- a/spec/core/float/divide_spec.rb
+++ b/spec/core/float/divide_spec.rb
@@ -8,19 +8,19 @@ describe "Float#/" do
   it "returns self divided by other" do
     (5.75 / -2).should be_close(-2.875,TOLERANCE)
     (451.0 / 9.3).should be_close(48.494623655914,TOLERANCE)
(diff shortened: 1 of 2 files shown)


seven1m:
Thanks for your contribution! <3

The failing spec seems to be unrelated to your change. I believe the GC is collecting an object it shouldn't be. I will investigate!

seven1m:
Fixed in a8249557a1f656308404afcad0d7c5656548f165!

stevegeek:
Yay! Thanks @seven1m !

schneems:
Great catch, quick followup. Congrats on the merge!

PR #644, opened 2022-10-12
Implement Float#to_r

timcraft:
There isn't any spec coverage provided by `spec/core/float/to_r_spec.rb` so I've added some examples from Ruby's Float#to_r documentation to `test/natalie/float_test.rb` for now

I had to change how `KernelModule::Rational` constructed the float object to get it to work—without any change that was resulting in the "This object is stack allocated..." failure from `NAT_GC_GUARD_VALUE`

### DIFF
--- a/src/kernel_module.cpp
+++ b/src/kernel_module.cpp
@@ -489,8 +489,8 @@ RationalObject *KernelModule::Rational(Env *env, double arg) {
     Value y = radix.pow(env, Value::integer(DBL_MANT_DIG));
 
     int exponent;
-    FloatObject significand { std::frexp(arg, &exponent) };
-    Value x = significand.mul(env, y)->as_float()->to_i(env);
+    FloatObject *significand = new FloatObject { std::frexp(arg, &exponent) };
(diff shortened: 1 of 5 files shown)


src/kernel_module.cpp
```diff
@@ -489,8 +489,8 @@ RationalObject *KernelModule::Rational(Env *env, double arg) {
     Value y = radix.pow(env, Value::integer(DBL_MANT_DIG));
 
     int exponent;
-    FloatObject significand { std::frexp(arg, &exponent) };
-    Value x = significand.mul(env, y)->as_float()->to_i(env);
+    FloatObject *significand = new FloatObject { std::frexp(arg, &exponent) };
+    Value x = significand->mul(env, y)->as_float()->to_i(env);
```
seven1m, line 493: This is good. We should either be doing this (`new FloatObject`) or using `Value::floatingpoint()` to avoid capturing a stack-allocated object.

timcraft: Got it, thanks for the explanation 👍

PR #647, opened 2022-10-13
Implement `numerator` & `denominator` on Float & Numeric

stevegeek:
Implement `numerator` & `denominator` on Float & Numeric

Also simplify `is_finite` to just use `is_infinity`

Adds specs for the Numeric `denominator` and `numerator` methods (I should upstream these specs to `ruby/spec`)


spec/core/numeric/denominator_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  it "calls self#to_r then returns the #denominator" do
+    result = mock("Numeric#to_r result")
+    result.should_receive(:denominator).and_return(12)
+    @obj.should_receive(:to_r).and_return(result)
+
+    @obj.denominator.should == 12
+  end
+end
```
seven1m, line 16: If you do indeed plan to PR these into ruby/spec upstream, then this is fine. However, our usual technique is to add our tests in the `test/natalie` directory, e.g. `test/natalie/float_test.rb` or similar.

seven1m (approved):
Let me know if you want to make changes before I merge. Looks good to me!

spec/core/numeric/denominator_spec.rb
stevegeek: I was planning to, but then I just noticed they already exist 🤦 

https://github.com/ruby/spec/blob/master/core/numeric/denominator_spec.rb

I will update PR to add them

seven1m: Ahh yeah, we just copy over files as-needed. Thanks!

PR #648, opened 2022-10-14
Implement String#slice and String#slice!

kddnewton:
👋 Hey there!

Saw your tweet about hacktoberfest and decided to attempt a contribution. I saw that `slice_spec` wasn't implemented yet, so attempted that here.

I learned a lot about the way you architected natalie and like it a lot! There are a couple of things that I'm not sure of in here - happy to change anything if/when you get a chance to review. I'll post as many comments on GitHub inline in the code to try to explain my thought process.


include/natalie/match_data_object.hpp
```diff
@@ -33,10 +33,11 @@ class MatchDataObject : public Object {
 
     size_t size() { return m_region->num_regs; }
 
-    size_t index(size_t);
+    long index(size_t);
```
kddnewton, line 36: This is important because the index can be `-1` if the capture group doesn't have a value, and it was underflowing with `size_t` previously.

include/natalie/match_data_object.hpp
```diff
@@ -33,10 +33,11 @@ class MatchDataObject : public Object {
...
+    long ending(size_t);
 
     Value array(int);
-    Value group(Env *, size_t);
+    Value group(int);
```
kddnewton, line 40: Env isn't actually needed here. Happy to add it back if it's a convention though?

include/natalie/string_object.hpp
```diff
@@ -223,7 +224,9 @@ class StringObject : public Object {
     Value reverse_in_place(Env *);
     Value rstrip(Env *) const;
     Value rstrip_in_place(Env *);
+    size_t char_count(Env *) const;
```
kddnewton, line 227: Any chance this is cached somewhere? There are a lot of string methods that play around with indices, and you'll need to know how to handle negative indices by wrapping the character count. I wasn't sure if I just missed it somewhere.

spec/core/string/shared/slice.rb
```diff
@@ -80,19 +80,21 @@
...
-    c.send(@method, 1, 3).encoding.should == Encoding::US_ASCII
-    c.send(@method, 8, 2).encoding.should == Encoding::US_ASCII
-    c.send(@method, 1, 10).encoding.should == Encoding::US_ASCII
+  quarantine! do # NATFIXME: more encodings!
+    it "returns a string with the same encoding" do
+      s = "hello there"
+      s.send(@method, 1, 9).encoding.should == s.encoding
+
+      a = "hello".force_encoding("binary")
+      b = " there".force_encoding("ISO-8859-1")
```
kddnewton, line 89: Fails on this line because of the other encoding.

src/match_data_object.cpp
```diff
@@ -6,24 +6,36 @@ Value MatchDataObject::array(int start) {
...
     auto array = new ArrayObject { size };
     for (int i = start; i < m_region->num_regs; i++) {
-        if (m_region->beg[i] == -1) {
-            array->push(NilObject::the());
-        } else {
-            const char *str = &m_string->c_str()[m_region->beg[i]];
-            size_t length = m_region->end[i] - m_region->beg[i];
-            array->push(new StringObject { str, length });
-        }
+        array->push(group(i));
```
kddnewton, line 9: This was basically just duplicating what `group` already did, so I consolidated it here.

src/match_data_object.cpp
```diff
@@ -6,24 +6,36 @@ Value MatchDataObject::array(int start) {
...
+    return m_region->end[index];
+}
+
+/**
+ * Return the capture indicated by the given index. This function supports
+ * negative indices as well, provided they are within one overall length of the
+ * number of captures.
+ */
+Value MatchDataObject::group(int index) {
+    if (index + m_region->num_regs <= 0 || index >= m_region->num_regs)
```
kddnewton, line 30: Moved the range checking into this function (as opposed to ref) since I needed to reference it in the slice! implementation.

spec/core/string/slice_spec.rb
```diff
@@ -0,0 +1,453 @@
...
+
+describe "String#slice with Regexp" do
+  it_behaves_like :string_slice_regexp, :slice
+end
+
+describe "String#slice with Regexp, index" do
+  it_behaves_like :string_slice_regexp_index, :slice
+end
+
+quarantine! do
```
kddnewton, line 27: These aren't implemented yet. Wasn't sure if you prefer this or some other method of marking specs as pending.

spec/core/string/slice_spec.rb
```diff
@@ -0,0 +1,453 @@
...
+    -> { "hello".freeze.slice!(1, 2)  }.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(10, 3) }.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(-10, 3)}.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(4, -3) }.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(10, 3) }.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(-10, 3)}.should raise_error(FrozenError)
+    -> { "hello".freeze.slice!(4, -3) }.should raise_error(FrozenError)
+  end
+
+  quarantine! do
```
kddnewton, line 124: Honestly not sure why this one is failing. Something to do with method_missing not being called in the implicit conversion probably?

spec/core/string/slice_spec.rb
```diff
@@ -0,0 +1,453 @@
...
+  end
+
+  it "returns the substring given by the character offsets" do
+    "hellö there".slice!(1,0).should == ""
+    "hellö there".slice!(1,3).should == "ell"
+    "hellö there".slice!(1,6).should == "ellö t"
+    "hellö there".slice!(1,9).should == "ellö ther"
+  end
+
+  quarantine! do
```
kddnewton, line 162: Yeah I really don't know why this one is failing at all.

spec/core/string/slice_spec.rb
```diff
@@ -0,0 +1,453 @@
...
+
+  ruby_version_is '3.0' do
+    it "returns String instances" do
+      s = StringSpecs::MyString.new("hello")
+      s.slice!(0...0).should be_an_instance_of(String)
+      s.slice!(0..4).should be_an_instance_of(String)
+    end
+  end
+
+  quarantine! do
```
kddnewton, line 210: Same thing here with the method_missing I think.

include/natalie/match_data_object.hpp
seven1m: We usually use `ssize_t` if it can be negative.

include/natalie/string_object.hpp
seven1m: No, not cached, but I will take note to do that later!

include/natalie/match_data_object.hpp
kddnewton: Updated!

FnControlOption:
Nice work 👍 (Btw you can use `xdescribe` and `xit` instead of `quarantine!`)

PR #651, opened 2022-10-15
Include required keywords in error message if wrong number of args

FnControlOption:


seven1m (approved):
This is great. I'd love a test or two to go with the code so we don't regress.

PR #658, opened 2022-10-20
implement String#sum

AI-Mozi:
[#217]

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -1959,4 +1959,29 @@ Value StringObject::chop_in_place(Env *env) {
     return this;
 }
 
+Value StringObject::sum(Env *env, Value val) {
+  int base = 16;
+  int sum = 0;
+  
+  if (val) {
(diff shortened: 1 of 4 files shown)


src/string_object.cpp
```diff
@@ -1959,4 +1959,29 @@ Value StringObject::chop_in_place(Env *env) {
...
+Value StringObject::sum(Env *env, Value val) {
+  int base = 16;
+  int sum = 0;
+  
+  if (val) {
+      if (!val->is_integer() && val->respond_to(env, "to_int"_s)){
+        val = val->send(env, "to_int"_s);
+      }
+      
+      if (val->is_integer()) {
```
seven1m, line 1971: `assert_type()` validates that the type is an integer, which is what we want. You don't need the `if` around it.

seven1m (changes requested):

src/string_object.cpp
AI-Mozi: make sense 😅

seven1m (approved):
Great work

PR #662, opened 2022-10-21
Add Time#usec spec

timcraft:
Time#usec is already implemented, but a couple of changes required to get the spec passing:

* Float microseconds argument to `Time.at` e.g. `Time.at(0, 3.75)`
* Rational second argument to `Time.local` e.g. `Time.local(1,2,3,4,5,Rational(6.78))`
  (second argument is documented as integer, but the implementation accepts other numeric values)

I've added `TimeObject::convert_rational` to handle the repeated type conversion logic, and changed the `TimeObject::build_time_struct` helper function so that it can store the subseconds.

### DIFF
--- a/include/natalie/time_object.hpp
+++ b/include/natalie/time_object.hpp
@@ -62,6 +62,7 @@ public:
 
 private:
     static struct tm build_time_struct(Env *, Value, Value, Value, Value, Value, Value);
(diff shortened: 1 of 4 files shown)


seven1m (approved):
Nice clean up :+1:

PR #664, opened 2022-10-21
Feature/string#partition

AI-Mozi:
[#217]
It's a bit messy 😅

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -1959,4 +1959,59 @@ Value StringObject::chop_in_place(Env *env) {
     return this;
 }
 
+Value StringObject::partition(Env *env, Value val) {
+    auto ary = new ArrayObject;
+
+    if (val->is_regexp()) {
+        auto match_result = val->as_regexp()->match(env, this);
+    
+        ssize_t start_byte_index;
+        ssize_t end_byte_index;
+
+        if (match_result->is_nil()){
+            auto empty = new StringObject { "" };
+            return new ArrayObject { this, empty, empty };
+        } else {
+            start_byte_index = match_result->as_match_data()->index(0);
+            end_byte_index = match_result->as_match_data()->ending(0);
+            ary->push(new StringObject(m_string.substring(0, start_byte_index)));
+        }
+
+        ary->push(new StringObject(m_string.substring(start_byte_index, end_byte_index - start_byte_index)));
+
+        auto last_char = start_byte_index + (end_byte_index - start_byte_index);
+        auto len = m_string.length() - 1;
+        if (last_char < static_cast<long>(len)) {
+            ary->push(new StringObject((m_string.substring(end_byte_index, m_string.length() - end_byte_index))));
+            return ary;
+        }
+    } else {
+        if (!val->is_string() && val->respond_to(env, "to_str"_s)) {
+        val = val->send(env, "to_str"_s);
+        }
+        val->assert_type(env, Type::String, "String");
+
+        auto query = val->as_string()->string();
+        auto query_idx = m_string.find(query);
+
+        if (query_idx == -1) {
+            auto empty = new StringObject { "" };
+            return new ArrayObject { this, empty, empty };
+        } else {
+            ary->push(new StringObject(m_string.substring(0, query_idx)));
+        }
+        
+        ary->push(val);
+
+        auto last_char = query_idx + query.length();
+        auto len = m_string.length() - 1;
+        if (last_char < len) {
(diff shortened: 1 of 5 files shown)


src/string_object.cpp
```diff
@@ -1959,6 +1959,62 @@ Value StringObject::chop_in_place(Env *env) {
...
+
+        auto last_char = start_byte_index + (end_byte_index - start_byte_index);
+        auto len = m_string.length() - 1;
+        if (last_char < static_cast<long>(len)) {
+            ary->push(new StringObject((m_string.substring(end_byte_index, m_string.length() - end_byte_index))));
+            return ary;
+        }
+    } else {
+        if (!val->is_string() && val->respond_to(env, "to_str"_s)) {
+        val = val->send(env, "to_str"_s);
```
seven1m, line 1990: I wonder why clang-format didn't catch the indentation issue here... :thinking:

src/string_object.cpp
```diff
@@ -1959,6 +1959,62 @@ Value StringObject::chop_in_place(Env *env) {
...
+
+    if (val->is_regexp()) {
+        auto match_result = val->as_regexp()->match(env, this);
+    
+        ssize_t start_byte_index;
+        ssize_t end_byte_index;
+
+        if (match_result->is_nil()){
+            auto empty = new StringObject { "" };
+            return new ArrayObject { this, empty, empty };
```
seven1m, line 1973: Unfortunately we cannot use the same `empty` object twice:

```ruby
irb(main):002:0> ''.partition(/foo/).map(&:object_id)
=> [180, 200, 220]
```

seven1m: Actually, it looks like we cannot return `this` either! :cry: 

```ruby
irb(main):006:0> s.object_id
=> 240
irb(main):007:0> s.partition(/foo/).map(&:object_id)
=> [260, 280, 300]
```

src/string_object.cpp
```diff
@@ -1959,6 +1959,62 @@ Value StringObject::chop_in_place(Env *env) {
...
+        val = val->send(env, "to_str"_s);
+        }
+        val->assert_type(env, Type::String, "String");
+
+        auto query = val->as_string()->string();
+        auto query_idx = m_string.find(query);
+
+        if (query_idx == -1) {
+            auto empty = new StringObject { "" };
+            return new ArrayObject { this, empty, empty };
```
seven1m, line 1999: Same here as above.

src/string_object.cpp
```diff
@@ -1959,6 +1959,62 @@ Value StringObject::chop_in_place(Env *env) {
...
+        } else {
+            start_byte_index = match_result->as_match_data()->index(0);
+            end_byte_index = match_result->as_match_data()->ending(0);
+            ary->push(new StringObject(m_string.substring(0, start_byte_index)));
+        }
+
+        ary->push(new StringObject(m_string.substring(start_byte_index, end_byte_index - start_byte_index)));
+
+        auto last_char = start_byte_index + (end_byte_index - start_byte_index);
+        auto len = m_string.length() - 1;
```
seven1m, line 1983: I think the naming here is confusing. This isn't `len` (length) -- it's actually the last character index. I'd prefer we use more descriptive names that match what they are, even if it means the names are longer, e.g. `last_char_index` would be fine!

src/string_object.cpp
```diff
@@ -1959,6 +1959,62 @@ Value StringObject::chop_in_place(Env *env) {
...
+            auto empty = new StringObject { "" };
+            return new ArrayObject { this, empty, empty };
+        } else {
+            ary->push(new StringObject(m_string.substring(0, query_idx)));
+        }
+        
+        ary->push(val);
+
+        auto last_char = query_idx + query.length();
+        auto len = m_string.length() - 1;
```
seven1m, line 2007: Same confusing `len` here as above.

seven1m (changes requested):
Very cool! Just a few small fixes, since the specs didn't help much on this one. :grimacing:

src/string_object.cpp
AI-Mozi: Hmm.. so what is the good solution?

seven1m: You'll just have to return an array with three new string objects, unfortunately:

```c++
return new ArrayObject { new StringObject, new StringObject, new StringObject };
```

seven1m:
I guess getting clang-format workflow to work against a fork is kinda hard. :man_shrugging:

seven1m (approved):

seven1m:
clang-format ran on master here: 21df1aaabd21fb01677a3bde711b9411e4fdde05

PR #673, opened 2022-10-23
Add String#getbyte and #setbyte methods

andrykonchin:


src/string_object.cpp
```diff
@@ -579,6 +579,29 @@ size_t StringObject::char_count(Env *env) const {
     return char_count;
 }
 
+Value StringObject::setbyte(Env *env, Value index_obj, Value value_obj) {
+    assert_not_frozen(env);
+
+    int index = IntegerObject::convert_to_int(env, index_obj);
```
seven1m, line 585: Technically `index` could be a larger index, yeah? I think this should use `IntegerObject::convert_to_nat_int_t()`. It may not matter practically, since an `int` could index into a string of 2 Gigs, but it would be nice to match everywhere else we index into strings (which uses either a `size_t` or a `nat_int_t`).

src/string_object.cpp
```diff
@@ -1304,6 +1327,25 @@ Value StringObject::gsub(Env *env, Value find, Value replacement_value, Block *b
     }
 }
 
+Value StringObject::getbyte(Env *env, Value index_obj) const {
+    int index = IntegerObject::convert_to_int(env, index_obj);
```
seven1m, line 1331: Same thing about int vs nat_int_t here as I mentioned above.

src/string_object.cpp
andrykonchin: Fixed.

I would like to squash commits into a single one before merging. Please let me know if there is anything to change.

src/string_object.cpp
```diff
@@ -579,6 +579,29 @@ size_t StringObject::char_count(Env *env) const {
...
+Value StringObject::setbyte(Env *env, Value index_obj, Value value_obj) {
+    assert_not_frozen(env);
+
+    nat_int_t index = IntegerObject::convert_to_nat_int_t(env, index_obj);
+    nat_int_t value = IntegerObject::convert_to_nat_int_t(env, value_obj);
+    nat_int_t length = static_cast<nat_int_t>(m_string.length());
+
+    if (index + length < 0) {
+        env->raise("IndexError", "index {} out of string", index);
+    }
```
seven1m, line 591: I don't understand this check... Is this supposed to check for a negative `index`? If so, why add `length` to it?

src/string_object.cpp
```diff
@@ -579,6 +579,29 @@ size_t StringObject::char_count(Env *env) const {
...
+
+    if (index >= length) {
+        env->raise("IndexError", "index {} out of string", index);
+    }
+
+    if (index < 0) {
+        index += length;
+    }
```
seven1m, line 599: Ahh, here is where we check for a negative index. But let's say the `index` is `-100` and the `length` is only `10`... the resulting `index` will still be negative. :grimacing: 

I think you need to move this code up higher, then check the result is positive.

andrykonchin: Yeah, negative index is a bit tricky.

Index is allowed to have the following values:
```
-length <= index < length
```

Otherwise `IndexError` should be raised.

This incrementing `index += length;` may be placed at the beginning but we still need an original value to form the proper exception message. So just try to avoid adding new local variable with original value here.

Some examples:

```ruby
s = "abc"; s.setbyte(-1, "d".ord); s
# => "abd"
s = "abc"; s.setbyte(-2, "d".ord); s
# => "adc"
s = "abc"; s.setbyte(-3, "d".ord); s
# => "dbc"
s = "abc"; s.setbyte(-4, "d".ord); s
# (irb):11:in `setbyte': index -4 out of string (IndexError)
```

seven1m: OK, I think I get it! This is fine then.

I tend to lean toward readability, so I think I would have created a temporary variable to hold the original value. Or, I would have used `index_obj` for the error message.

But in any case, this works and now I know how to read `if (index + length < 0)` which was a bit too clever for my brain. :laughing: 

:shipit:

seven1m (approved):
Thanks for adding these!

src/string_object.cpp
andrykonchin: 👍

andrykonchin:
I've refactored a bit and squashed working commits.

PR #679, opened 2022-10-25
Fix Float#<=> when other responds to `infinite?`

stevegeek:
Fixes `cmp` to support comparing infinity to rhs which responds to `infinite?`, and updates the spec to current upstream comparison spec.

**A quick point for discussion:**

here I use `->as_integer()` on the Value returned from `infinite?` (which should return `1`, `nil`, or `-1`) after checking for `nil`... however it is possible that an implementation of that method returns a non-integer value. If this is the case an assertion error will occur here.

If we look at Ruby, it handles this by calling `rb_cmpint` which effectively converts the value into an integer of the set of valid values `[-1, 0, 1]`. It does this for non-integer values by calling `< 0` and `> 0` on the object -  https://github.com/ruby/ruby/blob/5ccb625fbbd1e774636a9fdbe0bf1c3d38e085d5/bignum.c#L2969

I could implement that same behaviour here in-line, eg I see this is done in-line here for example: https://github.com/natalie-lang/natalie/blob/99df60fd6a25d9b77dcdbb4b08f0ccfccd746afe/src/time_object.cpp#L97

But I guess a function such as `rb_cmpint` would be better, to be used generally where return values for use with `cmp` need converting (eg in Time as linked above).

However I'm not sure where that function would live in natalie, any thoughts welcome!

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -1166,26 +1166,4 @@ StringObject *Object::to_str(Env *env) {
         result->klass()->inspect_str());
 }
 
-int Object::to_cmp_int(Env *env, Value lhs, Value rhs) {
-    if (is_nil()) {
-        env->raise_not_comparable_error(lhs, rhs);
-    }
-
-    if (is_integer()) {
-        auto i = as_integer();
-        if (i->send(env, ">"_s, { Value::integer(0) })->is_truthy())
-            return 1;
-        if (i->send(env, "<"_s, { Value::integer(0) })->is_truthy())
-            return -1;
-        return 0;
-    }
-
-    if (send(env, ">"_s, { Value::integer(0) })->is_truthy())
-        return 1;
-    else if (send(env, "<"_s, { Value::integer(0) })->is_truthy())
-        return -1;
-
-    return 0;
-}
-
 }
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -1166,4 +1166,26 @@ StringObject *Object::to_str(Env *env) {
         result->klass()->inspect_str());
 }
 
+int Object::to_cmp_int(Env *env, Value lhs, Value rhs) {
+    if (is_nil()) {
+        env->raise_not_comparable_error(lhs, rhs);
+    }
+
+    if (is_integer()) {
+        auto i = as_integer();
+        if (i->send(env, ">"_s, { Value::integer(0) })->is_truthy())
+            return 1;
+        if (i->send(env, "<"_s, { Value::integer(0) })->is_truthy())
+            return -1;
+        return 0;
+    }
+
+    if (send(env, ">"_s, { Value::integer(0) })->is_truthy())
+        return 1;
+    else if (send(env, "<"_s, { Value::integer(0) })->is_truthy())
+        return -1;
+
+    return 0;
+}
+
 }
--- a/include/natalie/env.hpp
+++ b/include/natalie/env.hpp
@@ -54,6 +54,7 @@ public:
     [[noreturn]] void raise_errno();
     [[noreturn]] void raise_no_method_error(Object *, SymbolObject *, MethodMissingReason);
     [[noreturn]] void raise_name_error(SymbolObject *name, String);
+    [[noreturn]] void raise_not_comparable_error(Value lhs, Value rhs);
 
     template <typename... Args>
     [[noreturn]] void raise_name_error(SymbolObject *name, const char *format, Args... args) {
--- a/include/natalie/object.hpp
+++ b/include/natalie/object.hpp
@@ -341,6 +341,10 @@ public:
     StringObject *to_s(Env *env);
     StringObject *to_str(Env *env);
 
+    // If the object value will be returned from `<=>` and used in a comparison,
+    // then this method converts it into an integer in the set {-1, 0, 1}.
+    // If the conversion is not possible, then an ArgumentError is raised.
+    int to_cmp_int(Env *env, Value, Value);
 protected:
     ClassObject *m_klass { nullptr };
 
--- a/include/natalie/object.hpp
+++ b/include/natalie/object.hpp
@@ -341,10 +341,6 @@ public:
     StringObject *to_s(Env *env);
     StringObject *to_str(Env *env);
 
-    // If the object value will be returned from `<=>` and used in a comparison,
-    // then this method converts it into an integer in the set {-1, 0, 1}.
-    // If the conversion is not possible, then an ArgumentError is raised.
-    int to_cmp_int(Env *env, Value, Value);
 protected:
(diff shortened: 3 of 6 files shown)


stevegeek:
For example, these should pass:

```ruby
  it "returns 1 when self is Infinity and other infinite? returns value that responds is neither > or < than 0" do
    obj = Object.new
    def obj.infinite?
      val = Object.new
      def val.<(other)
        false
      end
      def val.>(other)
        false
      end
      val
    end
    (infinity_value <=> obj).should == 1
  end

  it "returns -1 when self is -Infinity and other infinite?=1 via value that responds to > and is greater than 0" do
    obj = Object.new
    def obj.infinite?
      val = Object.new
      def val.>(other)
        true
      end
      val
    end
    (-infinity_value <=> obj).should == -1
  end
```

seven1m:
>But I guess a function such as rb_cmpint would be better, to be used generally where return values for use with cmp need converting (eg in Time as linked above).
>However I'm not sure where that function would live in natalie, any thoughts welcome!

We have the junk drawer that is `natalie.hpp/cpp` -- you can stick a function in there and that would be fine. But perhaps it would make sense to put it on the `Object` class, kind of like we do for `to_s` and `to_str` etc.

stevegeek:
@seven1m Do the latest changes look like they have gone in the right direction? 

If so I could then refactor:
1) places where `to_cmp_int` is needed and currently implemented inline (eg in Time)
2) places where comparison errors are raised to instead use the method I added on `Env` (if its appropriate to add that method...)

src/object.cpp
```diff
@@ -1166,4 +1166,26 @@ StringObject *Object::to_str(Env *env) {
         result->klass()->inspect_str());
 }
 
+int Object::to_cmp_int(Env *env, Value lhs, Value rhs) {
+    if (is_nil()) {
+        env->raise_not_comparable_error(lhs, rhs);
+    }
+
+    if (is_integer()) {
```
stevegeek, line 1174: This conditional is unnecessary

seven1m: Actually, this method is confusing. It accepts `lhs` and `rhs` but doesn't use them??

stevegeek: Yeah it looks odd! It accepts them as in Ruby the error message generated by values which can't be compared contains their class names. So lhs and rhs are forwarded to the error raise method. It's basically the same implementation as the one in Ruby. I'm not sure how strictly natalie follows the message formatting for errors of Ruby but I see these "not comparable" exceptions dotted around the code base and thought we could make a helper to generate them (as in Ruby)

seven1m: Hmm, I see. Still a bit of an odd method. It's just not intuitive how to use it.

I wonder if it would make more sense to remove `lhs` and use `this`.

Or, alternatively, move it off of Object, and make a regular function that takes `lhs` and `rhs`.

stevegeek: Ok thanks for the suggestion @seven1m. It prob makes sense for the method to simply live as a standalone function.

If on Object, its usage would be odd if `lhs` was `this` as the value we are converting is not the `lhs` of the comparison but rather the return value of the thing used in the comparison, eg in the case of `Infinity`s, the `this` here will be the return value of `infinite?` that we are converting (while the `lhs` will be the Infinite itself) or in sort the return value of the block (ie in`.sort { |a, b| c }`,  `c` is the one being converted, while `a` and `b` are the `lhs` and `rhs`)

One option could be to see if that `nil` guard be removed thus negating the need for `lhs` and `rhs`... [Ruby has it](https://github.com/ruby/ruby/blob/5ccb625fbbd1e774636a9fdbe0bf1c3d38e085d5/bignum.c#L2955) but maybe it could be avoided for `nil` guards at the call sites. 

I'll have a look at its usages in Ruby and see what I can do!

Btw for context, here is a case where that `nil` guard comes into play:

```ruby
irb(main):002:0> [1, 2, 3, 4].sort { |a, b| nil}
	20: from /Users/stephen/.asdf/installs/ruby/2.7.2/bin/irb:23:in `<main>'
	19: from /Users/stephen/.asdf/installs/ruby/2.7.2/bin/irb:23:in `load'
	18: from /Users/stephen/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.4.2/exe/irb:11:in `<top (required)>'
	 1: from (irb):2:in `<main>'
(irb):2:in `sort': comparison of Integer with 2 failed (ArgumentError)
```

The return value from `sort` should be converted into the range [-1, 0 ,1]. Here the value `nil` is being returned from the block, so when passed to `rb_cmpint` (https://github.com/ruby/ruby/blob/5ccb625fbbd1e774636a9fdbe0bf1c3d38e085d5/array.c#L3492) causes the guard to be triggered and raises our error (the `lhs` and `rhs` in this case is just 2 of the integers of the array being compared).

In `natalie` the return value conversion to the necessary range is duplicated and handled inline in various places, eg in `sort` https://github.com/natalie-lang/natalie/blob/f85770c630afb19102b9fd1e595c6381b9702e21/src/array_object.cpp#L1028 (also as the TODO there indicates, the error isnt created how Ruby does it, as it does some logic on the `lhs` before raising as I tried to replicated in `Env::raise_not_comparable_error`)

Btw interestingly for `sort`, the result is different to Ruby here:

```
irb(main):011:0> val = Object.new; def val.<(other); false; end; def val.>(other); false; end; [1, 2, 3, 4].sort { |a, b| val }
=> [1, 2, 3, 4]
```

compared to:
```
TLW:natalie stephen$ ./bin/natalie -e " val = Object.new; def val.<(other); false; end; def val.>(other); false; end; puts [1, 2, 3, 4].sort { |a, b| val }"
[4, 1, 2, 3]
```

Ill see if I can find the root cause of this too!

Sorry for the long discussion!

seven1m:
@stevegeek I had a minute to look at this today and decided to get a better handle on it myself. I made a commit to do pretty much the minimum to get the specs passing, and it looks good to me. I don't see the need for the `to_cmp_int` method right now, so I removed it. If you can think of a test that would exercise it, we can add it back in another PR!

Thanks so much for your work on this, I really appreciate it.

stevegeek:
Ok sure up to you! The point of `to_cmp_int` it to ensure a common and Ruby compatable way of handling the return values that are passed to `<=>`, in all cases where such a thing would happen, eg in Time comparison, Array sort etc, however understand your position! 

Ill see if I can get a set of specs added upstream to `ruby/spec` that make the need for such a method clearer and then if necessary come back to it!

seven1m:
Yeah I think I see why we need the method, but we should have specs/tests driving that. Let's do it in another PR. I'm ok with us adding tests for it in the `test/` directory.

PR #683, opened 2022-10-25
Fix enumerator size for String#each_char and String#each_byte

andrykonchin:
Return string size/byte size (not nil) as an enumerator size when `#each_char`/`#each_byte` are called without a block.

Example:

```ruby
enum = "abc".each_char
enum.size # => 3
```

### DIFF
--- a/spec/core/string/each_byte_spec.rb
+++ b/spec/core/string/each_byte_spec.rb
@@ -48,8 +48,7 @@ describe "String#each_byte" do
 
     describe "returned enumerator" do
       describe "size" do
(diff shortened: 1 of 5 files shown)


spec/core/string/each_byte_spec.rb
```diff
@@ -49,7 +49,7 @@
     describe "returned enumerator" do
       describe "size" do
         # NATFIXME: Make Enumerator#size spec compliant
```
seven1m, line 51: This comment can be deleted now, yes?

seven1m (approved):
Looks good. I'll merge once the checks pass.

PR #689, opened 2022-10-27
Add Array#pack directives 'V' and 'v'

Crystalsage:
This PR adds the Array#pack directives 'V' and 'v'. Related to issue #195.

The pack directives 'v' and 'V' pack an array of integers into 16-bit unsigned and 32-bit unsigned little endian byte order respectively. I tested my code with different inputs in the Natalie REPL and compared the outputs with the `irb`. It seems that everything is working fine.

Before merging this, I would like to confess that a spec gets skipped when I try to run the `v_spec.rb` file. 

image

## Debugging
I did some debugging in `test/support/spec.rb` and placed some inspections. It seems like the spec `encodes a Float truncated as an Integer` gets skipped because the `fn` for it is `nil`. Does this mean that this particular spec hasn't been implemented for 'v' and 'V' directives? 

This is my first time ever working on compilers and would like some inputs. I'm fairly confused on this one.

### DIFF
--- a/include/natalie/array_packer/integer_handler.hpp
+++ b/include/natalie/array_packer/integer_handler.hpp
@@ -57,6 +57,12 @@ namespace ArrayPacker {
             case 's':
                 pack_s();
                 break;
+            case 'v':
+                pack_v();
+                break;
+            case 'V':
+                pack_V();
+                break;
             default: {
                 char buf[2] = { d, '\0' };
                 env->raise("ArgumentError", "unknown directive in string: {}", buf);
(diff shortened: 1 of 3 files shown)


seven1m (approved):
Nice!

The skipped spec is [here](https://github.com/natalie-lang/natalie/blob/master/spec/core/array/pack/shared/integer.rb#L124-L125), and it seems someone found a bug in our float truncation. We can look into later... not a big deal.

Crystalsage:
Oh I see. Thanks for the review!

PR #690, opened 2022-10-27
Use Object::to_str(Env *) for implicit conversion to String

andrykonchin:
Changes:
- fixed an error message when implicit conversion to String fails because `#to_str` doesn't return String
- fixed several other (not related) error messages (unfortunately they aren't covered with specs)

A common pattern to convert an argument to String:

```ruby
    auto to_str = "to_str"_s;
    if (!arg->is_string() && arg->respond_to(env, to_str))
        arg = arg->send(env, to_str);
    arg->assert_type(env, Object::Type::String, "String");
```

leads to an error message

```
(repl):3:in `include?': no implicit conversion of Integer into String (TypeError)
```

But expected message usually is the following:

```
(irb):59:in `include?': can't convert Object to String (Object#to_str gives Integer) (TypeError)
```

Source code to reproduce the issue (for `String#include?`):

```ruby
a = Object.new
def a.to_str() 1 end
"".include?(a)
```

### DIFF
--- a/src/array_object.cpp
+++ b/src/array_object.cpp
@@ -897,10 +897,9 @@ Value ArrayObject::join(Env *env, Value joiner) {
 
             auto to_str = "to_str"_s;
             auto to_s = "to_s"_s;
-            if (!joiner->is_string() && joiner->respond_to(env, to_str))
-                joiner = joiner->send(env, to_str);
+            if (!joiner->is_string())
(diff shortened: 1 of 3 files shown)


seven1m (approved):
Oh thank you so much! We need to get better about always using the to_str helper!

PR #692, opened 2022-10-28
Add Array#pack directives 'X' and 'x'

Crystalsage:
This PR implements array pack directives 'X' and 'x'. Related to issue #195.

## 'x' directive
The 'x' directive packs `count` amount of null bytes. Null bytes are packed irrespective of the length of the array.

```ruby
nat> [1,2,3].pack('xxxx')
"\x00\x00\x00\x00"
nat>
```

## 'X' directive
The 'X' directive pops the rightmost `count` amount of characters from the string. `count` is 1 by default.

```ruby
# Pack all integers as characters and remove the rightmost two
nat> [65,66,67].pack('C*X2')
"A"
nat>
```

Asterisk modifiers have no effect on either of the directives. 

## Specs
All specs seem to pass! :)

image

### DIFF
--- a/include/natalie/array_packer/packer.hpp
+++ b/include/natalie/array_packer/packer.hpp
@@ -128,9 +128,33 @@ namespace ArrayPacker {
                     });
                     break;
                 }
-                case 'x':
-                    // TODO
+                case 'x': {
+                    // Asterisks have no effect on this directive.
+                    if (token.star)
+                        break;
+
+                    auto null_byte_count = (token.count < 0) ? 1: token.count;
+                    for (int count = 0; count < null_byte_count; count++) {
+                        m_packed.append_char('\0');
+                    }
                     break;
+                }
+                case 'X': {
+                    // Asterisks have no effect on this directive.
+                    if (token.star)
+                        break;
+
+                    auto amount_of_truncated_bytes = (token.count < 0) ? 1: token.count;
+
+                    // If the packed string is empty or if the amount of bytes to be truncated 
+                    // is greater than the packed string's current length, 
+                    // then we can't truncate it. Raise ArgumentError.
+                    if (m_packed.length() == 0 || (int)m_packed.length() < amount_of_truncated_bytes) {
+                        env->raise("ArgumentError", "X outside of the string");
+                    }
+                    m_packed.truncate(m_packed.length() - amount_of_truncated_bytes);
+                    break;
+                }
                 case '@': {
                     auto count = (token.count < 0) ? 1 : token.count;
                     auto missing_chars = static_cast<nat_int_t>(count) - static_cast<nat_int_t>(m_packed.size());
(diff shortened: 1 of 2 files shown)


Crystalsage:
I force pushed on this commit to improve the code formatting a bit and make a comment more readable.

include/natalie/array_packer/packer.hpp
```diff
@@ -128,9 +128,33 @@ namespace ArrayPacker {
                     });
                     break;
                 }
-                case 'x':
-                    // TODO
+                case 'x': {
+                    // Asterisks have no effect on this directive.
+                    if (token.star)
+                        break;
```
seven1m, line 134: I don't understand how `break` can work here. That means the 'x' token is ignored? But in the Ruby docs [it says](https://github.com/seven1m/ruby_spec/blob/6058ec330b3ed47c55c1dd6feef3ff8c33d6a0bb/core/string/unpack/x_spec.rb#L51-L53):

>it "moves the read index to the end of the string when passed the '*' modifier"

include/natalie/array_packer/packer.hpp
```diff
@@ -128,9 +128,33 @@ namespace ArrayPacker {
...
+                    auto null_byte_count = (token.count < 0) ? 1: token.count;
+                    for (int count = 0; count < null_byte_count; count++) {
+                        m_packed.append_char('\0');
+                    }
+                    break;
+                }
+                case 'X': {
+                    // Asterisks have no effect on this directive.
+                    if (token.star)
+                        break;
```
seven1m, line 145: Similarly, for this `break`, how doe this make the [spec](https://github.com/seven1m/ruby_spec/blob/6058ec330b3ed47c55c1dd6feef3ff8c33d6a0bb/core/string/unpack/x_spec.rb#L22-L24) pass?

>it "moves the read index back by the remaining bytes when passed the '*' modifier"

What I understand from that spec is this (but maybe I'm wrong):

1. Given the code:
 
   ```
   "abcd".unpack("C3X*C")
   ```

2. `C3` consumes three 8-bit bytes, resulting in:

   ```
   [97, 98, 99]
   ```

   ...and leaving the cursor (m_index) here:

   ```
   "abcd"
       ^
   ```

3. `X*` moves the cursor "back by the remaining bytes". In this case, the remaining bytes is `1`, since we are on the last byte of the string. So the cursor moves back to the `c`, like this:

   ```
   "abcd"
      ^
   ```

4. The last directive of `C` grabs an 8-bit byte, which is the representation of `c`, resulting in:

   ```
   [97, 98, 99, 99]
   ```

I'm curious how the specs are passing given that we are ignoring this directive. :thinking:

seven1m (approved):
Looks good!

(You might have seen a notification about some comments I made, but I was just confused about Array#pack vs String#unpack. Ignore those! :hugs:)

PR #693, opened 2022-10-29
String: return string value in the same encoding

andrykonchin:
Fixed String methods:

- `#chomp`
- `#succ`
- `#[]`
- `#byteslice`
- `#slice`/`#slice!`
- `#sub`
- `#split`
- `#ljust`
- `#strip!`/`#lstrip`/`#lstrip!`/`#rstrip`/`#rstrip!`
- `#downcase`
- `#upto`
- `#reverse`
- `#delete_prefix`
- `#delete_suffix`
- `#partition`

These methods might return String value in UTF-8 and ignore encoding of the original String, as far as don't set encoding explicitly and UTF-8 is used as a default one.

### Example to illustrate the issue

```ruby
"hello".encode("US-ASCII").sub("ll", "*").encoding
# => UTF-8
```

But expected one is US-ASCII.

### Tests

Not sure whether RubySpec project imports spec changes made in Natalie so I will add specs to reproduce the found issues directly in RubySpec.

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -162,7 +162,7 @@ Value StringObject::chomp(Env *env, Value record_separator) {
 
     size_t end_idx = m_string.length();
     if (end_idx == 0) { // if this is an empty string, return empty sring
-        return new StringObject();
+        return new StringObject("", m_encoding);
     }
 
     String global_record_separator = env->global_get("$/"_s)->as_string()->string();
@@ -177,7 +177,7 @@ Value StringObject::chomp(Env *env, Value record_separator) {
             --end_idx;
         }
 
-        return new StringObject(m_string.substring(0, end_idx));
+        return new StringObject(m_string.substring(0, end_idx), m_encoding);
         // When called with custom global record separator and no args
         // don't remove \r unless specified by record separator
     } else if (record_separator.is_null()) {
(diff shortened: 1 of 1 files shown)


src/string_object.cpp
```diff
@@ -1893,7 +1893,7 @@ Value StringObject::ljust(Env *env, Value length_obj, Value pad_obj) {
     }
 
     if (padstr->string().is_empty())
-        env->raise("ArgumentError", "can't pad with an empty string");
+        env->raise("ArgumentError", "zero width padding");
```
andrykonchin, line 1896: Not relevant change, just fixed an error message (this message isn't checked in RubySpec, heh):

```ruby
# ruby 3.1.2p20

"a".ljust(10, "")
(irb):184:in `ljust': zero width padding (ArgumentError)
```

seven1m (approved):

seven1m:
The his is great! This ok to merge now, or should I wait for spec coverage? (I'm ok either way.)

andrykonchin:
Let's have it merged now if you don't mind.

andrykonchin:
I've finally prepared specs for these changes - https://github.com/ruby/spec/pull/974. Probably we can import them after merging the PR.

PR #694, opened 2022-10-29
CSV header_row? and lineno test

imustafin:
Adds `CSV#lineno` spec for #551.

Also adds `CSV#header_row?` method and tests.

And moves some files from `spec` to `test`. Earlier I placed my own tests in `spec` by mistake in #652 (they should go to `test`, right?).


seven1m:
> they should go to test, right?

Yep, that's perfect!

PR #697, opened 2022-10-30
Add String#rjust

andrykonchin:


seven1m:
Thanks for making the methods const!

Issue #608, opened 2022-09-05
Implement methods of `Complex`


ryangjchandler:
Our main approach is to work on one spec from [ruby/spec](https://github.com/ruby/spec/tree/master/core/complex) at a time, implementing as much of the spec as possible, then check it off the list:

* [x] abs2_spec.rb
* [x] abs_spec.rb
* [x] angle_spec.rb
* [x] arg_spec.rb
* [x] coerce_spec.rb
* [x] comparison_spec.rb
* [x] conj_spec.rb
* [x] conjugate_spec.rb
* [x] constants_spec.rb
* [x] denominator_spec.rb
* [x] divide_spec.rb
* [x] eql_spec.rb
* [x] equal_value_spec.rb
* [x] exponent_spec.rb
* [x] fdiv_spec.rb
* [x] finite_spec.rb
* [x] hash_spec.rb
* [x] imag_spec.rb
* [x] imaginary_spec.rb
* [x] infinite_spec.rb
* [x] inspect_spec.rb
* [x] integer_spec.rb
* [x] magnitude_spec.rb
* [x] marshal_dump_spec.rb
* [x] minus_spec.rb
* [x] multiply_spec.rb
* [x] negative_spec.rb
* [x] numerator_spec.rb
* [x] phase_spec.rb
* [x] plus_spec.rb
* [x] polar_spec.rb
* [x] positive_spec.rb
* [x] quo_spec.rb
* [x] rationalize_spec.rb
* [x] real_spec.rb
* [x] rect_spec.rb
* [x] rectangular_spec.rb
* [x] to_c_spec.rb
* [x] to_f_spec.rb
* [x] to_i_spec.rb
* [x] to_r_spec.rb
* [x] to_s_spec.rb
* [x] uminus_spec.rb

herwinw:
Most of these have been fixed by now, the only one remaining is `exponent_spec.rb` where I've gotten kind of confused by the math behind it. The rest of the specs is equal to the current version of ruby_spec.

ryangjchandler:
@herwinw I'm gonna sort the exponent one tonight :^) The math is a little complex (no pun intended), I have to admit.

ryangjchandler:
I believe this one is all done now, whoop! 🎉

seven1m:
Thanks for all your work on this everybody! ❤️

PR #699, opened 2022-10-31
Add more Complex methods

tekknolagi:
Also add new specs from upstream ruby/spec.

Continues #608.

### DIFF
--- a/spec/core/complex/denominator_spec.rb
+++ b/spec/core/complex/denominator_spec.rb
@@ -1,0 +1,13 @@
+require_relative '../../spec_helper'
+
+describe "Complex#denominator" do
+  it "returns the least common multiple denominator of the real and imaginary parts" do
+    Complex(3, 4).denominator.should == 1
+    Complex(3, bignum_value).denominator.should == 1
+
+    Complex(3, Rational(3,4)).denominator.should == 4
+
+    Complex(Rational(4,8), Rational(3,4)).denominator.should == 4
+    Complex(Rational(3,8), Rational(3,4)).denominator.should == 8
+  end
+end
--- a/spec/core/complex/integer_spec.rb
+++ b/spec/core/complex/integer_spec.rb
@@ -1,0 +1,11 @@
+require_relative '../../spec_helper'
+
+describe "Complex#integer?" do
+  it "returns false for a Complex with no imaginary part" do
+    Complex(20).integer?.should be_false
+  end
+
+  it "returns false for a Complex with an imaginary part" do
+    Complex(20,3).integer?.should be_false
+  end
+end
--- a/spec/core/complex/numerator_spec.rb
+++ b/spec/core/complex/numerator_spec.rb
@@ -1,0 +1,19 @@
+require_relative '../../spec_helper'
+
+describe "Complex#numerator" do
+  it "returns self's numerator" do
+    Complex(2).numerator.should == Complex(2)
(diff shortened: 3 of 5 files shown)


tekknolagi:
It's been a long *long* time since I wrote Ruby, but I think from reading the source this (and potentially other methods) will need to construct an instance from the instance's type, not always base `Complex`. This will make `SomeComplexSubclass.numerator` return an instance of `SomeComplexSubclass` instead of base `Complex`.

I am unfortunately not sure how to call `self.class` from Ruby, so I followed the pattern from the other methods.

stevegeek:
Hi @tekknolagi ! 

Just chiming in here, is the problem here actually that `Complex` it not inheriting from `Numeric` ? 

Eg in Ruby

```
> Complex.ancestors
=> [Complex, Numeric, Comparable, Object, PP::ObjectMixin, Kernel, BasicObject]
```

and Numeric implements `integer?` for example, `https://github.com/ruby/ruby/blob/master/numeric.rb#L31`

While in natalie:

```
$ ./bin/natalie -e "puts Complex.ancestors"
[Complex, Object, Kernel, BasicObject]
```
`Numeric` is not the superclass of Complex. 

Numeric already implements `integer?` here: https://github.com/natalie-lang/natalie/blob/master/src/numeric.rb#L146

tekknolagi:
Sorry, accidentally closed. Great call @stevegeek. I made that change.

tekknolagi:
Hm, that's weird. The spec expects `Complex` not to define `positive?` or `negative?`, which `Numeric` provides. I wonder what upstream does.

tekknolagi:
Oh, looks like they call `rb_undef_method`.

src/numeric.rb
```diff
@@ -184,4 +184,7 @@ def numerator
   def denominator
     to_r.denominator
   end
+
+  undef :positive?
```
stevegeek, line 188: Yeah think thats right, `positive?` & `negative?` are [removed from Complex](https://github.com/ruby/ruby/blob/5ccb625fbbd1e774636a9fdbe0bf1c3d38e085d5/complex.c#L2400) in Ruby, but not from `Numeric`, so need to move these to `Complex`

tekknolagi: oh, lol, oops. was editing the wrong file

seven1m (approved):
I admit I don't know a lot about Complex. @stevegeek do you feel good about this PR now?

ryangjchandler:
Complex not inheriting from `Numeric` is my mistake!

stevegeek:
@seven1m @tekknolagi Will take a look bit later and let you know if I notice anything else!

seven1m:
I went ahead and merged. If there's anything we need to change, we can do it in another PR. Thanks all!

stevegeek:
Just one extra comment, seems like the ancestor chain has Comparable twice?

```
TLW:natalie stephen$ ./bin/natalie -e "puts Complex.ancestors"
[Complex, Numeric, Comparable, Comparable, Object, Kernel, BasicObject]
```

Compared to Ruby

```
irb(main):005:0> Complex.ancestors
=> [Complex, Numeric, Comparable, Object, PP::ObjectMixin, Kernel, BasicObject]
```

seven1m:
Whoops! Yeah that's wrong. It might be just reporting it twice though... will you file an issue?

stevegeek:
Done https://github.com/natalie-lang/natalie/issues/705 !

PR #700, opened 2022-10-31
Comment out missing socket constants

### LINKED ISSUE #633: Socket errors on macOS
Attempting to load the socket library on macOS fails with a number of undeclared identifier errors:

    $ bin/natalie -e "require 'socket'"
    /var/folders/wc/ng7w8yms69j6vpvqck4x2shh0000gn/T/natalie.cpp20221003-18022-x9mqwj:1181:81: error: use of undeclared identifier 'AF_AX25'
        auto const_set75 =  self->const_set(symbols[19]/*:AF_AX25*/, Value::integer(AF_AX25));
    /var/folders/wc/ng7w8yms69j6vpvqck4x2shh0000gn/T/natalie.cpp20221003-18022-x9mqwj:1188:83: error: use of undeclared identifier 'AF_PACKET'
        auto const_set82 =  self->const_set(symbols[26]/*:AF_PACKET*/, Value::integer(AF_PACKET));
    ...
    ...
    fatal error: too many errors emitted, stopping now [-ferror-limit=]
    20 errors generated.

Reading around I think we might need to `#define __APPLE_USE_RFC_3542` before including including <netinet/in.h>[^1] but I can't find the correct place to put that where it'll fix the errors.

Where should that line go where it will be picked up by `require 'socket.cpp'` correctly?

[^1]: See comment in https://opensource.apple.com/source/xnu/xnu-1699.32.7/bsd/netinet6/in6.h

timcraft:
Closes #633

I'm not sure what the "correct" fix is here—we could add some kind of conditional to check if these are defined? For now simply commenting out these missing constants makes it possible to compile the socket code on macOS.


seven1m:
I think #702 will be better, assuming it works on both Linux and Mac. (Only tested on mac so far, but we'll see if CI passes.)

timcraft:
@seven1m Much better 🤌❤️ Thank you!

PR #701, opened 2022-10-31
Fix Addrinfo.new with packed sockaddr string argument

timcraft:
Together with #700 this makes `examples/webserver.rb` work for me on macOS 😎


seven1m:
Fantastic!

PR #703, opened 2022-11-01
Add more Complex methods (part 2)

tekknolagi:
Based on #699 since that was accepted before I added more :)

Continues #608.

### DIFF
--- a/src/complex.rb
+++ b/src/complex.rb
@@ -290,11 +290,36 @@ class Complex
     Complex(real, imag)
   end
 
+  def to_r
+    imaginary = self.imaginary
+    if not _exact_zero?(imaginary)
+      raise RangeError, "can't convert #{self} into Rational"
+    end
+
+    self.real.to_r
+  end
+
   undef :positive?
   undef :negative?
 
   private
   def marshal_dump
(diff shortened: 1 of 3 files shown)


src/complex.rb
```diff
@@ -1,3 +1,19 @@
...
+    return obj.numerator == 0
+  else
+    return obj == 0
+  end
+end
+
+def _exact_zero?(obj)
+  not imaginary.is_a?(Float) and _zero?(imaginary)
+end
+
```
seven1m, line 16: :thinking: I wonder if there's a way to have these little helper methods without polluting the `main` object...?

```
bin/natalie                                                                                                              3.0.4
nat> _zero?(0)
true
nat> methods.grep(/^_/)
[:_zero?, :_exact_zero?, :__send__, :__id__]
nat> 
```

seven1m: I guess It'd be better to pollute `Complex` than to pollute main...

tekknolagi: I could make it `private` on `Complex`. I wish there was a way to have module-private methods.

seven1m:
Nice!

PR #709, opened 2022-11-04
Fix Kernel#is_a? for comparing Module to itself

timcraft:
In the general case where the receiver and the argument are the same value this should return false, but there's an edge case for the `Module` object where it should return true:

    $ irb
    irb(main):001:0> Module.is_a?(Module)
    => true

This also fixes Kernel#instance_of? which currently raises an error with `Module` as an argument, for example:

    $ bin/natalie
    nat> Enumerable.instance_of?(Module)
            2: from (repl):1:in `<main>'
            1: from src/kernel.rb:40:in `instance_of?'
    src/kernel.rb:40:in `raise': class or module required (TypeError)


seven1m (approved):
Nice find!

PR #713, opened 2022-11-04
Implement Marshal.dump and Marshal.load

timcraft:
Test coverage mostly in `test/natalie/marshal_test.rb` for now... a few more details/quirks to implement before we can get the ruby/spec specs passing (float encoding, symbol links, user classes, Struct, Time zones etc)


seven1m (approved):
Whoa cool!!!

PR #714, opened 2022-11-07
Fix String#valid_encoding? for UTF-8

andrykonchin:
There was one failing spec in `spec/core/string/setbyte_spec.rb`. I've investigated it and found out that current implementation of `Utf8EncodingObject::next_char` is not comprehensive and allows false positives.

Changes:
- fixed String#valid_encoding? for UTF-8
- added missing aliases for UTF-8 and US-ASCII

Specs are added to RubySpec here https://github.com/ruby/spec/pull/975.


seven1m (approved):
This is fantastic. You should also update `Natalie::build_top_env()` around [here](https://github.com/natalie-lang/natalie/blob/master/src/natalie.cpp#L178) in order to set the constants on `Encoding`.

andrykonchin:
Done.

PR #715, opened 2022-11-08
Add support for String#-@

nanobowers:
Added String unary minus method.
Currently, there does not appear to be any support for deduplication of frozen strings in Natalie, so some related tests have been skipped for now.  Moves test coverage in a positive direction.


seven1m:
Thank you!

PR #716, opened 2022-11-09
Implement String.try_convert

nanobowers:
Implemented `String.try_convert`, which is structurally similar to `Array.try_convert`.
Related to #217

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -2517,4 +2517,28 @@ Value StringObject::sum(Env *env, Value val) {
 
     return Value::integer(sum);
 }
+
+Value StringObject::try_convert(Env *env, Value val) {
+    auto to_str = "to_str"_s;
+
+    if (val->is_string()) {
+        return val;
+    }
+
+    if (!val->respond_to(env, to_str)) {
+        return NilObject::the();
+    }
+
+    auto result = val->send(env, to_str);
+
+    if (result->is_string() || result->is_nil())
+        return result;
+
+    env->raise(
+        "TypeError", "can't convert {} to String ({}#to_str gives {})",
+        val->klass()->inspect_str(),
+        val->klass()->inspect_str(),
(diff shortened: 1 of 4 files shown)


ryangjchandler:
@nanobowers There's a small merge conflict preventing this from being merged right now, would you be able to take a look at that please?

src/string_object.cpp
```diff
@@ -2517,4 +2517,32 @@ Value StringObject::sum(Env *env, Value val) {
...
+
+    if (conversion->is_string() || conversion->is_nil()) {
+        return conversion;
+    }
+
+    auto original_item_class_name = val->klass()->inspect_str();
+    auto new_item_class_name = conversion->klass()->inspect_str();
+    env->raise(
+        "TypeError",
+        "can't convert {} to Array ({}#to_str gives {})",
```
seven1m, line 2542: I think this should say `String` instead of `Array`.

src/string_object.cpp
```diff
@@ -2517,4 +2517,32 @@ Value StringObject::sum(Env *env, Value val) {
...
+        original_item_class_name,
+        original_item_class_name,
+        new_item_class_name);
+}
```
seven1m, line 2546: This whole method is very similar to `Object::to_str`, I wonder if we can just call that method inside here?

nanobowers: I didn't previously think to do that, but just tried it and it failed one of the specs:
```
Failed specs:
String.try_convert
  sends #to_str to the argument and returns the result if it's nil
    can't convert MockObject to String (MockObject#to_str gives NilClass) (TypeError)
      spec/core/string/try_convert_spec.rb:22:in `try_convert'
```
It seems that the behavior is subtly different in how nils are handled between the different approaches.

seven1m: Ahhh, ok! Very well, let's just use your code then.

PR #717, opened 2022-11-12
Implement some time stuff

richardboehme:
This implements some random time stuff. 

The `Struct#[]` method is actually needed to implement `Time#new` (the specs depend on the Struct method). However, I could not finish the work on `Time#new`, because there seems to be a bug in how we dispatch the `new`-method if redefined.

The following code should print `String` but with Natalie it prints `MethodHolder`. Maybe somebody has an idea 😅 

```ruby
class MethodHolder
  class << self
    define_method(:new, &String.method(:new))
  end
end

p MethodHolder.new.class
```

### DIFF
--- a/include/natalie/time_object.hpp
+++ b/include/natalie/time_object.hpp
@@ -46,6 +46,7 @@ public:
     Value to_r(Env *);
     Value to_s(Env *);
     Value usec(Env *);
+    Value utc_offset(Env *) const;
     Value wday(Env *) const;
     Value yday(Env *) const;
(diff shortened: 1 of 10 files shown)


seven1m:
Fascinating! File an issue about the define_method thing please. I might have time to look at it this next week.

richardboehme:
> Fascinating! File an issue about the define_method thing please. I might have time to look at it this next week.

Thank you! I've created an issue! (see #718)

PR #719, opened 2022-11-13
File and Dir additions and fixes

nanobowers:
## Adding `Dir` and improving `File` classes:

+ Add Dir class with methods for `mkdir` and `rmdir`
    + these are mainly to support mocking and mspec features and not heavily tested.  mkdir modes are not supported
+ Rename `file` and `directory` C++ methods of FileObject to `is_file` and `is_directory`
    + this is mainly for consistency due to the need to differentiate `symlink` and `is_symlink` C++ methods which correspond to `symlink` and `symlink?` Ruby methods
 +   Add methods to FileObject: `is_symlink`, `is_pipe`, `is_socket`, `is_readable`, `is_writable`, `is_executable`, `is_zero`, `is_identical`, `link`, `symlink`, `mkfifo`, and `chmod`
+  Fix FileObject `unlink`, `is_directory` and `is_file` to accept objects that respond to `#to_path`
+  Add `LOCK_*` constants to FileObject (File)

## Modifications and additions to integrated spec system

+  Add fs and mock_to_path spec_helpers from mspec, which are frequently called by File and Dir tests.  
+  Fix `platform_is` ruby-side spec function to support multiple arguments, e.g.: `platform_is :linux, :bsd, :android`
+  Fix issue with tmp file generation failing to work properly
+  Add `after(:all)` for spec context
+  Add `as_superuser` function stub which is empty.

Note that none of the file/dir specs were included in this PR, however the expectation is that the following will work cleanly with no modification in `spec/core/file`: `readable_spec`, `writable_spec`, `executable_spec`, `pipe_spec`, and `link_spec`.
Furthermore, there should be nearly good coverage on `empty_spec`, `zero_spec`, `socket_spec` and `symlink_spec`.  These improvements should also increase the number of passing tests in the nightlies for things like constants, chmod, mkfifo, file, & directory.

One remaining issue, which should be visible once file/dir tests are added is that tmp folders are created and sometimes not removed by the specs, especially if the tests themselves fail.  This is likely a deficiency of the after() not being run when tests raise errors.


src/file_object.cpp
```diff
@@ -7,6 +7,16 @@
 
 namespace Natalie {
 
+// If it's not a string but has a to_path method then execute that method.
+// make sure the path or to_path result is a String before continuing.
+// this is common to many functions probably belongs somewhere else
+Value ConvertToPath(Env *env, Value path) {
```
nanobowers, line 13: It wasnt clear to me where to put this helper method, so i put it here, but I think some of the `Dir` methods would also want to use it, so maybe it should be moved to natalie.hpp/natalie.cpp ?

test/support/spec_helpers/fs.rb
```diff
@@ -0,0 +1,65 @@
...
+    unless path[0, prefix.size] == prefix
+      raise ArgumentError, "#{path} is not prefixed by #{prefix}"
+    end
+
+    # File.symlink? needs to be checked first as
+    # File.exist? returns false for dangling symlinks
+    if File.symlink? path
+      File.unlink path
+    elsif File.directory? path
+      `rm -r #{path}`
```
nanobowers, line 48: Cheating here with a shell call, as `Dir.entries` is not yet implemented

test/support/spec.rb
```diff
@@ -233,19 +255,23 @@ def as_user
...
+
 def little_endian
   yield
 end
 
 def big_endian
 end
 
-@tmp_uniq_id = 0
+TMP_UNIQ_ID = 0
```
seven1m, line 269: Can you talk about this change from ivar to constant? MRI will warn on modifying constants, and we should too.

seven1m:
>Note that none of the file/dir specs were included in this PR, however the expectation is that the following will work cleanly with no modification in spec/core/file: readable_spec, writable_spec, executable_spec, pipe_spec, and link_spec.
>Furthermore, there should be nearly good coverage on empty_spec, zero_spec, socket_spec and symlink_spec. These improvements should also increase the number of passing tests in the nightlies for things like constants, chmod, mkfifo, file, & directory.

Please include the specs from ruby/spec so we know what's passing. We copy those specs in as we implement stuff to get them passing. If there are bits of those specs _not_ passing, then you can change `it` to `xit` and include a `NATFIXME:` comment with the reason it doesn't yet pass.

This helps us not have regressions in the future too!

test/support/spec.rb
nanobowers: Sorry about this.  This change broke the CI and ruby tests though it worked in Natalie on my end.
I think I was trying to fix some behavior and was trying to model off of what goes on in Mspec, but IIRC there they are using a constant with a String value and calling `.succ!` on it which is a bit of trickery.
Regardless, this edit has been reverted with the latest push.

seven1m: Ok, makes sense!

test/support/spec_helpers/tmp.rb
```diff
@@ -0,0 +1,56 @@
...
+    #slash = name.rindex "/"
+    #index = slash ? slash + 1 : 0
+    #name.insert index, "#{SPEC_TEMP_UNIQUIFIER.succ!}-"
+    slash = name.reverse.index("/")
+    index = slash ? name.size - slash - 1 : 0
+    name = name[0...index] + "#{SPEC_TEMP_UNIQUIFIER.succ!}-" + name[index..-1]
+  end
+
+  File.join SPEC_TEMP_DIR, name
+end
```
seven1m, line 56: D'oh! I didn't realize you were copying code from [mspec](https://github.com/ruby/mspec).

At the very least, we have to include a copyright header and a note in the README about such files copied.

Ideally, we would implement this stuff our selves without looking at mspec. I realize that's going to be very hard for you now that you've seen this file, so I'd accept option 1 above. Let's add a copyright notice to this file and the readme.

If you have copied any other files, we'll have to do the same.

I should have mentioned that we only copy files from `ruby/spec` and anything else beyond that, we need to discuss before we copy it. Even though they are the same MIT license, I want to be very careful not to claim ownership of something we didn't create.

test/support/spec_helpers/fs.rb
```diff
@@ -0,0 +1,65 @@
...
+
+# Creates a file +name+. Creates the directory for +name+
+# if it does not exist.
+def touch(name, mode = "w")
+  mkdir_p File.dirname(name)
+
+  File.open(name, mode) do |f|
+    yield f if block_given?
+  end
+end
```
seven1m, line 65: Here is another file that was [copied](https://github.com/ruby/mspec/blob/master/lib/mspec/helpers/fs.rb).

test/support/spec_helpers/tmp.rb
nanobowers: This is a rather misfortunate misunderstanding on my part.  I guess I had not mentally made the distinction between `ruby/spec` and `ruby/mspec`. I realize they are separate repos, but had considered that it was all part and parcel of the same thing.  Regarding the licensing, and the files themselves, for every file from ruby/spec copied in, no additional license header goes into the file included with Natalie, and I had assumed it was covered in the LICENSE.md at the project level and likewise any other "spec" stuff (fixtures, helpers, etc).

To further elaborate on the above, I was under the impression that the self-hosted "mini-mspec" was solely a matter of convenience to bootstrap the Natalie spec environment enough using the features that already existed in Natalie at the time, and once enough of Natalie was complete it would switch over to reference mspec directly.  I didn't get that a project goal was to build and maintain an independent testing environment for the long haul.  Perhaps that part also gets implemented in C++, dunno - I don't have a vision of the project goals.  I'm here because I want to learn more about Ruby internals and brush up on my poor C++ skills and hopefully help out a project at the same time.

In any case, I am interested in either: 

+ Rescinding my PR in its entirety if there is concern over this project being tainted.  
+ Reverting back to a commit prior to inclusion of mspec pieces if that is appropriate.
    + If you'd like any part of the contribution, i am happy to start over and submit piece by piece.
+ Keeping the code, and including the copyright notice in the appropriate place - _It was never my intention to not apply proper attribution_.  I am unclear if the best place would be in the `/test/support/spec_helpers` folder where the mspec files were brought in, in each file individually or at the project level.

seven1m: Let's put some copyright headers in the top of each file copied and call that done. I can do it if that would be helpful.

As for bootstrapping the spec suite until such time we can use mspec... That may be what we end up doing. We may indeed throw away our spec runner and use mspec some day. And in the light, I can certainly see how we would take shortcuts to get there!

But in any case, I think attribution is fine for this, maybe more fine than I implied in the earlier comment.

I do super appreciate all the work you've put into this PR and cannot wait to merge it!

nanobowers:
Finally something that passes through the CI.  The last commit message should say `Fix CI problem with as_superuser`, unsure how to fix that.  The issue regarding root access of the CI was solved by modifying the `as_user` function in `test/support/spec.rb` to temporarily use the `id` command to detect if the job is running as user==0 (root).

seven1m:
Nice! Thanks for all your work on this. 🤗

PR #721, opened 2022-11-19
Process module updates

nanobowers:
Added new sub-modules to `Process`:   `Process::Sys`, `Process::GID`, and `Process::UID`.
Added constants to `Process`, based on `#define`'d values from `sys/resource.h`.  Due to how numbers convert, there were some C++ methods added to `BigInt`
Added some methods to `Process` module and aliased some additional ones that were aliases from `Kernel`
Modified some of the helpers in `test/support/spec.rb` to handle things not previously handled, e.g. exit_code: from `ruby_exe`, fixes to `platform_is/_not`, etc.
If this PR is too large, let me know and I'll split it up.


include/natalie/process_module.hpp
```diff
@@ -11,10 +13,168 @@ namespace Natalie {
...
+#endif
+#ifdef RLIM_SAVED_MAX
+        klass->const_set("RLIM_SAVED_MAX"_s, IntegerObject::create(Integer(BigInt(RLIM_SAVED_MAX))));
+#endif
+#ifdef WNOHANG
+        klass->const_set("WNOHANG"_s, Value::integer(WNOHANG));
+#endif
+#ifdef WUNTRACED
+        klass->const_set("WUNTRACED"_s, Value::integer(WUNTRACED));
+#endif
```
seven1m, line 92: What do you think about defining these in a new `src/process.rb` kind of like we do with [lib/socket/constants.rb](https://github.com/natalie-lang/natalie/blob/master/lib/socket/constants.rb)? Just seems cleaner.

lib/natalie/compiler/binding_gen.rb
```diff
@@ -844,7 +844,34 @@ def generate_name
...
+gen.module_function_binding('Process', 'abort', 'KernelModule', 'abort_method', argc: 0..1, pass_env: true, pass_block: false, return_type: :Object)
+gen.module_function_binding('Process', 'egid', 'ProcessModule', 'egid', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
+gen.module_function_binding('Process', 'egid=', 'ProcessModule', 'setegid', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
+gen.module_function_binding('Process', 'euid', 'ProcessModule', 'euid', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
+gen.module_function_binding('Process', 'euid=', 'ProcessModule', 'seteuid', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
+
+gen.module_function_binding('Process', 'exit', 'KernelModule', 'exit', argc: 0..1, pass_env: true, pass_block: false, return_type: :Object)
+gen.module_function_binding('Process', 'exit!', 'KernelModule', 'exit_bang', argc: 0..1, pass_env: true, pass_block: false, return_type: :Object)
+
+
```
seven1m, line 859: The extra newlines here can be removed?

seven1m (approved):
Nice! (Just a few small things.)

include/natalie/process_module.hpp
nanobowers: Sounds like a good idea, agree that it'd be much cleaner - but i'm not sure how it should be loaded.
The stuff in /lib/ all seems to be `require`d by ruby code as needed, whereas `Process` should loaded automatically as part of ruby core.  Is there a place where process.rb should be called?

PR #724, opened 2022-11-23
Some fixes for macOS

seven1m:
There were some issues preventing Natalie from building on macOS that I fixed.

I'm still working on fixing a bunch of failing Socket-related specs...

I also added a macOS runner to CI so we can hopefully not keep breaking the macOS build. 😅


andrykonchin:
Could we have MacOS-related fixed merged separately plz? It blocks development on MacOS :zany_face:.

PR #727, opened 2022-11-26
Math and Integer

nanobowers:
This should complete the remainder of the functions in `Math` : `Math::ldexp` and `Math::lgamma`

`ldexp` required implementing `Kernel::Integer()`.  This is a "complete" implementation and passes all tests for spec/core/kernel/Integer_spec.rb.  Parsing stringified numbers was the hardest part and I leaned on the c-function `strtoll` to do much of the heavy lifting which may or may not have been appropriate.

To fully test `Kernel::Integer()`, without skips I also added `Integer#quo` and the companion `Rational#quo`, the former is a light wrap of `Rational#div`, the latter is an alias of `Rational#div`

Due to the onion peeling approach I used, this may have been better in reverse order.  If you'd prefer separate PRs, let me know.


seven1m:
Nice work!

PR #730, opened 2022-11-26
Implement JSON.generate

timcraft:


lib/json.rb
```diff
@@ -2,11 +2,105 @@ module JSON
...
+        generate_object(value)
+      else
+        @string << value.to_s
+      end
+    end
+
+    def generate_string(value)
+      @string << '"'
+      if value.is_a?(Symbol)
+        @string << value.to_s
```
seven1m, line 49: A symbol can have weird characters too... we should probably escape this like we do strings.

```
irb(main):002:0> sym = :"foo\0bar\n"
=> :"foo\x00bar\n"
irb(main):003:0> sym.to_s
=> "foo\x00bar\n"
```

timcraft: Yep good catch, I've added another commit to implement the escaping for symbols

PR #731, opened 2022-11-26
Add UTF-32LE encoding

andrykonchin:
Changes:
- added UTF-32LE
- fixed a minor issue with `Integer#chr` error message
- fixed a minor issue with `String#ord` error message

### DIFF
--- a/src/encoding/utf32le_encoding_object.cpp
+++ b/src/encoding/utf32le_encoding_object.cpp
@@ -1,0 +1,87 @@
+#include "natalie.hpp"
+#include <stdio.h>
+
+namespace Natalie {
+
+std::pair<bool, StringView> Utf32LeEncodingObject::prev_char(const String &string, size_t *index) const {
+    if (*index == 0)
+        return { true, StringView() };
+
+    if (*index < 4) {
+        size_t length = *index;
+        *index = 0;
+        return { false, StringView(&string, *index, length) };
+    }
+
+    *index -= 4;
+    StringView character = StringView(&string, *index, 4);
+
+    // check codepoint
+    nat_int_t codepoint = decode_codepoint(character);
+    if (codepoint < 0 || codepoint > 0x10FFFF || (codepoint >= 0xD800 && codepoint <= 0xDFFF)) {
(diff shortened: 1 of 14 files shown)


src/encoding/utf32le_encoding_object.cpp
```diff
@@ -0,0 +1,88 @@
...
+    StringView character = StringView(&string, i, 4);
+    nat_int_t codepoint = decode_codepoint(character);
+    if (codepoint < 0 || codepoint > 0x10FFFF || (codepoint >= 0xD800 && codepoint <= 0xDFFF)) {
+        (*index)++;
+        return { false, StringView(&string, i) };
+    }
+
+    *index -= 4;
+    return { true, StringView(&string, i, 4) };
+}
```
andrykonchin, line 27: Looks like this method isn't tested. What is a proper place for such test?

src/encoding/utf32le_encoding_object.cpp
```diff
@@ -0,0 +1,88 @@
...
+        return { false, StringView(&string, i) };
+    }
+
+    // check codepoint
+    StringView character = StringView(&string, i, 4);
+    nat_int_t codepoint = decode_codepoint(character);
+    if (codepoint < 0 || codepoint > 0x10FFFF || (codepoint >= 0xD800 && codepoint <= 0xDFFF)) {
+        (*index)++;
+        return { false, StringView(&string, i) };
+    }
```
andrykonchin, line 46: Probably `valid_codepoint` function can be used here to avoid duplication.

src/encoding/utf32le_encoding_object.cpp
seven1m: I think it's used by String#chomp

seven1m: We could add some tests for `String#chomp` with different encodings in string_test.rb.

andrykonchin: Done

seven1m:
Nice

PR #737, opened 2022-11-29
Use exact-width integers

seven1m:
I'd rather Natalie fail to build than to have weird behavior; seems like exact-width ints are a way to achieve that. This does fundamentally mean we only support 64-bit systems, but I'm sure that was already the case anyway. Now it's just explicit.

https://stackoverflow.com/a/9837399


seven1m:
Uggh, closing for now. Works on macOS just fine, but on Linux, `int64_t` is aliased to `long`, which breaks all kinds of stuff. :rage:

PR #738, opened 2022-11-29
Implement Array#intersect?

FnControlOption:
```rb
[1, 2].intersect? [2, 3] # => true
[1, 2].intersect? [3, 4] # => false
```


seven1m:
Interesting that one of the comments on that [bug](https://bugs.ruby-lang.org/issues/15198) said the implementation uses a hash.

>The implementation will internally use a hash to match the behavior of Array#&.

PR #739, opened 2022-12-02
Set name of anonymous class if assigned to constant

FnControlOption:
```
nat> Foo = Class.new
#<Class:0x7ff27a005420>
nat> Foo
#<Class:0x7ff27a005420>
```

should be

```
nat> Foo = Class.new
Foo
nat> Foo
Foo
```

For the sake of simplicity, this doesn't handle the edge case of assigning an anonymous singleton class to a constant (which doesn't matter as much since renaming the original class overwrites the singleton name anyways)

```
nat> foo = Class.new
#<Class:0x7ff0c3ef5420>
nat> Foo = foo.singleton_class
Foo
nat> Bar = foo
Bar
nat> Foo
#<Class:Bar>
```

```
irb(main):001:0> foo = Class.new
=> #<Class:0x00007f969d072ac0>
irb(main):002:0> Foo = foo.singleton_class
=> #<Class:#<Class:0x00007f969d072ac0>>
irb(main):003:0> Bar = foo
=> Bar
irb(main):004:0> Foo
=> #<Class:Bar>
```

### DIFF
--- a/lib/natalie/compiler/binding_gen.rb
+++ b/lib/natalie/compiler/binding_gen.rb
@@ -158,18 +158,8 @@ Value #{name}(Env *env, Value klass, Args args, Block *block) {
     end
 
     def args_to_pass
-      case argc
-      when :any
-        [env_arg, 'args', block_arg, klass_arg].compact.join(', ')
-      when Range
-        if argc.end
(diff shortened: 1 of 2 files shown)


seven1m:
Great fix

PR #740, opened 2022-12-03
Implement Dir.chdir and Dir.pwd

FnControlOption:
Todo: add relevant specs


nanobowers:
@FnControlOption what kind of help do you want with this one?

nanobowers:
hi @FnControlOption  I pulled your branch and tested it with `core/dir/chdir_spec` and `core/dir/pwd_spec` with some partial success (>50% of specs passing).

There are a few roadblocks that are probably unrelated to your code contribution and more to the current state of other things in the testing environment:
+ `File.stat` is missing, some tests rely on that
+ There is a problem with the tempfile (`tmp`) generation, which probably will be fixed when natalie gets `File.realpath`
    + The tmpfile issue in the `test/support/spec_helpers/tmp.rb` is such that running ruby (instead of bin/natalie) on the specs in question shows failures on the aforementioned specs due to my somewhat broken workaround ☹️ 
+ I'm fairly sure there is an issue with before_each/before_all/after_each/after_all not always being sequenced at the right times, especially when there is a failure or uncaught exception which makes isolation/debug kind of hard.  More specifically, from what I can tell if a test has an unexpected exception, the code jumps out beyond where the `after(:each)` is called ... to go to the rescue.


In my limited experience with this codebase, i've found that it's easy to get into yak-shaving trying to get some of the prerequisite test features implemented to add additional code.  As far as whether to `xit` the failing specs, or do something else as a workaround, would need guidance from @seven1m

seven1m:
Let's do the simplest possible thing to get this merged. If you have to `xit` out some tests, then that's fine. Merged code is better than code stuck in a PR, as long as we have a marker to know to come back and fix skipped tests.

nanobowers:
@FnControlOption and @seven1m  Not sure it is possible for me to contribute additional updates to this PR from my end since i dont have write access.  Should I branch @FnControlOption 's branch, add spec updates and then push it as a different PR?  Unsure what's appropriate here.

seven1m:
I invited you to the organization @nanobowers

nanobowers:
I hope I did this right.  I branched this `pwd` branch to another one locally, rebased to the latest master branch, then added the specs and made a few minor fixes to both the C++ code and fixed a couple of bugs in the test/support code.  Anything related was `xit`-ed with a NATFIXME note.  That was then remerged with `pwd`

test/support/spec.rb
```diff
@@ -1335,13 +1326,31 @@ def run_specs
         formatter.print_error(*@errors.last)
       else
         formatter.print_success(context, test)
+      ensure
```
nanobowers, line 1329: Adjusting where `after(:each)` and `after(:all)` live in the spec runner.  Previously there were problems where exceptions might prevent the after clause from running, and the after(:all) was being run after the first test, which was not good.

This is "more correct" but still not 100% correct, as all of the `after(:all)` clauses are run completely at the end of the test rather than executed on the final spec of each context

test/support/spec_helpers/tmp.rb
```diff
@@ -73,9 +73,8 @@ def tmp(name, uniquify = true)
     #index = slash ? slash + 1 : 0
     #name.insert index, "#{SPEC_TEMP_UNIQUIFIER.succ!}-"
     slash = name.reverse.index("/")
-    index = slash ? name.size - slash - 1 : 0
+    index = slash ? name.size - slash : 0
```
nanobowers, line 76: bugfix exposed when testing `Dir.chdir`, need to implement `String#rindex` and `String#insert` so that we can revert to the original ruby_mspec implementation.

nanobowers:
Mostly a note to self - it will be worth revisiting the specs that were `xit`ed once the `File::Stat`/ `File.stat` implementation is finished, see #744

PR #743, opened 2022-12-05
FileTest and aliases

nanobowers:
Add `FileTest` module and aliases of `File` class methods with specs
Add `FileTest` module methods: `size`, `size?`, `owned?`, `executable_real?`, `readable_real?`, and `writable_real?` and associated `File` aliased class methods
Fix `File.exist?` to accept objects responding to `to_path`
Fix `Dir.mkdir` to accept mode argument

Add aliased methods in Kernel and String

### Test support fixes
+ Fix `as_user` and `as_superuser` in test/support/spec.rb to use `Process.uid/euid` instead of shell calls.
+ Add `as_real_superuser` in test/support/spec.rb
+ Change temporary use of `rm -r` to `rm -rf` to prevent interactive prompt issue when trying to delete files as part of spec_helpers.  These shell calls will eventually be removed once more methods from `Dir` and `File` (e.g. File.delete) are fully implemented.

This has some somewhat disjoint things it it, can split the PR if it would help.


nanobowers:
AFAICT, this is bogged down due to tests failing on when run as superuser, and due to somewhat broken support for bare octals which are presumably part of natalie_parser.  I'm unsure if the best way forward is to `xit` some tests, hack the tests (e.g. change `0766` to `0o766`) temporarily w/ a NATFIXME or just wait until the octal fix goes into the parser.

PR #744, opened 2022-12-10
Add File::Stat

nanobowers:
Added the `File::Stat` class, and an implementation of `File::Stat#ftype` that is just functional enough so that the `spec/core/file/stat/new_spec.rb` test would pass.

I am not sure if I went about this the best way, I took a look at the `SPECIAL_CLASSES_WITHOUT_DEDICATED_TYPES` array in binding_gen.rb but didnt grok how I could use that to my advantage.


nanobowers:
I must have broken something when i tried to rebase, ugh. 😞

include/natalie/file_stat_object.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "natalie/forward.hpp"
+#include "natalie/object.hpp"
+#include <filesystem>
+
+namespace Natalie {
+
+class FileStatObject : public Object {
+public:
+    FileStatObject(const StringObject &other)
```
seven1m, line 11: This argument is unused, yes?

src/file_object.cpp
```diff
@@ -404,4 +404,9 @@ Value FileObject::chmod(Env *env, Value mode, Value path) {
     return Value::integer(1); // return # of files
 }
 
+Value FileObject::stat(Env *env, Value path) {
+    path = fileutil::convert_using_to_path(env, path);
+    return new FileStatObject { path->as_string()->c_str() };
```
nanobowers, line 409: It gets used here, though it is kind of stubbed out at this point.

include/natalie/file_stat_object.hpp
nanobowers: Unsure how to point to a line of code, but there is a comment below showing where the argument gets used in file_object.cpp line 409.  While it's stubbed out right now and the other argument goes unused (probably should be renamed from `other` to `path`, actually)... I do intend to use it it in the future for calls from `File.stat`, and `File#stat`, etc as opposed to `File::Stat.new`

src/file_object.cpp
seven1m: Oh oK I'll wait till you're done. 😁

nanobowers: This is my first foray into adding a new class and was mainly looking for early feedback on this draft to let me know if I really did things the hard way here - before I got too far along - or any other advice you might have.  It's still really unpolished/unfinished.

nanobowers:
This is currently gated on the macos compile and unfortunately I don't have access to a MacOS development environment.
In particular, the issue is that I'm including `<sys/sysmacros.h>` in linux, presumably there is an equivalent way to include the `major()` and `minor()` functions or defined macros.  If someone could offer advice or help on this, would be appreciated.

PR #745, opened 2022-12-12
Dir: Add home empty? delete unlink

nanobowers:
This PR adds class methods and tests for `Dir.home`, `Dir.empty?`, `Dir.delete` and `Dir.unlink`.

`Dir.rmdir` is the same as delete/unlink, but for some reason the tests are different, and it does not pass dependencies on code not yet written - so that test was not included.

Also, a common helper function on the C++ side (`convert_using_to_path`) was pushed into a namespace and shared between `FileObject` and `DirObject`.  Unsure if this is where it should go or how it should be done (maybe a static method on `FileObject` instead?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,8 @@ ENV CXX=$CXX
 ARG NAT_CXX_FLAGS
 ENV NAT_CXX_FLAGS=$NAT_CXX_FLAGS
 
+ENV USER=root
+
 COPY ext ext
 COPY Rakefile Rakefile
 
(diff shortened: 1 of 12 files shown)


nanobowers:
@seven1m I think I diagnosed this to be related to the docker image not setting the `$USER` env var, which is used by some of these tests.  I'm pretty sure the right thing to do here would be to add `ENV USER=root` to the Dockerfile, so am going to try that, though there is some differing opinion here:
https://stackoverflow.com/questions/54411218/docker-why-isnt-user-environment-variable-set
so if you have a preference, please lmk.

seven1m:
Your solution looks good to me. Is this ready to merge?

nanobowers:
Ready to merge!

Issue #742, opened 2022-12-05
block_given? shadowing


nanobowers:
There is a function in mspec to do a file `touch` that can be executed with or without a block.  If a block is given, the typical use model appears to be to write some data into the touched file.

In any event, the nominal touch function looks like:

```ruby
def touch(name, mode = "w")
  mkdir_p File.dirname(name)
  File.open(name, mode) do |f|
    yield f if block_given?
  end
end
```

when executed as 
```ruby
    touch(@exists) { |f| f.write 'rubinius' }
```

... the block is not yielded. (the resulting file is empty)

I have found that by saving off the block_given? result outside of the `File.open` this works as expected.

```ruby
def touch(name, mode = "w")
  mkdir_p File.dirname(name)
  has_block = block_given?
  File.open(name, mode) do |f|
    yield f if has_block
  end
end
```

It feels like somehow the File.open block is messing up the `block_given?`, perhaps some subtle scoping issue?

PR #746, opened 2022-12-12
Small fixes to test support

nanobowers:
Add `be_true_or_false` test method.
Provide workaround of issue #742 for `test/support/spec_helper/fs.rb`

### DIFF
--- a/include/natalie/string_object.hpp
+++ b/include/natalie/string_object.hpp
@@ -354,6 +354,7 @@ public:
     virtual String dbg_inspect() const override;
 
     virtual void visit_children(Visitor &visitor) override final {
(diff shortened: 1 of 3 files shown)


seven1m:
I'll have to look at this myself. This appears to be the GC collecting something that it shouldn't be, causing a segfault. Debugging these is usually a massive pain in the neck, so it may take me a bit of time.

seven1m:
I believe I fixed the GC issue. I will work on making the gc_lint.rb script a bit smarter so it can detect this sort of thing. (BTW, this was just a bug lurking until the exact right set of circumstances arose... you just got lucky.)

PR #747, opened 2022-12-22
Implement Complex#quo

herwinw:
This checks off one of the items of #608
It turned out to be an alias of `Complex#/`, which was implemented, but not yet checked off in #608


seven1m:
Thanks!

PR #750, opened 2022-12-24
Fix missing/failing specs for Complex

herwinw:


herwinw:
This change contains a couple of statements similar to:
```ruby
result = result.numerator if result.denominator == 1
```
In MRuby, this is done by a helper function in the C code:
```c
VALUE
rb_rational_canonicalize(VALUE x)
{
    if (RB_TYPE_P(x, T_RATIONAL)) {
        get_dat1(x);
        if (f_one_p(dat->den)) return dat->num;
    }
    return x;
}
```
This helper method should not be exposed to the user of Ruby. If Natalie were to implement such a helper, this should probably be stored in the C++ code without exposing it to Ruby, which means all these methods using it would need to be rewritten in C++ as well.

ryangjchandler:
@herwinw You could probably add a `private` method to the `Complex` object so that it doesn't get exposed via the public API, seems like a reasonable way to avoid re-writing everything in C++.

seven1m:
Yeah I'd be fine with that.

seven1m (approved):

herwinw:
I've been considering the option of a private method, but since private in Ruby is almost more of a suggestion than a rule (you can still use `send` to call it, and it shows up in the `private_instance_methods` call), I decided against it, since it would result in an interface that is slightly different from the regular Ruby.

For now it's a oneliner that occurs three times. I don't think we're going to need it more than this (although there are still a few missing specs for Complex, and there might be some for Rational as well). For now I would advise to just keep the duplication as it is, it can always be refactored at a later time.

PR #751, opened 2022-12-24
Some minor optimizations for TM::String

herwinw:
Mostly implementing (and using) move semantics. This can prevent some memory moves.

### DIFF
--- a/include/tm/string.hpp
+++ b/include/tm/string.hpp
@@ -75,6 +75,27 @@ public:
     }
 
     /**
+     * Constructs a new String by moving the contents
+     * from an existing String.
+     *
+     * ```
+     * auto str1 = String { "foo" };
+     * auto str2 = String { std::move(str1) };
+     * assert_str_eq("foo", str2);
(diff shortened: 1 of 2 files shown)


herwinw:
I did a few tests locally with a C++ script that used the move semantics of this class, running it in Valgrind showed that there were no memory leaks in this code.

seven1m:
What did you change to make clang-tidy happy?

herwinw:
Line 336 has been added:
```c++
String &operator=(String &&other) {
    delete[] m_str;
```
It's what happens when you call the following code:
```c++
TM::String str1("foo");
TM::String str2("bar");
str1 = std::move(str2);
```
In that case, you first want to free the memory of `str1.m_str`, before you copy the pointer of `str2.m_str`. So clang-tidy was correct, but the output was very confusing (it pointed at the copy of `m_length`)

seven1m:
OK cool. Yeah I was super confused by the clang-tidy error. 😅

PR #754, opened 2022-12-26
Update docs for move constructor of TM::Optional

herwinw:
The documentation showed the use of the copy constructor, not the move constructor.

### DIFF
--- a/include/tm/optional.hpp
+++ b/include/tm/optional.hpp
@@ -56,7 +56,7 @@ public:
      * ```
      * auto obj = Thing(1);
      * auto opt1 = Optional<Thing>(obj);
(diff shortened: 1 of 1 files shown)


herwinw:
This was an oversight from me in #752, where the documentation has been copied from the copy constructor, and this line was not changed.

Which actually makes me wonder: is there any system in place to validate these comments? I've seen a header file defining`Thing`, and there are definitions for things like `assert_eq`, but this code snippet should have failed because `assert_not(opt1);` would not be true because `opt1` would have been copied in this case.

seven1m:
Yes, there is a spec runner over at [seven1m/tm](https://github.com/seven1m/tm). I'll copy the recent changes over there eventually.

PR #755, opened 2022-12-26
Add converting move constructor for TM::Optional

herwinw:
This one was still missing. It reduces the number of copies.

### DIFF
--- a/include/tm/optional.hpp
+++ b/include/tm/optional.hpp
@@ -18,9 +18,22 @@ public:
      * assert(opt);
      * ```
      */
-    Optional(T value)
+    Optional(const T &value)
         : m_present { true }
         , m_value { value } { }
+
+    /**
+     * Constructs a new Optional with a value.
+     *
+     * ```
+     * auto obj = Thing(1);
+     * auto opt = Optional<Thing>(std::move(obj));
+     * assert(opt);
+     * ```
+     */
+    Optional(T &&value)
+        : m_present { true }
+        , m_value { std::move(value) } { }
 
     /**
      * Constructs a new Optional without a value.


include/tm/optional.hpp
```diff
@@ -18,10 +18,23 @@ class Optional {
      * assert(opt);
      * ```
      */
-    Optional(T value)
+    Optional(const T &value)
```
seven1m, line 21: I admit I'm a C++ noob, but I don't understand how this works with the case `Optional<SharedPtr<String>>`... The `SharedPtr` type must be copied in order to keep count of live pointers. If we don't ever copy the SharedPtr, but instead store a reference in the Optional, won't the SharedPtr count be wrong?

herwinw: Let's explain this by copying the relevant parts of the old source (I'm going to skip over the `m_present` variable, since that one is not relevant):
```c++
Optional(T value) : m_value { value } {  }
```
And imagine this calling code:
```c++
const auto t1 = Thing(1);
const auto o1 = TM::Optional(t1);
```
This code actually makes two copies of `t1`: `value` from the method signature is a copy of `t1`. The statement `m_value { value }` creates another copy by calling the constructor of `Thing` with the argument `value`, and storing the result in `m_value`.

I've copy-pasted some code from Thing to make it noisy: it prints every time a constructor is called, and how it was called. The orginal code results in the following output:
```
Called Thing(int) with 1
Called Thing(const Thing&) with 1
Called Thing(const Thing&) with 1
```
This changes removes the copy operation in the function signature, but keeps the copy when constructing `m_value`. The result indeed prints one less line:
```
Called Thing(int) with 1
Called Thing(const Thing&) with 1
```

Semi related: the move constructor (`Optional(T &&value)`) is only called when it's has been given an rvalue reference. E.g:
```c++
const auto t1 = Thing(1);
const auto o1 = TM::Optional(t1);
const auto o2 = TM::Optional(Thing(2))
```
`t1` and `o1` now behave as previously, but the added `o2` results in the following output:
```
Called Thing(int) with 2
Called Thing(Thing&&) with 2
```
The difference is that this calls the move constructor `Thing(Thing&&)` instead of the copy constructor seen with `o` (`Thing(const Thing&)`).

PR #761, opened 2022-12-26
Implement Rational(NumericSubclass)

herwinw:
I'm not sure if this is the idiomatic way to do an `is_a` call from C++ code. The code feels a bit clunky compared to the `respond_to` in the same line, but something like `x->is_a?("Numeric"_s)` does not appear to be working.


seven1m (approved):
Yeah that's the best we got for now.

PR #762, opened 2022-12-26
Complete the constructors and assignment operators of Thing

herwinw:
Some code snippets in the TM library are using the move constructor of Thing, so it might be a good idea to actually speficy that.

I've made this PR over here instead of at https://github.com/seven1m/tm, since this repository contains all the other updates for `TM` as well, so this is the most up to date version that can be copied over to the other locations.

### DIFF
--- a/include/tm/tests.hpp
+++ b/include/tm/tests.hpp
@@ -62,13 +62,30 @@ class Thing {
 public:
     Thing() = default;
 
-    Thing(int value)
+    Thing(const int value)
         : m_value { value } { }
 
-    Thing(const Thing &other)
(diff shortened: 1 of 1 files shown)


herwinw:
Related: https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_five

herwinw:
The move operations now reset the source object, so it can be used in tests as well:
```c++
auto t1 = Thing(1);
Thing t2(std::move(t1));
assert_eq(0, t1.value()); // t1 has been reset
assert_eq(1, t2.value());
```

PR #768, opened 2022-12-28
Fix unit tests of include/tm

herwinw:
* string.hpp:
This was a compile error, there were two variables named `str`.

* optional.hpp:
The value from ob2 was moved. Thing has been updated to reset the value after a move operation, so we need a new Thing object to validate the Thing inside the optional.

I've found these by copying the files to a checkout of the `tm` repo and running the unit tests over there. I'm fixing them here, since this repository contains the latest version of the tm headers.


seven1m:
Thanks!

PR #771, opened 2022-12-30
Implement String#to_i for BigInt values

herwinw:
Mostly just to let people know I'm working on this, just in case somebody else would decide to take a shot at this (very small chance, I know).

I'll probably finish it tomorrow, the remainder is pretty straight forward if you take the simple solution of walking backwards, using 1 character every time and carrying over the result, but I can think of a few optimizations to do this in bigger chunks

### DIFF
--- a/include/natalie/big_int.hpp
+++ b/include/natalie/big_int.hpp
@@ -64,7 +64,7 @@ public:
     BigInt(const unsigned int &);
     BigInt(const int &);
     BigInt(const double &);
-    BigInt(const TM::String &);
+    BigInt(const TM::String &, const int base = 10);
 
     // Assignment operators:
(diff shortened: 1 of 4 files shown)


herwinw:
Current status: it works, but it's not really efficient. It can be merged, it does extend the functionality a bit, but I'll just leave it at draft status because it could be much faster. But I'm not going to be able to do that this weekend.

herwinw:
This should do it. Input is now parsed in chunks of 12 characters. I didn't measure anything, but running the relevant spec file feels about two seconds faster.

One possible optimization might be to make the chunk size depend on the base: the current value 12 is the maximum size that works for the maximum base (36). This could be bigger for smaller bases.

seven1m:
Sweet!

PR #775, opened 2023-01-01
Miscellaneous minor fixes for String

nanobowers:
In support of #217

+ Support for kwargs on `String#initialize` (note that `capacity:` kwarg doesn't currently do anything)
+ Import spec for `String#length`
+ Add support for `String#end_with?`
+ Add support for `String#index` optional second argument
+ Fix a bug on `String#split` splitting with an empty-string which previously got stuck in a while-loop and caused test timeouts
+ Fix raise on frozen string for `String#succ!`


src/string_object.cpp
```diff
@@ -516,11 +537,19 @@ nat_int_t StringObject::index_int(Env *env, Value needle, size_t start) const {
     return ptr - c_str();
 }
 
-Value StringObject::initialize(Env *env, Value arg) {
+// TODO: according to documentation, "the given capacity is
+//  advisory only, and may or may not set the size of the internal
+//  buffer, which may in turn affect performance".  This said, it
+//  should be possible to set the m_capacity field of the
+//  underlying String object in the future.
```
seven1m, line 544: I don't think the we need the TODO note here. The `capacity` is an implementation detail and I'm fine with us ignoring it indefinitely. 😄

PR #778, opened 2023-01-02
Update some specs, resolve the "NATFIXME: bug in spec" entries

herwinw:
I went through the chore of upstreaming all these FIXME statements. I tried to give credit to the original author, but since I just used git blame this might sometimes be the wrong person (in case did a minor change to the line after the FIXME was added). All the FIXME statements now have the upstream pull requests listed in them, so it's easier to match them.

I marked this pull request as a draft, response time from ruby-spec is generally pretty quick, so I guess I can update the PR later today with the FIXMEs removed.

### DIFF
--- a/spec/core/array/keep_if_spec.rb
+++ b/spec/core/array/keep_if_spec.rb
@@ -1,4 +1,3 @@
-# NATFIXME: bug in spec; missing require_relative '../../spec_helper'
 require_relative '../../spec_helper'
 require_relative 'shared/keep_if'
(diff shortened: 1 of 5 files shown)


TheGrizzlyDev:
>  I tried to give credit to the original author, but since I just used git blame this might sometimes be the wrong person

thanks, much appreciated! :heart:

seven1m:
Wow thanks for upstreaming these! ❤️

PR #779, opened 2023-01-02
Fix escaping of \ in string output of Regex

herwinw:
The \ character might need some extra escaping, but not when it's used for a character class. The old code had the following output:

    /\d+/.to_s == "(?-mix:\\\\d+)"

This changed the behaviour: instead of matching one of more decimals, this regex would match a literal \ followed by one or more d's. This made it impossible to embed these kind of regexes into another regexp.

This scenario was not defined in the regex specs, but it did accidentally get caught in the spec for the case statement.

### DIFF
--- a/spec/language/case_spec.rb
+++ b/spec/language/case_spec.rb
@@ -116,8 +116,7 @@ describe "The 'case'-construct" do
     $1.should == "44"
   end
 
-  # NATFIXME: regexp interpolated within another regexp
-  xit "tests with a regexp interpolated within another regexp" do
+  it "tests with a regexp interpolated within another regexp" do
     digits_regexp = /\d+/
     case "foo43"
     when /oo(#{digits_regexp})/
--- a/src/regexp_object.cpp
+++ b/src/regexp_object.cpp
@@ -56,7 +56,11 @@ Value RegexpObject::inspect(Env *env) {
             if (i < (len - 1) && str[i + 1] == '/') {
                 break;
             }
-            out->append("\\\\");
+            out->append("\\");
+            if (i < (len - 1) && str[i + 1] == '\\') {
+                out->append("\\");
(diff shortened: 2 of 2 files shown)


herwinw:
This is still not right, even though all the specs pass. The following works in Ruby:
```ruby
/\\\d+/.to_s == "(?-mix:\\\\\\d+)"
```
But the output of Natalie is still different.

I'll probably make some improvements to the specs too, once I figured out what the actual rules are.

herwinw:
New version, this one passes the following extra specs:
```ruby
it "herwin: missing test 1" do
  Regexp.new('\d+').inspect.should == "/\\d+/"
  /\d+/.inspect.should == "/\\d+/"
  /\d+/.to_s.should == "(?-mix:\\d+)"
end

it "herwin: missing test 2" do
  Regexp.new('\\d+').inspect.should == "/\\d+/"
  /\\d+/.inspect.should == "/\\\\d+/"
  /\\d+/.to_s.should == "(?-mix:\\\\d+)"
end

it "herwin: missing test 3" do
  Regexp.new('\\\d+').inspect.should == "/\\\\d+/"
  /\\\d+/.inspect.should == "/\\\\\\d+/"
  /\\\d+/.to_s.should == "(?-mix:\\\\\\d+)"
end

it "herwin: missing test 4" do
  Regexp.new('\\\\d+').inspect.should == "/\\\\d+/"
  /\\\\d+/.inspect.should == "/\\\\\\\\d+/"
  /\\\\d+/.to_s.should == "(?-mix:\\\\\\\\d+)"
end
```
The syntax `Regexp.new(String)` has the internal escape of string input, which is why these behave differently from the `//` syntax. I generated the expected values with some copy-paste into IRB, just to validate the behaviour was the same as MRI.

PR #780, opened 2023-01-02
Add UTF-16LE encoding

andrykonchin:


test/natalie/string_test.rb
```diff
@@ -554,6 +554,48 @@
...
+
+        # FIXME
+        # s = 'foo'.encode('UTF-16LE') + "\xFF".force_encoding('UTF-16LE') # incorrect 1-byte sequence
+        # s.chop!
+        # s.bytes.should == 'foo'.encode('UTF-16LE').bytes
+
+        # FIXME
+        # s = 'foo'.encode('UTF-16LE') + "\x00\xD8\x00".force_encoding('UTF-16LE') # incorrect 3-byte sequence
+        # s.chop!
+        # s.should == 'foo'.encode('UTF-16LE') + "\x00\xD8".force_encoding('UTF-16LE')
```
seven1m, line 585: @andrykonchin are these `FIXME` notes for a future us to fix, or do you want more time on this PR?

andrykonchin: Yeah, it's for a future.

Right now I don't have any idea how to delete incorrect bytes properly (or at least how CRuby does) when we check bytes from the end of a byte sequence (with `EncodingObject.prev_char`). 

As far as UTF-16 is a variable character length encoding we can remove 1, 2 or 4 bytes. It's simple if we traverse bytes array from the beginning, but traversing from the end is an issue.

PR #781, opened 2023-01-03
Enabled a few specs

herwinw:
The issues have been fixed, but these specs were still disabled.


seven1m:
Thanks so much for all the cleanup work you've been doing lately. :hugs:

PR #783, opened 2023-01-03
Reword FIXME in specs to NATFIXME

herwinw:
This way it's a bit easier to distinguish between FIXME markers from ruby-spec upstream and FIXME markers that indicate a missing feature in Natalie.

I skipped a few of them: some are fixed via #781 and #782, and one other would conflict with the changes in #780.


seven1m:
There was a conflict with master that I think I fixed up. You might take another peek at it just to make sure @herwinw.

herwinw:
I did a rebase on my local branch and compared the two: they're exactly the same

Issue #667, opened 2022-10-21
Implement String#unpack


seven1m:
I split this issue off from #217.

We implement these in [StringUnpacker](/blob/master/include/natalie/string_unpacker.hpp).

- [x] unpack1_spec.rb
- [x] a_spec.rb
- [x] at_spec.rb
- [x] b_spec.rb
- [x] comment_spec.rb
- [x] c_spec.rb
- [x] d_spec.rb
- [x] e_spec.rb
- [x] f_spec.rb
- [x] g_spec.rb
- [x] h_spec.rb
- [x] i_spec.rb
- [x] j_spec.rb
- [x] l_spec.rb
- [x] m_spec.rb
- [x] n_spec.rb
- [x] percent_spec.rb
- [x] p_spec.rb
- [x] q_spec.rb
- [x] s_spec.rb
- [x] u_spec.rb
- [x] v_spec.rb
- [x] w_spec.rb
- [x] x_spec.rb
- [x] z_spec.rb

seven1m:
Closing!! 🎉

PR #787, opened 2023-01-04
Implement Array#pack('w') and String#unpack('w')

herwinw:
Another fix for #667

The current state is very much a work in progress, I'm just pushing it here prematurely to both let people know somebody is working on this and to keep myself accountable for actually finishing it

### DIFF
--- a/include/natalie/array_packer/integer_handler.hpp
+++ b/include/natalie/array_packer/integer_handler.hpp
@@ -35,7 +35,7 @@ namespace ArrayPacker {
         void pack_s();
         void pack_V();
         void pack_v();
-        void pack_w();
+        void pack_w(Env *env);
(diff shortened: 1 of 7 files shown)


herwinw:
@seven1m Is it okay if I start by splitting the header file into a header and a cpp file? Currently I have to recompile pretty much the whole application when I change an internal of `unpack_w`, which feels a bit off.

seven1m:
Yeah that's fine. Go for it!

herwinw:
This should fix item for both #195 and #667. We can probably close #667 once this one is merged: there are two FIXMEs left in `spec/core/string/unpack`, but those are the "cannot unpack a string except from the same object that created it, or a duplicate of it" which is really deep into the internals, and does not exactly match the "good first issue" tag.

PR #788, opened 2023-01-05
ENV updates

nanobowers:
+ Fix `ENV.delete` to support block syntax
+ Fixed off-by-one bug in `ENV.size`
+ Add `ENV.clear` and `ENV.replace`
+ Import other passing ENV specs, many were dependent on `ENV.replace`


nanobowers:
@seven1m this got merged before all the CI tests passed.  macos is failing one test and it's not obvious to me why, and i don't have a way to test this, so probably the quickest thing to do would be to delete the spec/core/env/delete_spec.rb to get this back to green.  Apologies, I probably should tag everything with *draft* if I'm not confident they are mergeable.

seven1m:
Oops! My bad, I just got trigger happy. I'll work on fixing it.

PR #790, opened 2023-01-05
Split packer and unpacker code into header and cpp file

herwinw:
And synced the specs folders for these methods as well, some specs had upstream updates, some specs were disabled for issues that have been fixed.


herwinw:
Don't worry about merge conflicts with #787, that one is still a work in progress, I'll fix that with a rebase after this one is merged.

herwinw:
The failing test looks like the issue described in #788, and not related to this PR

PR #792, opened 2023-01-06
Remove spec/core/env/delete_spec.rb due to macos issue

nanobowers:
Addresses broken maco ENV.delete spec which fails for unknown reasons, see #788


seven1m:
I fixed the bug in 8b4f13aa7e8bea48035410c44a687b0d27554a99. Sorry I didn't mention that anywhere.

PR #795, opened 2023-01-09
File.realpath

nanobowers:
+ Add `File.realpath`
+ Fix `test/support/spec_helpers/fs.rb` and `tmp.rb` to use correct output path locations and use native ruby functions.  These were originally imported from Mspec and then hacked due to lack of native functionality in Natalie.
+ Import specs for the `File` class unblocked by realpath


test/support/spec_helpers/fs.rb
```diff
@@ -70,9 +69,10 @@ def rm_r(*paths)
     if File.symlink? path
       File.unlink path
     elsif File.directory? path
-      `rm -rf #{path}`
+      # NATFIXME: Revert when Dir.entries and Dir.rmdir are available
```
nanobowers, line 72: this previously should've had a NATFIXME, so added

PR #804, opened 2023-01-11
Fix null byte termination in String#split

herwinw:
The last push to the result array created a String based on a `const char*` without a length. This means an embedded null byte in the last part of the String would prematurely end the last value in the result.

### DIFF
--- a/spec/core/string/split_spec.rb
+++ b/spec/core/string/split_spec.rb
@@ -1,0 +1,642 @@
+# -*- encoding: utf-8 -*-
+require_relative '../../spec_helper'
+require_relative 'fixtures/classes'
+
+describe "String#split with String" do
+  # NATFIXME: Encoding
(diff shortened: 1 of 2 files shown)


seven1m:
GitHub won't let me comment directly on the lines, so I'll comment here:

We should make the same changes on these two lines:

https://github.com/natalie-lang/natalie/pull/804/files#diff-42e7d24554cea8cdc4e2ba1d1414204721b3112f0d609734dcba4862c723987bR1940

https://github.com/natalie-lang/natalie/pull/804/files#diff-42e7d24554cea8cdc4e2ba1d1414204721b3112f0d609734dcba4862c723987bR1915

herwinw:
You're right, I've added those as well.
I noticed this one when I was working on the implementation of `String#lines` and kind of fixed them both at the same time, so I guess I missed those because the `lines` method does not have a max

PR #807, opened 2023-01-12
Some fixes for splat and double splat operations

herwinw:
Calling a method like this:
```ruby
foo(*args, **kwargs)
```
caused the keyword arguments to be added to the positional arguments. This appears to fix it (I added a bunch of tests, and they all pass), but the fix almost looks too simple to be true.

This started out as an attempt to implement `String#each_line`, and quite a few yaks have lined up to be shaved.

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -267,7 +267,7 @@ module Natalie
             instructions: instructions,
             with_block_pass: !!block,
             args_array_on_stack: true,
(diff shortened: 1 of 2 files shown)


herwinw:
Tests fail under Ruby 2.7, but that version had different behaviour for the keyword arguments. I'll try to whip up a local Ruby 2.7 and add some version specifics on the tests.

seven1m:
Nice!

PR #809, opened 2023-01-12
Update max_long and min_long in spec runner

herwinw:
Integer#size is implemented. The old and new code yields the same results.

### DIFF
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -192,16 +192,11 @@ def fixnum_min
 end
 
 def max_long
-  # 2**(0.size * 8 - 1) - 1
-  # NATFIXME: Support Integer#size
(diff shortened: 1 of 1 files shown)


test/support/spec.rb
```diff
@@ -192,16 +192,11 @@ def fixnum_min
 end
 
 def max_long
-  # 2**(0.size * 8 - 1) - 1
-  # NATFIXME: Support Integer#size
-  # NATFIXME: Make Integer#** spec compliant with bignums
```
herwinw, line 197: I was not sure what this FIXME meant. I checked the output of both statements and they result in the same value. If `Integer#**` still needs fixes, I guess that should be documented in one of the specs of Integer, not here

PR #810, opened 2023-01-13
Implement fixture method in spec runner

herwinw:
Most of the specs that use this method still fail for different reasons. Document those reasons with a NATFIXME comment.


herwinw:
Even though there are not that many occurrences of `fixture` in the specs in Natalie, there are some specs that haven't been imported yet that depend on them (for example `core/argf/argv_spec.rb`). I'm really curious as to what the effect of this change will be on the specs graph on the Natalie homepage.

PR #814, opened 2023-01-13
Convert base argument to integer in Kernel.Integer using to_int

herwinw:
There was an explicit check for this in the spec for `String#to_i`, but not in the spec of `Kernel.Integer`. It turned out MRI does this conversion.

I added https://github.com/ruby/spec/pull/997 as an upstream request to add this to the spec.

### DIFF
--- a/src/kernel_module.cpp
+++ b/src/kernel_module.cpp
@@ -135,8 +135,12 @@ Value KernelModule::exit_bang(Env *env, Value status) {
 
 Value KernelModule::Integer(Env *env, Value value, Value base, Value exception) {
     nat_int_t base_int = 0; // default to zero if unset
-    if (base && base->is_integer())
+    if (base) {
+        if (!base->is_integer() && base->respond_to(env, "to_int"_s))
+            base = base->send(env, "to_int"_s);
+
         base_int = base->as_integer()->to_nat_int_t();
+    }
     return Integer(env, value, base_int, exception ? exception->is_true() : true);
 }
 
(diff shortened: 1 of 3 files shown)


src/kernel_module.cpp
```diff
@@ -135,8 +135,12 @@ Value KernelModule::exit_bang(Env *env, Value status) {
 
 Value KernelModule::Integer(Env *env, Value value, Value base, Value exception) {
     nat_int_t base_int = 0; // default to zero if unset
-    if (base && base->is_integer())
+    if (base) {
+        if (!base->is_integer() && base->respond_to(env, "to_int"_s))
+            base = base->send(env, "to_int"_s);
+
```
seven1m, line 141: I think we need an assert_type() call here in case to_int returns something else.

herwinw: Well, actually...

I just tried this in IRB:
```
irb(main):001:0> Integer('123', Object.new.tap { |x| x.define_singleton_method(:
to_int) { 8 } })
=> 83
irb(main):002:0> Integer('123', '8')
=> 123
```
(Keep in mind: `String` does not have a `to_int` method, it has a `to_i` but that is not what's in use here). So if the base argument does not have a `to_int` method, it gets ignored. I'm not sure whether this is the intended behaviour, given that `Kernel.Integer` is otherwise more strict than the `to_i` methods.

I already added a spec update upstream for this issue, that hasn't been merged yet, so I'm just going out on a limb and drop a question here. We might want to wait for a second before we merge this change.

seven1m: >We might want to wait for a second before we merge this change.

OK, holding. 😄

herwinw:
The pull request has been updated:
* Import the latest version of the spec of `Kernel.Integer,` this one includes the `to_int` check
* Call `to_int` if desired (which was the old version)
* Add a commit to ignore the base argument to be compatible with MRI. I reported an upstream bug, and added a link to the bug to the code. I'm a watcher of the bug report, so I should get an update if there are any decisions made about the behaviour. I think that for now we should just follow MRI's behaviour

PR #816, opened 2023-01-14
Make enum_for compatible with keyword arguments

herwinw:
I ran into this issue when working on `String#each_line`, which is one of the few occasions where you can pass on keyword arguments into an enumerator. Something breaks, but I haven't gotten to the exact cause yet. #807 was just the beginning of this quest :wink: 

This case adds a dumb example to show the issue. The spec passes in Ruby (I've used 3.1), and errors in Natalie.

The spec adds an ugly monkey patch for String, which probably should be removed at some points, but it's good enough to test and work with.

### DIFF
--- a/src/kernel.rb
+++ b/src/kernel.rb
@@ -14,11 +14,11 @@ module Kernel
   end
   alias yield_self then
 
-  def enum_for(method = :each, *args, &block)
+  def enum_for(method = :each, *args, **kwargs, &block)
     enum =
       Enumerator.new do |yielder|
         the_proc = yielder.to_proc || ->(*i) { yielder.yield(*i) }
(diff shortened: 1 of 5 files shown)


seven1m:
@herwinw do you want to convert this to an issue and work on it later?

herwinw:
I just retried this one with the trivial change in `kernel.rb` and it appears to be working now (still waiting for the full test suite to run, but the added spec passes). I guess the changes in # #897 may have fixed it.

If this works, I'm going to fix the last bit of `String#each_line` and we're done.

herwinw:
That looks like it has been fixed, even though the optional pop_hash feels like a hack.

seven1m:
Nice!

PR #819, opened 2023-01-14
Implement String#each_line

herwinw:
Or at least like 99% of it. It does not pass the keyword argument `chomp` to the enumerator, the underlying issue is described in # #816


lib/natalie/compiler/binding_gen.rb
```diff
@@ -1085,6 +1085,7 @@ def generate_name
 gen.binding('String', 'each_byte', 'StringObject', 'each_byte', argc: 0, pass_env: true, pass_block: true, return_type: :Object)
 gen.binding('String', 'each_char', 'StringObject', 'each_char', argc: 0, pass_env: true, pass_block: true, return_type: :Object)
 gen.binding('String', 'each_codepoint', 'StringObject', 'each_codepoint', argc: 0, pass_env: true, pass_block: true, return_type: :Object)
+gen.binding('String', 'each_line', 'StringObject', 'each_line', argc: 0..1, kwargs: [:chomp], pass_env: true, pass_block: true, return_type: :Object)
```
seven1m, line 1088:
```suggestion
gen.binding('String', 'each_line', 'StringObject', 'lines', argc: 0..1, kwargs: [:chomp], pass_env: true, pass_block: true, return_type: :Object)
```

In the past, when the same method is available under different names, we have just pointed the binding to a single C++ method.

Then we don't need the `lines_inner` static function. But it's possible there is a reason we need two separate C++ methods... will the behavior be different between `lines` and `each_line` at some point in the future?

herwinw: It is slightly different: if you don't pass in a block, `lines` returns an array, `each_line` returns an enumerator. That's why the inner method has one extra boolean argument.

PR #820, opened 2023-01-15
Various improvements to File and File::Stat

nanobowers:
+    Add `File.atime`, `File.ctime`, `File.mtime` and `File.utime`
+   Add `File#atime`, `File#ctime`, and `File#mtime`
+    Move `File#stat` to `IO#stat`
+   Add `File::Stat#socket?` and Fix `File::Stat#<=>`
+   Replace `FNM_*` constants with MRI compatible constants and add remaining missing `FNM_*` constants
+   Add `File.basename`, `File.split`, `File.extname` and add specs
+   Fix `File.dirname` to pass specs
+  Fix `exactly()` test function


lib/natalie/compiler/binding_gen.rb
```diff
@@ -554,11 +554,18 @@ def generate_name
 gen.static_binding('File', 'realpath', 'FileObject', 'realpath', argc: 1..2, pass_env: true, pass_block: false, return_type: :Object)
 # NATFIXME: realdirpath is subtly different from realpath and should bind to a unique function
 gen.static_binding('File', 'realdirpath', 'FileObject', 'realpath', argc: 1..2, pass_env: true, pass_block: false, return_type: :Object)
+gen.static_binding('File', 'atime', 'FileObject', 'atime', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
+gen.static_binding('File', 'ctime', 'FileObject', 'ctime', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
+gen.static_binding('File', 'mtime', 'FileObject', 'mtime', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
+gen.static_binding('File', 'utime', 'FileObject', 'utime', argc: 2.., pass_env: true, pass_block: false, return_type: :Object)
+
+gen.binding('File', 'atime', 'FileObject', 'atime', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
```
nanobowers, line 562: Re-sorted this list of method bindings

nanobowers:
I found an intermittent bug in the File.atime/mtime tests that was documented in the tests and reported here https://bugs.ruby-lang.org/issues/17926 but never fixed, so will work on upstreaming the fixes.  That is apparently what is causing the tests to fail in CI here (they are guarded in ruby_spec with `ENV['TRAVIS']`)

nanobowers:
Thinking now that the atime/mtime tests are failing due to a  broken implementation of `Integer()` when base 10 is given and there is a leading zero on the string.  Simplest reproduction is comparing results of `Integer("08")` and `Integer("08",10)` on both Natalie (fails, fails) and MRI (fails, works).

Oddly enough this was probably a weird one to hit, b/c it would require the a file modificaiton timestamps microseconds field to start with `0` and contain either an `8` or `9`. 🤔

herwinw:
> Thinking now that the atime/mtime tests are failing due to a broken implementation of `Integer()` when base 10 is given and there is a leading zero on the string. Simplest reproduction is comparing results of `Integer("08")` and `Integer("08",10)` on both Natalie (fails, fails) and MRI (fails, works).

Shall I take a shot at this one? I've been looking at `Kernel.Integer` and `String#to_i` this past week, and I think it's relatively easy to fix

seven1m:
I'm thinking 11d0a9578de0a39ddb95386ec18c3703066197da maybe isn't needed now? (If this were rebased on master, that is.)

nanobowers:
I'll rebase with master and see if those specs that were `xit`ed in https://github.com/natalie-lang/natalie/commit/11d0a9578de0a39ddb95386ec18c3703066197da can be safely brought back in.

PR #825, opened 2023-01-15
Fix coercion calls in Integer operations

herwinw:


herwinw:
(Marking this one as a draft for a second, since there are some hairy yaks lining up)

herwinw:
I added subtraction too, and only then I realized this is the wrong way to fix things.

It started out with this code for adding to an Integer:
```c++
Value IntegerObject::add(Env *env, Value arg) {
    if (arg->is_float()) {
        return Value::floatingpoint(m_integer + arg->as_float()->to_double());
    } else if (arg->is_complex()) {
        return arg->send(env, "+"_s, { this });
    } else if (!arg->is_integer()) {
        arg = Natalie::coerce(env, arg, this).second;
    }
    arg->assert_type(env, Object::Type::Integer, "Integer");

    return create(m_integer + arg->as_integer()->integer());
}   
```
The issue here is in the use of the call to `coerce`. The check for `arg->is_complex()` is a workaround to make a statement like `1 + Complex(2)` work, which should be handled by type coercion. The call to `coerce` should not convert the argument into an Integer, but convert the current object into a class compatible with the argument. This means the combination of `arg = Natalie::coerce(...)` and `assert_type` won't work.

The code should more like this (I haven't tried this code yet, so please read it as pseudocode):
```c++
Value IntegerObject::add(Env *env, Value arg) {
    if (arg->is_float()) { // (1)
        return Value::floatingpoint(m_integer + arg->as_float()->to_double());
    } else if (!arg->is_integer()) {
        auto coerced = Natalie::coerce(env, arg, this);
        if (!coerced.first->is_integer()) // (2)
            return coerced.first.send(env, "+"_s, { coerced.second });
        arg = coerced.second;
    }
    arg->assert_type(env, Object::Type::Integer, "Integer");

    return create(m_integer + arg->as_integer()->integer());
}
```
I'm not sure if this still needs the `arg->is_float()` check `(1)`, this should probably be handled by the coercion as well.
I'm also not sure about the second `is_integer` check `(2)`, I don't think it should be here, but we might end up in infinite loops if we call this with an argument that has a coerce method that returns Integers.

A short Ruby script as an illustration:
```ruby
rat = Rational(3)
p rat.coerce(1)
p 1 + rat

puts ''
puts 'Update Rational#coerce to return Integers'
class Rational
  def coerce(other)
    [other, numerator]
  end
end
p rat.coerce(1)
p 1 + rat
```
Running this with MRI shows how the updated `coerce` method is used:
```
[(1/1), (3/1)]
(4/1)

Update Rational#coerce to return Integers
[1, 3]
4
```
But Natalie (with this patch) still returns a Rational:
```
[(1/1), (3/1)]
(4/1)

Update Rational#coerce to return Integers
[1, 3]
(4/1)
```

So I guess I should pretty much drop these changes, rewrite them to using `coerce` (and update the other operations like `IntegerObject::mul` as well). And there are probably many more locations where this pattern is present.

seven1m:
I'll trust you on this one. 😅 It's been awhile since I looked at the coerce stuff, so I think I've forgotten how that works for the most part.

herwinw:
> I'll trust you on this one. sweat_smile It's been awhile since I looked at the coerce stuff, so I think I've forgotten how that works for the most part.

Honestly, I don't think I ever understood it before I started working on this case :blush:

seven1m:
What do we need to do here? Can we go ahead and merge this?

herwinw:
I still feel like I've mixed some assignments up, I just need to take some time to go over all of it again. I'll try to do it somewhere this week.

herwinw:
It looks like the upstream specs for coercion with bitwise operators were broken too, I pushed an upstream fix in https://github.com/ruby/spec/pull/1011

TODO list for myself:
* Clean up `test/natalie/integer_spec.rb`, it has some duplication with upstream specs
* Add some tests for failure cases (like calling `Integer#&` with a mock that has a coerce result with objects not supporting `&`)

herwinw:
Well, that took way longer than it should have, but I think this is it.

PR #826, opened 2023-01-15
Fix several issues discovered while working on `Module#define_method`

richardboehme:
This fixes some issues that I discovered while working on `Module#define_method` (which I could not make spec compliant yet). I thought it might be useful to upstream those before continuing to work on `define_method`.


seven1m:
Thanks Richard!

PR #827, opened 2023-01-15
Fix Kernel.Integer with input with leading zeroes and explicit base

herwinw:
This should fix the issue described in  #820


seven1m:
Can we add a test with this?

herwinw:
I added some tests (and a bugfix). This branch now includes the commits of #830, to prevent merge conflicts in the new test file. So I would advise to merge the other case first.

PR #831, opened 2023-01-16
Array#pack q/Q and buffer: kwarg support

nanobowers:
This implements the 'q' and 'Q' directives for `Array#pack` and also implements the `buffer:` keyword arg which allows an existing buffer to be manipulated rather than starting from an empty new buffer and returning the new object.

Also, one extra `xit`ed NATFIXME was removed.

This should close issue #195 , as the percent_spec was already passing.


seven1m:
Cool!

PR #835, opened 2023-01-18
Implement String#capitalize and add capitalize_spec.rb

4hg:
One for #217. I deliberately left out `String#capitalize!` from the spec file. Let me know if there is anything that should be changed or things I should know for the future.


herwinw:
> I deliberately left out `String#capitalize!` from the spec file

Most times it's easier to start with the bang method, That way, you can just implement the normal method with `dup.capitalize!` (pseudo-Ruby).

After a quick glance at the code: this looks like it works fine with ASCII-strings, but I don't think we should be looking at separate bytes when using a multi-byte encoding (unicode etc.)
And looking back at #217: A lot of those methods depend on string encodings, which are only partially implemented in Natalie at the moment. I'm not really sure how much of it still matches the "first good issue"

src/string_object.cpp
```diff
@@ -2256,6 +2256,38 @@ Value StringObject::upcase(Env *env) {
...
+    if (length() == 0) {
+        return str;
+    } else {
+        auto ary = chars(env)->as_array();
+        auto first_char = ary->shift()->as_string();
+        if (first_char->bytesize() > 1) {
+            str->append(first_char);
+        } else {
+            auto c = first_char->c_str()[0];
+            if (c >= 97 && c <= 122) {
```
herwinw, line 2270: You can use 'a' and 'z' as char values as well, which makes it a bit more readable:
```c
if (c >= 'a' && c <= 'z') {
```

herwinw:
Second look: the lack for unicode support is pretty much covered in the code, so I guess does does improve Natalie

4hg:
> And looking back at #217: A lot of those methods depend on string encodings, which are only partially implemented in Natalie at the moment. I'm not really sure how much of it still matches the "first good issue". But that's an issue that is outside the scope of this change.

Yep, I am well aware of this. I simply wanted to include it for completeness sake since upcase and downcase are included (though I could not find the particular PR that included those).

src/string_object.cpp
```diff
@@ -2256,6 +2256,38 @@ Value StringObject::upcase(Env *env) {
     return str;
 }
 
+Value StringObject::capitalize(Env *env) {
+    auto str = new StringObject { "", m_encoding };
+    if (length() == 0) {
+        return str;
+    } else {
+        auto ary = chars(env)->as_array();
```
seven1m, line 2264: Splitting the string into chars is a very expensive operation. I wonder if there's something we can do with the bytes in-place. And I agree with @herwinw we should do the bang method first and use `dup.upcase!` for the non-bang method.

4hg: I'll poke around some more in the codebase in search of a different solution and get to work on a bang implementation. If you would like to close this in favor of a separate PR down the line, that's entirely fine with me. I'm not well-versed with git, so I don't know if this one can be modified.

seven1m: So I'm thinking we could call `next_char(0)` to just get the first character. That would be way faster.

Issue #836, opened 2023-01-18
Automatic validation of NATFIXME markers in specs


herwinw:
A lot of specs have a `NATFIXME` comment to describe what needs to be done before they can be enabled again. In many cases, it's a dependency on something external. As an example, in `spec/core/rational/to_r_spec.rb`:
```ruby
it "works when a BasicObject has to_r" do
  obj = BasicObject.new; def obj.to_r; 1 / 2.to_r end
  # NATFIXME: Implement Rational(String)
  # Rational(obj).should == Rational('1/2')
  Rational(obj).should == Rational(1, 2)
end
```
This spec tries to test the call to `to_r`, which works fine, but it tries to validate it using `Rational('1/2')` which has not yet been implemented. As a result, we had to update that statement to `Rational(1, 2)` to perform the same test.
Now when we add String support in `Kernel.Rational`, we would naturally update the spec `spec/shared/rational/Rational.rb`, but since the first one is pretty much unrelated to the actual change we just made, this will probably be forgotten.
Over the past few weeks, I feel like the majority of the Pull Requests I've made are just updates to the specs because we were skipping a lot of specs that were passing by now.

I think it would be better to add automated guards instead of comments, so we would have an option to find those issues easier. Something like this:
```ruby
it "works when a BasicObject has to_r" do
  obj = BasicObject.new; def obj.to_r; 1 / 2.to_r end
  NATFIXME(-> { Rational('1/2') == Rational(1, 2) }) do
    Rational(obj).should == Rational('1/2')
  end
  Rational(obj).should == Rational(1, 2)
end
```
I'm just recycling the name `NATFIXME` to get a uniform way of finding the FIXME markers in the specs. The method takes a callable argument and a block. If the callable argument returns false or raises an error, the block is skipped. Otherwise, the block is valid and we should probably raise an error to let the user know this spec needs an update.
This should work with including other blocks as well:
```ruby
NATFIXME(-> { Rational('1/2') == Rational(1, 2) }) do
  it "works when a BasicObject has to_r" do
    # Actual specs
  end
end
```
We should probably add an envvar to control these blocks, and don't run them by default, so we don't burden people with extra work when they fix something, and one of those NATFIXME blocks would reactivate and then fail again because there is another issue with the block. It's just a way to make it easier to find them.

Does anyone see any problems with this?

herwinw:
Since we're discussing this change in #837 now, I'm closing this issue. It's pretty much a duplicate now.

PR #837, opened 2023-01-18
NATFIXME guard in specs

herwinw:
This is a proof of concept to illustrate the proposal of #836 The Checks are going to fail

### DIFF
--- a/spec/core/rational/to_r_spec.rb
+++ b/spec/core/rational/to_r_spec.rb
@@ -11,8 +11,9 @@ describe "Rational#to_r" do
 
   it "works when a BasicObject has to_r" do
     obj = BasicObject.new; def obj.to_r; 1 / 2.to_r end
-    # NATFIXME: Implement Rational(String)
-    # Rational(obj).should == Rational('1/2')
+    NATFIXME 'Support Rational(String)' do
+      Rational(obj) == Rational('1/2')
+    end
     Rational(obj).should == Rational(1, 2)
   end
 
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -152,6 +152,17 @@ def specify(test = nil, &block)
   @specs << [$context.dup, test, block]
 end
 
+def NATFIXME(description)
+  remove_natfixme = begin
+                      yield
+                    rescue
+                      false
+                    end
+  if remove_natfixme
+    raise SpecFailedException, 'issue has been fixed, please remove or update the NATFIXME marker'
+  end
+end
+
 # NATFIXME: implement this guard
 def with_block_device
   nil


seven1m:
I really like this idea! But I wonder if we can make it a bit more ergonomic. What it's almost as simple as commenting out the broken code... but instead we wrap it in a NATFIXME block.

So something like this:

```ruby
Rational(obj).should == Rational('1/2')
```

...can be disabled like this:

```ruby
NATFIXME do
   Rational(obj).should == Rational('1/2')
end
```

Principally, this will disable the broken code. But it will also run the code each time to make **sure** it's still broken. If it no longer raises an error, then we should know about it.

What do you think?

herwinw:
@seven1m I've been thinking about that as well, but pretty often the specs have multiple reasons for failure. For example a small snippet from [`dash_e_spec`](https://github.com/ruby/spec/blob/master/command_line/dash_e_spec.rb#L20):
```ruby
it "uses '-e' in $0" do
  system(*ruby_exe, '-e', 'exit $0 == "-e"').should == true
end
```
A few weeks ago this failed because our version of `ruby_exe` required an argument to the file or the code to run. So we would have disabled the spec like this:
```
NATFIXME 'Support calls to ruby_exe with no arguments' do
  it "uses '-e' in $0" do
    system(*ruby_exe, '-e', 'exit $0 == "-e"').should == true
  end
end
```
That issue has now been fixed, `ruby_exe` with no arguments returns the path of the executable. This tests currently fails with "undefined method `system' for main", but since the NATFIXME check only sees that it's still failing, it won't raise any errors here. This means the code is still disabled, but the description on why it's disabled is now outdated.

nanobowers:
I suppose you could work it both ways: the arg to `NATFIXME` could either be a String or a Proc.  In theory I like the added potential safety of the Proc, but it seems like it might cause cascading NATFIXME updates trying to root-cause the new issue.  Also the String case looks a little cleaner than passing a lambda.  

I'd be in favor of some additional constraints:
1) not allowing NATFIXME without a block
2) requiring at least one arg (String or Proc) so that some reason is documented

Having either solution would be helpful to find cases where we can remove the NATFIXME, so looking forward to this.

herwinw:
I'm just going to start with a String as argument. If we ever want to support blocks, we can just check to see if if the argument is callable

seven1m:
I think we should do _something_ with this. I'm not opposed to going with your original idea, if you want to work on getting this mergable.

herwinw:
> I think we should do _something_ with this. I'm not opposed to going with your original idea, if you want to work on getting this mergable.

I've actually been considering closing this PR, and open up an issue with the Help wanted and Good first issue tags. It's probably not that hard, and it's all in Ruby, but you have to read up on the spec framework first before you can implement it.

herwinw:
I converted this one to the most basic implementation I could think of. It generates output like this:
```
Failed specs:
Rational#to_r
  works when a BasicObject has to_r
    issue has been fixed, please remove or update the NATFIXME marker
    (spec/core/rational/to_r_spec.rb:14:in `block in block in block')
```
Some limitations:
* It is not possible to use the spec matchers inside the block
* It is not possible to wrap `it` or `describe` blocks inside the `NATFIXME` block

nanobowers:
It would be nice if the spec matchers could work inside the block so that they wouldn't have to be edited/unedited each time the NATFIXME is added/removed.

If there isn't urgency to getting this added, i could take a look.  Maybe i'll find it's a dead end, but i'm willing to poke at it.

herwinw:
@nanobowers  I pretty much had three options:
1. Figure out how the spec implementation actually works and integrate the NATFIXME method into that
2. Create an issue with a detailed explanation as to what it's supposed to do and wait until somebody else implements it
3. Take the easy route and implement something that works a bit and can be improved later on

But since you've already started a much better implementation, I'm just going to close this one.

PR #844, opened 2023-01-21
String Upcase Downcase, including in_place

nanobowers:
Somewhat related to #835 - i had had this edit open for a while and I believe I improved the current implementation of upcase/downcase even though Unicode folding and turkic don't work.  Specs relating to Unicode casing have been "x"d out.

Some other notes:
+ I think this might improve performance over the original because i'm no longer using chars() to split into an ArrayObject, however I haven't done any benchmarking.
+ Some of the changes were made to support the UTF-16 related testcases.

I've tracked down the unicode document that has the case-folding here so may work on trying that part next: http://[www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt](http://www.unicode.org/Public/UCD/latest/ucd/CaseFolding.txt)

If anyone has thoughts on ways to map this efficiently, please let me know.  My best guess would be to Hash the codepoint mapping but only access it if the value was outside the normal ascii range to keep performance reasonable for the non-unicode case.


include/natalie/string_object.hpp
```diff
@@ -319,6 +337,8 @@ class StringObject : public Object {
 
     static size_t byte_index_to_char_index(ArrayObject *chars, size_t byte_index);
 
+    static CaseFoldType check_case_options(Env *env, Value arg1, Value arg2, CaseFoldType flags);
```
nanobowers, line 340: i think these same options may be used in `#casecmp` and `#capitalize`
#835

src/string_object.cpp
```diff
@@ -2220,42 +2220,97 @@ Value StringObject::rstrip_in_place(Env *env) {
...
+}
+
+Value StringObject::downcase(Env *env, Value arg1, Value arg2) {
+    CaseFoldType flags = check_case_options(env, arg1, arg2, Downcase);
     auto str = new StringObject { "", m_encoding };
-    for (auto c_val : *ary) {
-        auto c_str = c_val->as_string();
-        if (c_str->bytesize() > 1) {
-            str->append(c_str);
+    for (StringView c : *this) {
```
nanobowers, line 2263: Fewer lines than the older method, but maybe there was a reason for the old way.  Decoding/modifying and re-encoding codepoints feels more appropriate in the context of supporting multi-byte encodings like UTF-16/32, etc.

It feels like the case-mapping functions may eventually want to live with the encoding object itself?

PR #846, opened 2023-01-24
Implement Rational#truncate

herwinw:


seven1m (commented):
Do we need to check for divide-by-zero?

herwinw:
It turns out we do, but not in this change.

In IRB:
```
irb(main):001:0> Rational(1, 0)
(irb):1:in `Rational': divided by 0 (ZeroDivisionError)
```
But in Natalie:
```
nat> Rational(1, 0)
(1/0)
```
I'm kind of surprised that this is not covered in the specs.

And in this change, we might want to add an explicit check for the argument being zero, which now kind of works with the long way around.

herwinw:
I added https://github.com/ruby/spec/pull/1002 as an upstream PR to extend the specs of this method.

herwinw:
The upstream PR has been accepted, I included the updated spec in this PR. We're fully compliant with MRI.

PR #847, opened 2023-01-25
Fix behaviour when creating a Rational with zero as denominator

herwinw:
See #846


herwinw:
I will try to make a change for the upstream ruby-specs later today. If that gets merged, we can remove our extra test added in this change.

herwinw:
https://github.com/ruby/spec/pull/1003 for the upstream PR

herwinw:
This one has been updated with the upstream PR for the spec

PR #848, opened 2023-01-25
Time Improvements

nanobowers:
+ Improve `Time.now` to handle additional arguments.
+ Add `Time#isdst` (`Time#dst?`) and `Time#to_a`
+ Add `Time#zone` as a stub for use in `Time#to_a`.  Work still needed for legitimate TimeZone support
+ Add aliases for Time methods: `gm`, `mktime`, `gmt?`, `gmtoff` and `gmt_offset`, most of which are aliases for `utc*` methods.
+ Improve handling of arguments for `Time.new` and related methods:
    + Decoding of month abbreviation strings (e.g. `Jan`, `NOV`, `dec`) 
    + Argument coercion and bounds checking.

As a means of enhancing debug I also added this, which got wrapped up into this PR.
+    Add Spec formatter for test framework. (e.g. `bin/natalie some/test_spec.rb -f spec`)

The formatting is similar to what Mspec produces but is not the same.  *No Mspec code was contributed*

If you feel it should be broken out into a separate PR, please LMK.


src/time_object.cpp
```diff
@@ -244,6 +266,76 @@ Value TimeObject::year(Env *) const {
...
+
+nat_int_t TimeObject::normalize_month(Env *env, Value val) {
+    if (val->is_nil()) return 0;
+    if (!val->is_integer()) {
+        if (!val->is_string() && val->respond_to(env, "to_str"_s)) {
+            val = val->send(env, "to_str"_s)->as_string();
+        }
+        if (val->is_string()) {
+            auto monstr = val->as_string()->downcase(env, nullptr, nullptr)->as_string()->string();
+            if (isdigit(monstr[0])) {
```
seven1m, line 298: If we use `monstr->at(0)` we get bounds checking for free. Otherwise, this could be an out-of-bounds read? (Unless I'm missing a check somewhere.)

nanobowers: Oh that's a good point. I'm not sure any of the existing specs had the empty strings which we'd want to make sure to guard against.  I'll look into improving this here.

herwinw: This won't read out of bounds: if it's an empty string, this would read the null terminator, which is not a digit, which simply means this yields `false`.

nanobowers:
I made another update which changed the order of the String-Month-Abbreviation vs. String-Integer handling and completely removed the `monstr[0]` (or alternatively `monstr->at[0]`).  I arrived at the changes after manually "fuzzing" the inputs to the function in MRI to see what types of things it would accept rather than take the specs at face value.  For example the previous code would have not recognized `" 9"` as a month (since it naively checked the first character with `isdigit()` before), whereas MRI accepted this presumably because it was using `Integer()` to internally convert, based on an error message I received.  It's unclear to me if this means I should add things to the natalie test/ dir or see if ruby_spec is interested in more "trashy" input, or be okay the fact that there are lots of functions that try to interpret string input and the tests are anything but exhaustive.

PR #858, opened 2023-01-27
Implement UnboundMethod#bind_call

herwinw:
And import all passing specs for UnboundMethod


herwinw:
I honestly did not know this method existed until I noticed the failing specs for it.

Issue #840, opened 2023-01-19
Implement methods of `Set`


seven1m:
Our main approach is to work on one spec from [ruby/spec](https://github.com/ruby/spec/tree/master/library/set) at a time, implementing as much of the spec as possible, then check it off the list:

- [x] add_spec.rb
- [x] append_spec.rb
- [x] case_compare_spec.rb
- [x] case_equality_spec.rb
- [x] classify_spec.rb
- [x] clear_spec.rb
- [x] collect_spec.rb
- [x] compare_by_identity_spec.rb
- [x] comparison_spec.rb
- [x] constructor_spec.rb
- [x] delete_if_spec.rb
- [x] delete_spec.rb
- [x] difference_spec.rb
- [x] disjoint_spec.rb
- [x] divide_spec.rb
- [x] each_spec.rb
- [x] empty_spec.rb
- [x] eql_spec.rb
- [x] equal_value_spec.rb
- [x] exclusion_spec.rb
- [x] filter_spec.rb
- [x] flatten_merge_spec.rb
- [x] flatten_spec.rb
- [x] hash_spec.rb
- [x] include_spec.rb
- [x] initialize_clone_spec.rb
- [x] initialize_spec.rb
- [x] inspect_spec.rb
- [x] intersect_spec.rb
- [x] intersection_spec.rb
- [x] join_spec.rb
- [x] keep_if_spec.rb
- [x] length_spec.rb
- [x] map_spec.rb
- [x] member_spec.rb
- [x] merge_spec.rb
- [x] minus_spec.rb
- [x] plus_spec.rb
- [ ] pretty_print_cycle_spec.rb
- [x] pretty_print_spec.rb
- [x] proper_subset_spec.rb
- [x] proper_superset_spec.rb
- [x] reject_spec.rb
- [x] replace_spec.rb
- [x] select_spec.rb
- [x] size_spec.rb
- [x] subset_spec.rb
- [x] subtract_spec.rb
- [x] superset_spec.rb
- [x] to_a_spec.rb
- [x] to_s_spec.rb
- [x] union_spec.rb

nanobowers:
Hi @seven1m  
[...]

PR #862, opened 2023-01-28
Couple more `Set` methods

borisromanov:
#840

### DIFF
--- a/spec/library/set/shared/inspect.rb
+++ b/spec/library/set/shared/inspect.rb
@@ -1,4 +1,4 @@
-describe "set_inspect", shared: true do
+describe :set_inspect, shared: true do
   it "returns a String representation of self" do
     Set[].send(@method).should be_kind_of(String)
     Set[nil, false, true].send(@method).should be_kind_of(String)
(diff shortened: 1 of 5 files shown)


spec/library/set/shared/inspect.rb
```diff
@@ -7,9 +7,10 @@
...
   end
 
-  it "correctly handles self-references" do
-    (set = Set[]) << set
-    set.send(@method).should be_kind_of(String)
-    set.send(@method).should include("#<Set: {...}>")
-  end
+  # FIXME: Uncommenting this test causes VariableGetInstruction#generate to choke.
+  # it "correctly handles self-references" do
+  #   (set = Set[]) << set
```
herwinw, line 12: It's just the line that creates the set that causes the error here. Instead of uncommenting the whole test, it's probably better to rewrite it to something Natalie can understand
```ruby
it "correctly handles self-references" do
  # NATFIXME: Support this syntax
  # (set = Set[]) << set
  set = Set[]
  set << set
```
I'm not sure if this test is going to pass, the self reference is not handled in this code, it might be handled by the Array class, but I wouldn't be surprised when this would cause an infinite loop. That means we still would have to disable the test, but with a bit more specific message.

A few more generic pointers:
* To disable a spec, it's easier to replace `it` by `xit` (or `describe` by `xdescribe` when you're disabling a group of specs). This makes the difference between the upstream spec and our copy smaller, it's easier to search for disabled spec just by searching for those x-words, and we could add a test to check if they are still failing
* We use `NATFIXME` instead of `FIXME` to mark things that have to be fixed in Natalie. There are occasional FIXME markers in the upstream specs, and it's useful to distinguish between upstream and Natalie issues.

borisromanov: @herwinw, thank you for your detailed feedback. You were correct, even changing the self reference syntax still does cause the spec to fail; therefore, it's x'd out for now.

seven1m (approved):
Looks good!

PR #871, opened 2023-01-29
Time 20230126

nanobowers:
+   Add `in:` kwarg for `Time.new`, 
+    Fix `Time.new` to set UTC mode when in/zone provided.
+   Add timezone/utc-offset support for `Time.new`, including numeric, `+/-HH:MM(:SS)` and Military TZ support,
+    Fix second handling in `Time.new` .
+    Preserve utc_offset in result for `Time#+` and `Time#-`
+    Remove resolved NATFIXMEs gmt_offset specs because more of Time.new is functional.

### DIFF
--- a/src/time_object.cpp
+++ b/src/time_object.cpp
@@ -11,6 +11,15 @@ TimeObject *TimeObject::at(Env *env, Value time, Value subsec, Value unit) {
     return TimeObject::create(env, rational, Mode::Localtime);
 }
 
+TimeObject *TimeObject::at(Env *env, Value time, Value subsec, Value unit, Value in) {
+    auto result = at(env, time, subsec, unit);
+    if (in) {
+        result->m_time.tm_gmtoff = normalize_timezone(env, in);
+        result->m_time.tm_zone = (char*)"UTC";
+    }
+    return result;
+}
+
 TimeObject *TimeObject::local(Env *env, Value year, Value month, Value mday, Value hour, Value min, Value sec, Value usec) {
     TimeObject *result = new TimeObject {};
     result->build_time(env, year, month, mday, hour, min, sec);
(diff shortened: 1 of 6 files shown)


src/time_object.cpp
```diff
@@ -8,7 +8,16 @@ TimeObject *TimeObject::at(Env *env, Value time, Value subsec, Value unit) {
...
     }
-    return TimeObject::create(env, rational, Mode::Localtime);
+    return create(env, rational, Mode::Localtime);
+}
+
+TimeObject *TimeObject::at(Env *env, Value time, Value subsec, Value unit, Value in) {
+    auto result = at(env, time, subsec, unit);
+    if (in) {
+        result->m_time.tm_gmtoff = normalize_timezone(env, in);
+        result->m_time.tm_zone = (char *)"UTC";
```
seven1m, line 18: Does this `(char *)` cast need to be there? According to [this doc](https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html), the field takes a `const char *` anyway.

seven1m (approved):
Looks good, just have one question about the cast to remove const.

src/time_object.cpp
nanobowers: Ah yes, that recast was unnecessary. Removed it, reran the spec locally and am re-pushing through the CI.

seven1m: Weird, apparently it is needed on macos. Sorry! I wonder if we should strdup it then? But then we have to free it later? Gross.

nanobowers: I tested locally with Linux and didnt need it, didn't occur to me that macos would be different here, but I should know better at this point to take anything for granted across platforms.
I don't have enough C++ experience to know the pros and cons of the non-const re-cast, though I tend to avoid malloc/free whenever I can 😸

seven1m: I pushed a commit to hopefully fix this without a cast to remove const.

seven1m:
@nanobowers let me know what you think here. I'm not sure if the `m_zone` char* will conflict with something later on...

PR #872, opened 2023-01-29
Fixes for singleton classes and frozen properties

herwinw:
A fine example of yak shaving when I was looking into fixing the spec of [Singleton method on Numeric](https://github.com/ruby/spec/blob/master/core/numeric/singleton_method_added_spec.rb). I found some inconsistencies between objects being frozen with MRI. A few new Natalie tests have been added, I'm going to try to push those to the upstream specs.

### DIFF
--- a/include/natalie/complex_object.hpp
+++ b/include/natalie/complex_object.hpp
@@ -13,20 +13,28 @@ class ComplexObject : public Object {
 
 public:
     ComplexObject()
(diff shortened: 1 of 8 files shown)


herwinw:
Added two upstream PRs:
* https://github.com/ruby/spec/pull/1007
* https://github.com/ruby/spec/pull/1008

Issue #875, opened 2023-01-29
REPL does not use the -r statements


herwinw:
The set module has to be required explicitly. So this does fail:
```
bin/natalie -e 'p Set[1]'
-e:1:in `<main>': uninitialized constant Set (NameError)
```
But adding a `-r` switch does enable it
```
bin/natalie -rset -e 'p Set[1]'
#<Set: {[1]}
```
But if you start the repl with `-r`, the file is not loaded
```
bin/natalie -rset
nat> Set[1]
(repl):1:in `<main>': uninitialized constant Set (NameError)
```

PR #876, opened 2023-01-29
Add missing ">" to Stringification of Set

herwinw:
Just noticed this while I was writing  #875: The stringification methods of Set were missing the final `>`. Since the specs are pretty incomplete in this case (we're skipping the spec for the recursive case, so the only test left was just to see if the output was a String), this went unnoticed in the specs

### DIFF
--- a/lib/set.rb
+++ b/lib/set.rb
@@ -56,7 +56,7 @@ class Set
   end
 
   def inspect
-    "#<Set: {#{to_a}}"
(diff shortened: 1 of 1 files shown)


herwinw:
Just to keep things documented: I added https://github.com/ruby/spec/pull/1009 as an upstream PR to improve the specs for `Set#inspect` and `Set#to_s`.

PR #877, opened 2023-01-29
Enable passing `-e` multiple times

herwinw:
This did require a slight fix in the `ruby_exe` method as well, since this hit a case we didn't see before.


herwinw:
The change in `ruby_exe` should help a few other command line specs as well, with a quick search I can see `dash_c_spec`, `dash_s_spec` and `syntax_error_spec` failing because the code passed via `args:` is not executed.

PR #882, opened 2023-01-29
Pass -r arguments to Natalie to legacy REPL

herwinw:
This is the first step in fixing #875. The default/legacy REPL now works. The experimental REPL does receive the argument, but ignores it for now.

I might have a look at the v2-repl one day, but for now this resolves a few issues I had to work around.


seven1m:
The friend who started the new REPL hasn't been able to help out as much lately, and I wonder what the future of it will be... 😬

seven1m:
I started #888 btw. 😉

PR #883, opened 2023-01-29
Import basicobject_spec

herwinw:
This unearthed an error with include without parentheses. Support for the argument `inherit` in `Module#constants` has been added.

### DIFF
--- a/spec/core/basicobject/fixtures/common.rb
+++ b/spec/core/basicobject/fixtures/common.rb
@@ -1,0 +1,11 @@
+module BasicObjectSpecs
+  class BOSubclass < BasicObject
+    def self.kernel_defined?
+      defined?(Kernel)
+    end
+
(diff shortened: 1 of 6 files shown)


spec/core/basicobject/fixtures/common.rb
```diff
@@ -0,0 +1,11 @@
+module BasicObjectSpecs
+  class BOSubclass < BasicObject
+    def self.kernel_defined?
+      defined?(Kernel)
+    end
+
+    # NATFIXME: This results in "wrong number of arguments (given 0, expected 1+) (ArgumentError)"
+    # include ::Kernel
```
seven1m, line 8: This is most likely a parser bug

herwinw: This fixes it at well:
```ruby
include Kernel
```
So yes, parser error is very likely here.

seven1m: Fixed in natalie-lang/natalie_parser#53

PR #894, opened 2023-01-31
Fix String#include? for empty strings

timcraft:
Won't be fully spec compliant until we implement EUC-JP encoding, for now just fixing the simple empty string cases.

### DIFF
--- a/spec/core/string/include_spec.rb
+++ b/spec/core/string/include_spec.rb
@@ -13,6 +13,22 @@ describe "String#include? with String" do
     StringSpecs::MyString.new("hello").include?(StringSpecs::MyString.new("lo")).should == true
   end
 
+  it "returns true if both strings are empty" do
+    "".should.include?("")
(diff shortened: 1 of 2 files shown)


spec/core/string/include_spec.rb
```diff
@@ -13,6 +13,22 @@
     StringSpecs::MyString.new("hello").include?(StringSpecs::MyString.new("lo")).should == true
   end
 
+  it "returns true if both strings are empty" do
+    "".should.include?("")
+    # NATFIXME: Implement EUC-JP encoding
+    # "".force_encoding("EUC-JP").should.include?("")
+    # "".should.include?("".force_encoding("EUC-JP"))
+    # "".force_encoding("EUC-JP").should.include?("".force_encoding("EUC-JP"))
```
herwinw, line 21: We can still cover the intention of this spec by replacing EUC-JP with a different non-default encoding (e.g. one of the UTF16/UTF32 variants).

Issue #78, opened 2021-08-22
Implement respond_to_missing and method missing


TheGrizzlyDev:
These 2 methods allow a subclass of Object to dynamically provide methods when respond_to does not. It is a useful mechanism for dynamic delegation and is used throughout the core library, including Array which needs it for flatten and flatten! (among other methods) #75 . 

An example of these two methods can be found [here](https://ruby-doc.org/core-3.0.2/BasicObject.html). 
Links:
respond_to_missing (part of Kernel): 
- Docs: https://ruby-doc.org/core-3.0.2/Object.html#method-i-respond_to_missing-3F
- Spec: https://github.com/ruby/spec/blob/master/core/kernel/respond_to_missing_spec.rb

~method_missing (part of Object):~
- ~Docs: https://ruby-doc.org/core-3.0.2/Object.html#method-i-method_missing~
- ~Spec: https://github.com/ruby/spec/blob/master/core/basicobject/method_missing_spec.rb~

seven1m:
`method_missing` was implemented in d2a45f275b9948826787b69a1b2d98b812bdae0b :tada: 

We still need `respond_to_missing?` though...

PR #896, opened 2023-01-31
Implement Kernel#respond_to_missing?

timcraft:
For #78

### DIFF
--- a/include/natalie/object.hpp
+++ b/include/natalie/object.hpp
@@ -292,8 +292,9 @@ public:
 
     bool is_a(Env *, Value) const;
     bool respond_to(Env *, Value, bool = true);
-    bool respond_to_method(Env *, Value, Value) const;
-    bool respond_to_method(Env *, Value, bool) const;
+    bool respond_to_method(Env *, Value, Value);
+    bool respond_to_method(Env *, Value, bool);
+    bool respond_to_missing(Env *, Value, Value);
 
     const char *defined(Env *, SymbolObject *, bool);
     Value defined_obj(Env *, SymbolObject *, bool = false);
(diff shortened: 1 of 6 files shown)


include/natalie/object.hpp
```diff
@@ -292,8 +292,9 @@ class Object : public Cell {
 
     bool is_a(Env *, Value) const;
     bool respond_to(Env *, Value, bool = true);
-    bool respond_to_method(Env *, Value, Value) const;
-    bool respond_to_method(Env *, Value, bool) const;
+    bool respond_to_method(Env *, Value, Value);
+    bool respond_to_method(Env *, Value, bool);
+    bool respond_to_missing(Env *, Value, Value);
```
seven1m, line 297: I'm sad to lose the `const` here. Is there a way we can keep these methods const?

herwinw: I haven't really checked this code, but I remember having to remove const somewhere during a change because the code had to call another method that wasn't marked as const. It's pretty much turtles all the way down, which probably could have its own ticket

timcraft: Yep turtles... as I understand it we would have to at least make `Object::send` const?

herwinw: In this case we're calling `Object#send`, which it something that can modify the receiver, so it shouldn't be `const`, so this can't be `const` either.

timcraft: I think we could potentially avoid `Object::send` and use `Method::call` directly (we don't need to call #method_missing or raise NoMethodError in this context), but that's not `const` either.

herwinw: It looks possible to constify that one, but I guess we should consider that to be a future work. I think the added value of this change is bigger than the added value of adding some const correctness at more or less random code paths.

seven1m: OK, thanks for taking a look. This is fine for now.

PR #897, opened 2023-01-31
Fix crash with some incantations of send

herwinw:
And fix some keyword issues with the Args class as well.


herwinw:
FYI: I have the fixes, I just need to extract the commits and order them. I pushed early so I could add a link to the Github issue in the code comments.

seven1m:
Good find!

PR #907, opened 2023-02-01
Refactor of EncodingObject into subclasses

nanobowers:
+ Move functionality for `is_dummy`, `is_ascii_compatible`, and `in_encoding_codepoint_range` into the subclasses of
 `EncodingObject`.  There are a lot more encodings to implement and am trying to setup something more modular before adding those.
Remove calls to `EncodingObject::num()` where possible.  AFAICT there is one remaining external call to this function.  

Maybe removing this is a bad idea, but it feels like maintaining an enum list of `Encodings` would become fragile or harder to maintain when adding more encodings.  That said, it is convenient (and fast) and pretty pervasive, so I expect it would be non-trivial to get rid of it.


seven1m:
This is superseded by #928 I believe...

nanobowers:
Yes, superseded by #928 , closing.

PR #914, opened 2023-02-02
Implement SecureRandom methods

timcraft:
This implements `Random.urandom` using `arc4random_buf` and then uses that to implement a number of methods for the `SecureRandom` module (#bytes, #random_bytes, #base64, and #hex).


seven1m:
Nice! If it works on macos, it will _probably_ work on OpenBSD. I've been working off-and-on to get the [builds for OpenBSD](https://builds.sr.ht/~tim/natalie) running again.

PR #918, opened 2023-02-02
Implement Random#bytes

herwinw:
The tests had to be slightly altered because our Random implementation differs from MRI. This is an issue in the base object, and does not affect the code written in this change.

### DIFF
--- a/spec/core/random/bytes_spec.rb
+++ b/spec/core/random/bytes_spec.rb
@@ -1,0 +1,32 @@
+# -*- encoding: binary -*-
+require_relative '../../spec_helper'
+require_relative 'shared/bytes'
+
+describe "Random#bytes" do
+  it_behaves_like :random_bytes, :bytes, Random.new
+
+  it "returns the same output for a given seed" do
+    Random.new(33).bytes(2).should == Random.new(33).bytes(2)
+  end
+
+  # Should double check this is official spec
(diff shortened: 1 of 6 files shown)


spec/core/random/bytes_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  # Should double check this is official spec
+  # NATFIXME: Our algorithm differs from MRI, see e.g. `Random.new(0).rand(100)`
+  #           where MRI return 44 and Natalie returns 54. And MacOS has
+  #           different output compared to linux.
+  it "returns the same numeric output for a given seed across all implementations and platforms" do
+    rnd = Random.new(33)
+    # rnd.bytes(2).should == "\x14\\".b
+    if RUBY_PLATFORM =~ /darwin/i
+      rnd.bytes(2).should == "\x14\a".b
```
seven1m, line 21: I wonder if we would be better off just checking that the number of bytes get returned as expected vs the actual specific bytes.

As I said in #919, this feels like an implementation detail to me.

herwinw: I disabled these specs for now. The important part here is that the same seed generates the same out (that's a property of the PRNG/Mersenne Twister), but it has a test to check that:
```ruby
it "returns the same output for a given seed" do
  Random.new(33).bytes(2).should == Random.new(33).bytes(2)
end
```

PR #928, opened 2023-02-03
New Encodings

nanobowers:
This adds preliminary support for five new encodings:
* `IBM437`
* `IBM866`
* `ISO-8859-1`
* `Shift_JIS`
* `EUC_JP`

The first three are single-byte encodings and are similar to the existing ASCII-8bit encoding.
The latter two are variable-byte encodings and the `next_char` has been tested with a side script (not part of this commit) to ensure it works properly.  The `prev_char` functionality for these is unlikely to be accurately and is noted in the NATFIXME.

Per some simple analysis these encodings are the next five most frequently used in the specs after the existing Ascii/UTF variants, so I expect that adding these will improve some of the nightly runs.

Existing specs that were pending development of these encodings have been updated as appropriate.

Additionally, some of the functionality in the encodings.hpp (switch) has been moved into the encoding subclasses so that methods in those subclasses can respond to the messages by using overrides, Namely the `in_encoding_codepoint_range`, `is_ascii_compatible` and `is_dummy` methods.


src/encoding_object.cpp
```diff
@@ -70,6 +70,9 @@ Value EncodingObject::encode(Env *env, EncodingObject *orig_encoding, StringObje
     }
 
     str->set_str(temp_string.string().c_str(), temp_string.string().length());
+
```
nanobowers, line 73: I need to remove this comment...

src/encoding_object.cpp
```diff
@@ -24,7 +24,7 @@ Value EncodingObject::encode(Env *env, EncodingObject *orig_encoding, StringObje
         if (unicode_codepoint < 0) {
             StringObject *message;
 
-            if (num() != Encoding::UTF_8) {
+            if (name()->string() != "UTF-8") {
```
nanobowers, line 27: this should be reverted, this was done in an attempt to reduce/eliminate using `num()` (enum) but that effort was abandoned

src/encoding/eucjp_encoding_object.cpp
```diff
@@ -0,0 +1,82 @@
+#include "natalie.hpp"
+
+namespace Natalie {
+
+// NATFIXME: actual implementation is wrong for much of this file
```
seven1m, line 5: Is this comment still accurate, or just leftover while you were working on the file?

If it's still accurate, I hesitate to merge a known-wrong implementation. I worry we'll have subtle bugs that are hard to track down later.

If it's known to be wrong, perhaps we should put a `NAT_NOT_YET_IMPLEMENTED()` in the methods that are wrong... 🤔

nanobowers: The comment is still accurate: With the exception of `next_char`, I don't think any of these methods (e.g. `prev_char` or `*_codepoint` related methods) are tested by any specs we have (and possibly not by any specs in ruby's spec).  To that end, these may or may not work, but would venture that the Japanese encodings would give incorrect results for some inputs for some of the methods.  IIRC few if any of these are exposed as ruby methods so that makes it harder to test.

`NAT_NOT_YET_IMPLEMENTED()` is a great idea.  Looks like after adding to N.N.Y.I. the tests still pass which confirms my  suspicion that none of the tests hit that code.

If you'd rather - I can mark as draft, then work on fixing the rest of the methods (and add new tests) before trying to merge.

seven1m: Wait, I think @andrykonchin was adding some sort of test in the `test/` directory for prev_char(). I'm on mobile right now so it's hard to search. But I can check later!

seven1m: OK, I found it here: https://github.com/natalie-lang/natalie/blob/355deeec90ecf5dbd7ac29e8b934c3b76845354e/test/natalie/string_test.rb#L571

For this PR, I'm fine with us using `NAT_NOT_YET_IMPLEMENTED()`, and we can come back to work on those methods in a later PR.

Thanks!

PR #929, opened 2023-02-04
Implement MatchData#dup

herwinw:
That resolves another crashing state in the nightly spec result. I added the readers `MatchData#string` and `MatchData#regexp` as well, to fully make this spec pass


seven1m:
Yay, my GC Lint script caught something! 🎉

herwinw:
I tried to fix the GC issue, but that just resulted in other crashes. For now I just took the easy way out and removed the commit with `MatchData#regexp`.

PR #933, opened 2023-02-04
Fix #define_method with blocks converted from a Method

### LINKED ISSUE #718: Wrong behavior when defining `new` with another class initializer
The following code should return `String` but does return `MethodHolder`

```ruby
class MethodHolder
  class << self
    define_method(:new, &String.method(:new))
  end
end

p MethodHolder.new.class
```

richardboehme:
Blocks created from a method object (using #to_proc or the ampersand) do behave different than normal blocks passed to #define_method. When calling the newly defined method the "self" object should be the "self" object of the block rather than the current "self" object.

For example:

```ruby
class MethodHolder
  class << self
    define_method(:new, &String.method(:new))
  end
end

p MethodHolder.new.class
```

This should not invoke the "new"-Method on the MethodHolder class but on the String class.

Closes #718

### DIFF
--- a/include/natalie/block.hpp
+++ b/include/natalie/block.hpp
@@ -11,11 +11,18 @@ class Block : public Cell {
     friend ProcObject;
 
 public:
-    Block(Env *env, Value self, MethodFnPtr fn, int arity)
+    enum class BlockType {
+        Proc,
+        Lambda,
+        Method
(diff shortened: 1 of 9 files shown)


seven1m:
Yay!! 🎉

PR #934, opened 2023-02-05
Implement OpenSSL::Random.random_bytes

herwinw:
Which also means this adds a hard dependency on the openssl headers to be able to compile. It would probably be nice to just skip the extension if those can't be found.

I might want to look into implementing more of OpenSSL later on, since cryptography is pretty much a requirement to do anything these days.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,6 +46,7 @@ jobs:
       - name: install openssl
         run: brew install openssl
       - name: test with clang on macOS
-        env:
-          PKG_CONFIG_PATH: /opt/homebrew/opt/openssl@3/lib/pkgconfig
-        run: rake test
+        run: |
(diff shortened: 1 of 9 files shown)


herwinw:
This compiles on MacOS. The construction used to compile it feels like a hack, so I'm all ears if anyone has suggestions to improve it. Does `pkg-config` work on a Mac?
We're probably going to run into this with other libs as well.

seven1m:
🤔 Hrmm, this worked on my Mac. I'll look at it again tomorrow.

seven1m:
The shell out to homebrew makes too many assumptions. pkg-config should be able to get us the flags we need. I will look into it further.

seven1m:
OK I’m happy with this for now. We may need some instructions in the readme for Mac users wanting to use openssl; I can do that later.

PR #935, opened 2023-02-05
Implement Struct#dig

herwinw:
And improve `Struct#[]` to check the argument before calling a method.

### DIFF
--- a/src/struct.rb
+++ b/src/struct.rb
@@ -73,11 +73,13 @@ class Struct
         define_method :[] do |arg|
           case arg
           when Integer
-            attribute = attrs.fetch(arg)
-            send(attribute)
-          else
-            send(arg)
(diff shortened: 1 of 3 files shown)


src/struct.rb
```diff
@@ -73,10 +73,31 @@ def self.new(*attrs)
...
+            end
+          end
+          send(arg)
+        end
+
+        define_method :dig do |*args|
+          if args.empty?
+            raise ArgumentError, 'wrong number of arguments (given 0, expected 1+)'
+          end
+          arg = args.shift
```
herwinw, line 89: This looks very convoluted, but it seems like `define_method :dig do |arg, *args|` does not validate the size of the arguments, which means `arg` would be set to `nil` instead of raising an `ArgumentError`

PR #937, opened 2023-02-05
Improve Struct.new

herwinw:


herwinw:
https://bugs.ruby-lang.org/issues/19416
The last commit describes some curious behaviour in `Struct.new`, I added an upstream bug report for this.

herwinw:
I fixed the issue with `const_set` in #939, so I'm just going to add a dependency here. I would guess the change to `const_set` is more relevant

PR #938, opened 2023-02-05
Implement Process.groups

herwinw:
This is a dependency in [`FileTest.grpowned?`](https://github.com/ruby/spec/blob/738c8f6e64031d3a2083b585dd8ea6b813766f49/shared/file/grpowned.rb).

Ironically, this spec depends on `String#scan`


herwinw:
Does not seem to work on macos, and I have no idea what changes are required to fix that

PR #942, opened 2023-02-06
Add implementation of NATFIXME for review

nanobowers:
Based on ideas from @herwinw 👏 

This PR implements #837 with some additions:
- Able to wrap expectations/matchers inside an 'it' block without modification
- Adds a new `NatalieFixMeException` class.
- Adds optional `klass:` and `match:` keyword arguments so that the expected failure signature is captured. If the specs no longer fail or the failure is different (no longer same as the given klass / match), then it will raise `NatalieFixMeException`

Some tests were added, hopefully I haven't missed some critical use cases.  Feedback would be appreciated.

### DIFF
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -1448,37 +1448,38 @@ at_exit { run_specs }
 #   some_spec.should be "bad"
 # end
 #
-# 2) Hide any error matching a klass
-# NATFIXME "some description", klass: NoMethodError do
+# 2) Hide any error matching a exception
+# NATFIXME "some description", exception: NoMethodError do
 #   some_spec.foo.should be "also bad"
 # end
 #
-# 3) Hide any error matching a klass and with error-message match
-# NATFIXME "some description", klass: NoMethodError, match: /method foo undefined/ do
+# 3) Hide any error matching a exception with message
+# NATFIXME "some description", exception: NoMethodError, message: /method foo undefined/ do
 #   some_spec.foo.should be "also bad"
 # end
 #
-# If the code inside the block does not cause the error or the error it caues does not match the klass (or klass/match) then
-# NatalieFixMeException will be raised.
+# If the code inside the block does not cause the error or the error it causes does
+# not match the exception (or exception/message) then NatalieFixMeException will be raised.
 #
-def NATFIXME(description, klass: nil, match: nil)
+def NATFIXME(description, exception: nil, message: nil)
   raise SpecFailedException, "NATFIXME requires a block" unless block_given?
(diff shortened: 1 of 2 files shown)


test/support/spec.rb
```diff
@@ -1434,3 +1436,60 @@ def run_specs
...
+#
+# 3) Hide any error matching a klass and with error-message match
+# NATFIXME "some description", klass: NoMethodError, match: /method foo undefined/ do
+#   some_spec.foo.should be "also bad"
+# end
+#
+# If the code inside the block does not cause the error or the error it caues does not match the klass (or klass/match) then
+# NatalieFixMeException will be raised.
+#
+def NATFIXME(description, klass: nil, match: nil)
```
seven1m, line 1464: I think we should rename the `match` argument to `message`.

seven1m:
Very cool, friends! Ghanks to @herwinw for kicking this off with an idea.

test/support/spec.rb
herwinw: And maybe rename `klass` to `exception` as well? I think class is a bit too broad of a term.

test/support/spec.rb
```diff
@@ -1434,3 +1436,60 @@ def run_specs
...
+  raise SpecFailedException, "NATFIXME requires a block" unless block_given?
+  klass ||= StandardError
+  match = case match
+          when String then Regexp.new(match)
```
herwinw, line 1468: This does not escape the regex characters, so that it probably not what you want:
```
nat> match = Regexp.new('foo.')
/foo./
nat> match.match?('foox')
true
```
It is possible to use quoting:
```
nat> match = Regexp.new(Regexp.quote('foo.'))
/foo\./
nat> match.match?('foox')
false
nat> match.match?('foo.')
```

But I would personally just rewrite it to some lambdas:
```ruby
case match
when String then ->(message) { message.include?(match) }
when Regexp then ->(message) { match.match?(message) }
when nil then ->(_) { true }
else raise ArgumentError, "match must be nil, Regexp or String"
end
```
And of course update the caller too (which can't be nil at that line, so we can drop the first check too):
```ruby
if match.call(ex.message)
```

test/support/spec.rb
nanobowers: Updated both kwargs in the latest push

test/support/spec.rb
nanobowers: Awesome idea, I updated and pushed changes based on your example above.

seven1m (approved):
Nice!

PR #959, opened 2023-02-07
Convert NATFIXME comments to blocks in integer specs

herwinw:
A few have been left as comments, they either take too long or abort the program


herwinw:
#963

Issue #963, opened 2023-02-07
Convert NATFIXME comments in the specs into blocks


herwinw:
A lot of spec files in Natalie have `NATFIXME` comments to indicate something is not yet working. [An example](https://github.com/natalie-lang/natalie/blob/37b97c6675cfeddfa5f44f9f0b5b8a03c077b2b0/spec/core/rational/to_r_spec.rb):
```ruby
obj = BasicObject.new; def obj.to_r; 1 / 2.to_r end
# NATFIXME: Implement Rational(String)
# Rational(obj).should == Rational('1/2')
Rational(obj).should == Rational(1, 2)
```

These kind of comments can be replaced by [NATFIXME blocks](https://github.com/natalie-lang/natalie/blob/37b97c6675cfeddfa5f44f9f0b5b8a03c077b2b0/test/support/spec.rb), so it should read something like this
```ruby
obj = BasicObject.new; def obj.to_r; 1 / 2.to_r end
NATFIXME 'Implement Rational(String)', exception: ArgumentError do
  Rational(obj).should == Rational('1/2')
end
Rational(obj).should == Rational(1, 2)
```
This has the benefit of actually running the block and generating an error once the issue has been resolved, so we can update the spec to make it behave like the upstream specs again.

Not every block of disabled code can be converted, some code aborts the program, some code takes too long, that code should remain disabled.

You might find disabled tests that actually pass, or disabled tests with outdated comments. These can all be fixed.

# Directories with NATFIXME comments:

- [x] spec/command_line
- [x] spec/core/array
- [x] spec/core/basicobject
- [x] spec/core/class
- [x] spec/core/dir
- [x] spec/core/encoding
- [x] spec/core/enumerable
- [x] spec/core/enumerator
- [x] spec/core/enumerator/lazy
- [x] spec/core/file
- [x] spec/core/float
- [x] spec/core/hash
- [x] spec/core/integer
- [x] spec/core/kernel
- [x] spec/core/matchdata
- [x] spec/core/math
- [x] spec/core/method
- [x] spec/core/module
- [x] spec/core/process
- [x] spec/core/random
- [x] spec/core/range
- [x] spec/core/rational
- [x] spec/core/regexp
- [x] spec/core/string
- [x] spec/core/string/unpack
- [x] spec/core/struct
- [x] spec/core/symbol
- [x] spec/core/time
- [x] spec/language
- [x] spec/library/rbconfig
- [x] spec/library/set
- [x] spec/library/socket
[...]

PR #964, opened 2023-02-07
Implement MatchData#pre_match

timcraft:


spec/core/matchdata/pre_match_spec.rb
```diff
@@ -0,0 +1,27 @@
+require_relative '../../spec_helper'
+require_relative 'fixtures/classes'
+
+describe "MatchData#pre_match" do
+  # NATFIXME: Implement $` special global variable
+  xit "returns the string before the match, equiv. special var $`" do
+    /(.)(.)(\d+)(\d)/.match("THX1138: The Movie").pre_match.should == 'T'
+    $`.should == 'T'
+  end
```
herwinw, line 9: There is a block syntax for `NATFIXME` now, I think it's better to use that whenever possible (since it validates the failure and will croak whenever the issue is resovled):
```ruby
it "returns the string before the match, equiv. special var $`" do
  NATFIXME 'Implement $` special global variable', exception: SpecFailedException do
    /(.)(.)(\d+)(\d)/.match("THX1138: The Movie").pre_match.should == 'T'
    $`.should == 'T'
  end
end
```

See #942 and #963 for more information about this feature

timcraft: Sure, can we update the [code submission policy](https://github.com/natalie-lang/natalie/blob/master/CONTRIBUTING.md#code-submission-policy) if this is now a requirement?

seven1m: I'm not sure if it should be a _requirement_ yet. It's brand new--let's see how it works for a few days before requiring it. I also don't want to put undue burden on newcomers.

PR #974, opened 2023-02-07
Group tests in test/runner.rb

herwinw:
This has two advantages.
First of all, this groups the compilation step, so we only have to run the compiler one. This results in a small performance boost (running the tests of `spec/library/set` on my machine drop from around 1m35 to 1m20, so that's around 15% performance increase).
Second, this does continue the run if a single test is failing, and shows a readable overview at the end of the run.

Because Natalie does not support variables in `load` or `require`, we had to jump through a few hoops to combine all the arguments into a single file.

### DIFF
--- a/test/runner.rb
+++ b/test/runner.rb
@@ -4,6 +4,10 @@
 #
 #     bin/natalie test/runner.rb spec/core/array/pack/{a,b}_spec.rb
 #
+# to run all the tests grouped:
+#
+#     bin/natalie test/runner.rb --run-grouped spec/core/array/pack/{a,b}_spec.rb
+#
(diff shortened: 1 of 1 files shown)


seven1m:
We actually had something like this before, which I ended up [reverting](https://github.com/natalie-lang/natalie/commit/493f2284e332d5d16708751b6e12360a473bd7e5) due to odd behavior when specs are grouped together with Natalie.

seven1m:
Perhaps we could make it an optional flag?

herwinw:
It now supports a `--run-grouped` argument, the old behaviour is the default.

PR #976, opened 2023-02-08
Raise argumenterror when raising negative to fractional

herwinw:
This code path first returned Float::NaN, which made the specs fail, so these specs were disabled. Pull Request #969 updated some behaviour of Float::NaN, which means the test would now erroneously pass. At the same time, Pull Request #959 did enable the tests with a NATFIXME block. These two changes combined resulted in passing tests within a NATFIXME block.

For now, just raise an exception so it's clear this still has to be implemented, and make the specs pass again.


herwinw:
Currently  #975 shows an example of the error raised by running the full test suite

PR #984, opened 2023-02-09
Use explicit ownership in Args class

### LINKED ISSUE #898: Constructors of the Args class should use move or copy semantics
The `Args` class can be constructed with an existing list like structure like a `Vector` or an `ArrayObject`. I am going to use the `Vector` in this example, the `ArrayObject` works in a similar fashion, as do the copy constructors. The code is abbreviated to only show the relevant parts:
```c++
class Args {
    Args(TM::Vector<Value> vec)
        : m_size { vec.size() }
        , m_data { vec.data() } { }

    size_t m_size { 0 };
    const Value *m_data { nullptr };
};
```
In other words: when you construct an Args from a Vector, we don't copy the data itself, we just copy the data pointer and the size. Now we hold a pointer which we don't own, the Vector object is still the owner of the memory, and this might reallocate or free it. This might result in dangling pointers.

It would be better to use move semantics here: after you copied the data pointers, you reset the state of the Vector object without calling the destructor, so that object is no longer the owner of the memory. But since we can't access the private instance variables of the Vector, we cannot do that. A friend relation would become really messy, since that adds a dependency in TM to Natalie. We probably have to consume the Vector and save it in an instance variable of the Args.

herwinw:
Fixes #898

### DIFF
--- a/include/natalie/args.hpp
+++ b/include/natalie/args.hpp
@@ -43,12 +43,10 @@ public:
     Args(Args &&other)
         : m_size { other.m_size }
         , m_data { other.m_data }
-        , m_has_keyword_hash { other.m_has_keyword_hash }
-        , m_array { other.m_array } {
+        , m_has_keyword_hash { other.m_has_keyword_hash } {
         other.m_size = 0;
         other.m_data = nullptr;
         other.m_has_keyword_hash = false;
-        other.m_array = nullptr;
     }
 
     Args &operator=(const Args &other);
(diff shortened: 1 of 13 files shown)


herwinw:
Okay, I am never going to understand clang-tidy. This code had some issues with memory allocations, but those issues were in master as well.
I rebased this branch and now it starts with a rewrite and some bugfixes of the `spawn` method.

PR #995, opened 2023-02-11
File.new, File#size

nanobowers:
Add optional 3rd argument to `File.new` (perm argument)
Enhance `File.new` to pass additional specs.
Add `File#size` instance method
Add check for `is_closed()` to some instance methods (atime, ctime, mtime, size) to prevent attempted stat on a closed file-descriptor

### DIFF
--- a/spec/core/file/new_spec.rb
+++ b/spec/core/file/new_spec.rb
@@ -1,0 +1,167 @@
+require_relative '../../spec_helper'
+require_relative 'shared/open'
+
+describe "File.new" do
+  before :each do
+    @file = tmp('test.txt')
+    @fh = nil
+    @flags = File::CREAT | File::TRUNC | File::WRONLY
+    touch @file
+  end
+
+  after :each do
+    @fh.close if @fh
+    rm_r @file
+  end
+
+  it "returns a new File with mode string" do
+    @fh = File.new(@file, 'w')
+    @fh.should be_kind_of(File)
+    File.should.exist?(@file)
+  end
+
+  it "returns a new File with mode num" do
+    @fh = File.new(@file, @flags)
+    @fh.should be_kind_of(File)
+    File.should.exist?(@file)
+  end
+
+  it "returns a new File with modus num and permissions" do
+    rm_r @file
+    File.umask(0011)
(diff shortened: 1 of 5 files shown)


spec/core/file/new_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    rm_r @file
+    begin
+      f = File.new(@file, "w", 0444)
+      -> { f.puts("test") }.should_not raise_error(IOError)
+    ensure
+      f.close
+    end
+    File.should.exist?(@file)
+    File.read(@file).should == "test\n"
```
seven1m, line 54: I paused the spec here and examined the file:

```sh
→ ls -lh /Users/tim/pp/natalie/rubyspec_temp/96092/1-test.txt
----------  1 tim  staff     5B Feb 11 08:08 /Users/tim/pp/natalie/rubyspec_temp/96092/1-test.txt
```

It is as if the permissions were all set to 000 rather than 0444.

But if I run `chmod 0444` manually in the terminal, it works:

```sh
→ chmod 0444 /Users/tim/pp/natalie/rubyspec_temp/96092/1-test.txt
→ ls -lh /Users/tim/pp/natalie/rubyspec_temp/96092/1-test.txt
-r--r--r--  1 tim  staff     5B Feb 11 08:08 /Users/tim/pp/natalie/rubyspec_temp/96092/1-test.txt
```

seven1m: I also tried other numbers such as:

```ruby
f = File.new(@file, "w", 0755)
f = File.new(@file, "w", 600)
f = File.new(@file, "w", 0600)
f = File.new(@file, "w", 292) # octal 0444
```

...and all produced the permission denied error.

The only thing that I found not to produce the error was:

```ruby
f = File.new(@file, "w")
```

seven1m: OK I think I figured it out in e6d71c30f4b13e75ed49f2e75207ee739fc74c4c

nanobowers:
@seven1m Sorry about this, I was moving code around and apparently became confused about the real source of the error (relating to umask).  Thanks much for figuring out the bug I introduced.

PR #1000, opened 2023-02-11
Include OpenSSL::VERSION constant

herwinw:
Which is good enough to make the config/freeze_spec pass, since that once doesn't do anything on OpenSSL 3.0 and above.

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -6,6 +6,18 @@
 using namespace Natalie;
 
 Value init(Env *env, Value self) {
+    ModuleObject *OpenSSL = new ModuleObject { "OpenSSL" };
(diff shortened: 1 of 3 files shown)


seven1m:
I believe the OpenSSL version running in CI on macos is 1.1 😅

herwinw:
> I believe the OpenSSL version running in CI on macos is 1.1 sweat_smile

Okay, I'll just fix that by hardcoding 3.0.0 in the result :stuck_out_tongue_closed_eyes: 

I'll keep an eye on the CI, it might take a few attempts (and a few hours waiting for the results) to make it pass on MacOS.

herwinw:
Wow, it looks like everything breaks except for my machine

PR #1003, opened 2023-02-11
Convert NATFIXME comments in Regexp to blocks

herwinw:
#963


herwinw:
It looks the tests in `match_spec` behave differently when running the single spec compared to the full test suite. I resorted to `xit`-ing those tests, since I couldn't find a way to get the behaviour of both into these statements.

PR #1013, opened 2023-02-12
Improvements for Dir / DirObject

nanobowers:
+ Implement Dir as an instantiatable object inheriting from Object.
+ Add `open` and `chroot` class methods.
+ Add `initialize`, `fileno`, `close`, `path`, `pos`, `read`, `rewind`, `seek`, `tell`, `to_path`, and `inspect` instance methods.
+ Fix crash in `change_current_path` C++ method when current-working-directory did not exist.

Marking as draft since I have a few questions.


seven1m:
Nice!

Issue #869, opened 2023-01-28
Implement format strings (%)


seven1m:
While working with ChunkyPNG, I got the error:

```
lib/chunky_png/color.rb:573:in `to_hex': undefined method `%' for "#%08x":String
```

The code looks like this:

```ruby
def to_hex(color, include_alpha = true)                                                                                                                                                       
  include_alpha ? ("#%08x" % color) : ("#%06x" % [color >> 8])                                                                                                                                
end
```

herwinw:
For completeness: This is `string/module_spec` from #217

herwinw:
Since we've almost got the second version of this feature, I guess this case can be closed

seven1m:
The first version was completed before I abandoned it for a rewrite. 😆 

I think we should keep this issue until all the specs are passing?

seven1m:
[note: I deleted an earlier comment that was unclear]

This feature isn't complete. I was trying to work on the specs and realized I hated working on this code because it was so hard to follow. (My fault, given the way I set it up to start.)

I think the new code is much easier to hack on and should allow us to complete this issue soon-ish. Let's leave it open until all those `string/modulo_specs.rb` are passing.

seven1m:
There's a few specs still marked with NATFIXME, but they are related to encoding stuff I don't care to work on right now. 😅 

All the core functionality of Kernel#sprintf (aka String#%) should be working now.

PR #1017, opened 2023-02-16
Start work on String#%

seven1m:
[#217] [#869]

We have to start somewhere!

I found it easier at this stage to use `xit` et al to disable broken specs. We have NATFIXME now for marking known-broken tests, but when it's hundreds of broken tests at the beginning of a project, it's easier to `%s/it/xit` 😄

### DIFF
--- a/spec/command_line/dash_e_spec.rb
+++ b/spec/command_line/dash_e_spec.rb
@@ -31,14 +31,14 @@ describe "The -e command line option" do
     end
 
     it "mimics an awk conditional by comparing an inclusive-end range with $." do
-      NATFIXME 'Support -n', exception: NoMethodError, message: /undefined method `%'/ do
+      NATFIXME 'Support -n', exception: SpecFailedException do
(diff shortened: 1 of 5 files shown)


seven1m:
So, this makes me wonder... could we rename `NATFIXME` to `xit` so that it's a bit more natural to use... You can `%s/it/xit` with ease, and we'll know when tests start passing? Just a though. 🤔

nanobowers:
I think there are some tests in ruby/spec that are xit'd (or it statements that dont have a block which are treated as an xit) so maybe that would run afoul of renaming NATFIXME to xit ?  The way the NATFIXME is written right now is more of a wrapper around the expectation than a replacment for `it`.

Altenatively we could wrap `it`+`NATFIXME` as `nxit` or `fixit`, this more naturally covers the case where you want to skip the whole test.

seven1m:
>I think there are some tests in ruby/spec that are xit'd

Yeah, good point. OK, well what we have is fine for now.

PR #1022, opened 2023-02-17
Handle ... argument forwarding syntax

### LINKED ISSUE #1021: Forward args support
Some specs use forward-args syntax `def foo(...)` and this is currently unsupported by the Natalie compiler.

I think nearly all of the `IO` class specs hit this.

seven1m:
Fixes #1021


nanobowers:
I poked at trying this several days ago between when i filed the issue and when you made this and let's just say your implementation looks way better :smiley:

PR #1025, opened 2023-02-20
Fix some spec after/after-all ordering issues

### LINKED ISSUE #1020: Mini-mspec after-each/after-all sequencing
Our mini-mspec has problems with where the after-each/after-all is executed relative to the context.

From what I can tell, this causes quite a few false failures on specs because the after block gets executed at the wrong time.
In many cases either the object is prematurely destroyed and subsequent blocks have issues, or the object should have been destroyed and was not, also causing issues.  The File/Dir specs make heavy use of creation/deletion inside of before/after so I've seen this a bunch.

Additionally, I think nesting contexts may also interact with the sequencing of before/afters

I'll try to come up with some testcases, but wanted to document it while I was thinking about it, I suspect it would clean up a lot false failures and crashes in the nightly tests.

seven1m:
This an attempt to fix #1020.

To compare, I downloaded [mspec](https://github.com/ruby/mspec) and ran the following command to compare the outputs:

```sh
diff -y <(../mspec/bin/mspec --format spec test/natalie/test_order_example.rb | grep -o "\[\[.*\]\]") <(bin/natalie -r./test/support/spec.rb test/natalie/test_order_example.rb | grep -o "\[\[.*\]\]")
```

### DIFF
--- a/test/natalie/test_order_example.rb
+++ b/test/natalie/test_order_example.rb
@@ -1,0 +1,51 @@
+describe 'before/after' do
+  before :all do
+    puts '[[before all]]'
+  end
+
+  before :each do
+    puts '[[before each]]'
+  end
+
+  it 'it 1' do
+    puts '[[it 1]]'
+  end
+
+  it 'it 2' do
+    puts '[[it 2]]'
+  end
+
+  context 'context 1' do
+    before :all do
+      puts '[[before all 1]]'
+    end
+
+    before :each do
+      puts '[[before each 1]]'
+    end
+
+    it 'it 1' do
+      puts '[[it 1]]'
+    end
+
+    it 'it 2' do
+      puts '[[it 2]]'
+    end
+
+    after :each do
+      puts '[[after each 1]]'
+    end
+
+    after :all do
+      puts '[[after all 1]]'
+    end
+  end
+
+  after :each do
+    puts '[[after each]]'
+  end
+
+  after :all do
+    puts '[[after all]]'
+  end
+end
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -1411,7 +1411,7 @@ def run_specs
         @formatter.print_success(context, test)
       ensure
         # ensure that the after-each is executed
-        context.each { |con| con.after_each.each { |a| a.call } }
+        context.reverse_each { |con| con.after_each.reverse_each { |a| a.call } }
       end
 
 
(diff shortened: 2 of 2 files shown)


seven1m:
Current diff:

```
[[before all]]                                                  [[before all]]
[[before each]]                                                 [[before each]]
[[it 1]]                                                        [[it 1]]
[[after each]]                                                  [[after each]]
[[before each]]                                                 [[before each]]
[[it 2]]                                                        [[it 2]]
[[after each]]                                                  [[after each]]
[[after all]]                                                 <
[[before all]]                                                <
[[before all 1]]                                                [[before all 1]]
[[before each]]                                                 [[before each]]
[[before each 1]]                                               [[before each 1]]
[[it 1]]                                                        [[it 1]]
[[after each 1]]                                                [[after each 1]]
[[after each]]                                                  [[after each]]
[[before each]]                                                 [[before each]]
[[before each 1]]                                               [[before each 1]]
[[it 2]]                                                        [[it 2]]
[[after each 1]]                                                [[after each 1]]
[[after each]]                                                  [[after each]]
[[after all 1]]                                                 [[after all 1]]
[[after all]]                                                   [[after all]]
```

seven1m:
Just needs a bit more work!

seven1m:
Success! 🎉 

```
[[before all]]                                                  [[before all]]
[[before each]]                                                 [[before each]]
[[it 1]]                                                        [[it 1]]
[[after each]]                                                  [[after each]]
[[before each]]                                                 [[before each]]
[[it 2]]                                                        [[it 2]]
[[after each]]                                                  [[after each]]
[[after all]]                                                   [[after all]]
[[before all]]                                                  [[before all]]
[[before all 1]]                                                [[before all 1]]
[[before each]]                                                 [[before each]]
[[before each 1]]                                               [[before each 1]]
[[it 1]]                                                        [[it 1]]
[[after each 1]]                                                [[after each 1]]
[[after each]]                                                  [[after each]]
[[before each]]                                                 [[before each]]
[[before each 1]]                                               [[before each 1]]
[[it 2]]                                                        [[it 2]]
[[after each 1]]                                                [[after each 1]]
[[after each]]                                                  [[after each]]
[[after all 1]]                                                 [[after all 1]]
[[after all]]                                                   [[after all]]
```

nanobowers:
Woot!   This is going to make a huge difference in the IO/File/Dir tests, thanks so much for sorting it all out.

PR #1030, opened 2023-02-22
Fix order of instructions when assignment is used

### LINKED ISSUE #873: Evaluation order of receiver and arguments is backwards
One of the leftovers of #862, where this code caused issues:
```ruby
(set = Set[]) << set
```
This raises an error (backtrace from commit 490d07224b66a98db6fb297ab475204a73ceb54a):
```
lib/natalie/compiler/instructions/variable_get_instruction.rb:48:in `generate': variable not defined set (RuntimeError)
	from lib/natalie/compiler/backends/cpp_backend/transform.rb:31:in `block in transform'
	from lib/natalie/compiler/instruction_manager.rb:27:in `walk'
	from lib/natalie/compiler/backends/cpp_backend/transform.rb:29:in `transform'
	from lib/natalie/compiler/backends/cpp_backend.rb:33:in `transform_instructions'
	from lib/natalie/compiler/backends/cpp_backend.rb:18:in `generate'
	from lib/natalie/compiler.rb:224:in `transform'
	from lib/natalie/compiler.rb:143:in `instructions'
	from lib/natalie/compiler.rb:102:in `write_file'
	from lib/natalie/compiler.rb:81:in `compile'
	from bin/natalie:139:in `run'
	from bin/natalie:305:in `<main>'
```
A few more ways to trigger this error (which all work in MRI)
```ruby
(a = 1) + a
(a = "") +  a
```
The following comparable statements have the same behaviour in MRI and Natalie:
```ruby
a = a # a => nil
a = "" + a # raises TypeError with message "no implicit conversion from nil to string"
a = "foo" + a.to_s # a => "foo", because a is nil before the assignment
```
I think the last one is the most interesting of the bunch, since adding parentheses causes Natalie to break again, which probably is the minimal change to trigger this error
```ruby
(a = "foo") + a.to_s
```
If the variable exists, the statement works, and the operations are done in correct order
```ruby
a = nil
(a = 1) + a # a => 1, return value => 2
```

seven1m:
Fixes #873

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -242,7 +242,16 @@ module Natalie
         if receiver.nil?
           instructions << PushSelfInstruction.new
         else
-          instructions << transform_expression(receiver, used: true)
+          receiver_instructions = transform_expression(receiver, used: true)
+          case receiver.sexp_type
+          when :lasgn, :iasgn, :gasgn
+            # If the receiver contains an assignment, we need it to run first
+            # e.g. (a = 1) + a
+            instructions.unshift(receiver_instructions)
+            instructions << MoveRelInstruction.new(args.size + 1)
+          else
+            instructions << receiver_instructions
+          end
         end
 
         instructions << SendInstruction.new(
--- a/test/natalie/assign_order_test.rb
+++ b/test/natalie/assign_order_test.rb
@@ -1,0 +1,9 @@
+require_relative '../spec_helper'
+
+describe 'assignment order' do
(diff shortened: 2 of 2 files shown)


seven1m:
No, this isn't a complete fix. The following code is still executing out of order:

```ruby
describe 'call order' do
  @order = []

  def first
    @order << 1
    1
  end

  def second
    @order << 2
    2
  end

  it 'evaluates the receiver before the arguments' do
    (first + second).should == 3
    @order.should == [1, 2]
  end
end
```

I'm going to explore the idea of rearranging the inputs to SendInstruction...

Issue #1019, opened 2023-02-17
Behavior of begin/end/rescue/ensure


nanobowers:
Natalie's return value for `begin/end/rescue/ensure` appears to be different than MRI.
This causes trouble in some specs, particularly `spec/core/dir/fileno_spec.rb`

An example of this is:
```ruby
has_dir_fileno = begin
  dir = Dir.new('.')
  dir.fileno
  true
rescue NotImplementedError
  false
rescue Exception
  true
ensure
  dir.close
end
p has_dir_fileno
```
For this code, MRI returns `true` and Natalie returns `nil`.

MRI is returning the value from the begin-block (since the begin-block does not raise) whereas Natalie is returning the return value of the `ensure` code block.

PR #1031, opened 2023-02-23
Improvements mostly to Dir but also File and IO

nanobowers:
+ Add emacs scratch-files and test temp-dir to gitignore, mostly so that I don't accidentally checkin a scratchfile.
+ Add `Dir#pos=`
+ Move `Dir.foreach` and `Dir.each_child` into C++ 
+ Change return-type of `Dir#fileno` to int 
+ Add `IO#to_i` alias of `IO#fileno`
+ Change functions that always return a zero to return `int` instead of `Value` (`File.link`, `symlink`, `mkfifo`) 
+ Add `File.path`,  `File.rename`, `File.truncate` and `File#truncate`
+ Fix `Dir.open` close-ensure issue using `Defer` trick.  👏 
+ Restore `helpers/fs.rb` to upstream as #742 is fixed.
+ Import `Dir.fileno` test as #1019 is fixed.
+ Import passing specs for `Dir children`, `each_child`, `each`, `entries`, and `foreach` methods.  Some of those specs were `NATFIXME`'d due to lack of `EUC_KR` Encoding.

IIRC the remaining big to-do in `Dir` is `Dir.glob`, which looks like it'll be a big pain. 😦


seven1m:
Sweet

PR #1038, opened 2023-02-27
Make Kernel#method and Method#call spec compliant

timcraft:
Another one for #78

If #respond_to_missing?(name) returns true then Kernel#method returns a Method object with that name, but which calls #method_missing. For example:

    $ cat example.rb
    class Example
      def respond_to_missing?(method, bool)
        method == :foo
      end

      def method_missing(method, *args)
        method == :foo ? args : super
      end
    end

    instance = Example.new

    method = instance.method(:foo)

    p method

    p method.call(1, 2, 3)

    $ ruby example.rb
    #<Method: Example#foo(*)>
    [1, 2, 3]


seven1m:
Cool!

seven1m:
Cool!

PR #1039, opened 2023-02-28
String updates

nanobowers:
+ Fix `String#force_encoding` to pass specs
+ Fix `String#initialize` to pass specs
+ Add String methods `#casecmp`, `#casecmp?` and specs.  Worth noting that these may not be the most efficient implementations for non-worst case scenarios.
+ Add String methods `#capitalize` and `#capitalize!` and specs.  
+ Add String methods `#swapcase` and `#swapcase!` and specs.  

The above two were patterned off of upcase/downcase.  Similar to upcase/downcase, they only work for the ascii ranges, we would need to improve the `EncodingObject`s to handle non-ASCII case-swapping.


nanobowers:
Fwiw, I had intended to leverage the abandoned #835 but it seems like the related code in that PR had since been removed and I found it pretty straightforward to leverage the code in the existing upcase/downcase methods.

nanobowers:
Ho hum. Appears that in refactoring some code, it has exposed otherwise identical MRI error messages differ in their use of (or lack of) double-quotes. 😞

PR #1044, opened 2023-02-28
Fix REPL input on macOS

timcraft:
When I type `1+1` into the REPL I get messed up output like this:

    $ bin/natalie
    nat>     nat> 1      nat> 1+       nat> 1+1

Using these different escape codes for save_cursor, reset_cursor, and clear fixes the issue.

Please could someone test these work okay on BSD and Linux...? I've taken the codes from the tty-cursor gem which I'm pretty sure is platform independent (it uses different codes for Windows, but we're not supporting Windows here currently)

### DIFF
--- a/lib/natalie/repl/repl.rb
+++ b/lib/natalie/repl/repl.rb
@@ -55,11 +55,15 @@ module Natalie
     end
 
     def save_cursor
-      echo "\u001B[s"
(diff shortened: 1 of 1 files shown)


seven1m:
This works fine on Linux, so I'm merging it.

PR #1047, opened 2023-03-05
Symbol class updates

nanobowers:
+ Fixed `Symbol#start_with?` to take multiple args and added `Symbol#end_with?`
+ Fixed `Symbol#[]` to take one/two args and added `Symbol#slice` which is an alias for `Symbol#[]`
+ Added `Symbol#downcase/upcase/swapcase/casecmp/casecmp?/capitalize`
+ Fixed `Symbol#inspect` to not quote `$` vars.  There is still a ways to go in getting `Symbol#inspect` to be spec compliant and the method that MRI uses seems to have lots of special cases for knowing when to quote.


src/symbol_object.cpp
```diff
@@ -34,6 +34,8 @@ StringObject *SymbolObject::inspect(Env *env) {
         if (c < 33 || c > 126) // FIXME: probably can be a smaller range
             quote = true;
     }
+    if (m_name.length() > 1 && m_name[0] == '$')
```
nanobowers, line 37: Special case where symbols that start with `$` but are not exactly `:$` should get quotes.  This cleans up many failing specs for `#inspect`

src/symbol_object.cpp
```diff
@@ -101,9 +145,10 @@ Value SymbolObject::name(Env *env) const {
     }
     return symbol->m_string;
 }
-
-Value SymbolObject::ref(Env *env, Value index_obj) {
-    return to_s(env)->send(env, intern("[]"), { index_obj });
+Value SymbolObject::ref(Env *env, Value index_obj, Value length_obj) {
+    // The next line worked in nearly every case, except it did not set `$~`
+    //return to_s(env)->send(env, intern("[]"), { index_obj, length_obj });
```
nanobowers, line 150: I poked around for a while but never could figure out why `$~` (last match) did not get set by sending the `[]` method, but directly using `to_s(env)->ref(...)` worked okay.

seven1m: I'm not sure, but we can revisit later.

PR #1052, opened 2023-03-06
Etc module

nanobowers:
Adds the `Etc` library module and several methods / constants.

### DIFF
--- a/include/natalie/process_module.hpp
+++ b/include/natalie/process_module.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <grp.h>
 #include <pwd.h>
 #include <sys/resource.h>
 #include <sys/types.h>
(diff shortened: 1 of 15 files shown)


nanobowers:
I think the CI failures are caused by a sequencing issue in the generated .cpp, whereby the function is referenced (in the .rb binding to a cpp function) before it is declared (in an included chunk of cpp code)

I believe I have see this sort of issue occur (occasionally) in some tests where `socket` was included.  It's unclear to me what the exact cause for this failure is, but in this case the `require "etc"` happens midway through the test. 

And also for macOS, there's a different branch used for one test which calls `system()`, which we havent implemented yet, so that needs fixing.

seven1m:
Here is an example that triggers the `a function-definition is not allowed here before ‘{’` error:

```ruby
# wat.rb

def foo
  require 'socket'
  p Socket
end

foo
```

nanobowers:
Once #1054 is merged, i'll rebase the commits from this PR and fix the `system()` call issue, then re-push

nanobowers:
Foiled by macOS again :disappointed: .  Will need some help to debug it from someone w/ access to that hardware/OS.  Or else i can `xit` the tests.

seven1m:
OK I think I fixed it. We'll show what CI says. 😄

nanobowers:
Thanks @seven1m for the help on this!

PR #1056, opened 2023-03-09
Use [[maybe_unused]] in place of (void)

stefanos82:
For unused variables, it preferred to declare them as `[[maybe_unused]]`.


seven1m:
Thanks!

PR #1060, opened 2023-03-11
File.chown and Process methods

nanobowers:
Add the following methods to `Process` module:
+ `getpgid`
+ `getpgrp`
+ `getpriority`
+ `getrlimit`
+ `getsid`
+ `setpgrp`

Add `File.chown` and `File#chown`

As part of this PR, I added a few additional specs that passed in spec/core/file and spec/core/process, even in cases where they were unrelated to functional improvements.


include/natalie/process_module.hpp
```diff
@@ -96,6 +144,66 @@ class ProcessModule : public Object {
...
+        if (resstr == "RLIMIT_NOFILE") { return RLIMIT_NOFILE ; }
+        //if (resstr == "NPTS") { return RLIMIT_NPTS; }
+        if (resstr == "RSS") { return RLIMIT_RSS; }
+        if (resstr == "RLIMIT_RTPRIO") { return RLIMIT_RTPRIO ; }
+        //if (resstr == "RLIMIT_RTTIME") { return RLIMIT_RTTIME ; }
+        //if (resstr == "RLIMIT_SBSIZE") { return RLIMIT_SBSIZE ; }
+        if (resstr == "RLIMIT_SIGPENDING") { return RLIMIT_SIGPENDING ; }
+        if (resstr == "STACK") { return RLIMIT_STACK; }
+        env->raise("ArgumentError", "invalid resource {}", resstr);
+        */
```
seven1m, line 206: What's the story with this commented out code? Is this something we will want for later, or just leftover during debugging? If you want to leave it in, please add a note about what it will be used for.

nanobowers: Oops, it should've been removed and was an earlier version before I thought of a better way. I need some sort of notifier when I check in new comments > 2 lines.

PR #1065, opened 2023-03-15
Stop parsing CLI args when file path is found

### LINKED ISSUE #901: Stop command line processing once a script has been found
With this simple script:
```ruby
p ARGV
```
Calling this in Ruby:
```
ruby -rset args.rb -c foo
["-c", "foo"]
```
But Natalie keeps consuming the arguments, so instead of running the script with "-c foo" as arguments, it compiles an executable named foo. We have to manually add `--` to mark a boundary:
```
bin/natalie -rset -- args.rb -c foo
["-c", "foo"]
```

We should probably stop consuming the arguments once we've found an argument we shouldn't consume

seven1m:
This adds some functionality to optparse that MRI does not have. We were already a bit off from MRI's implementation anyway, but now we're fully embracing it. Fortunately, this should only add functionality and be non-breaking for any existing code that uses OptionParser.

Fixes #901


spec/library/rbconfig/rbconfig_spec.rb
```diff
@@ -30,17 +30,15 @@
...
+      RbConfig::CONFIG.each do |k, v|
+        if v.frozen?
+          puts "\#{k} Failure"
         end
-        puts 'Done'
-      RUBY
-    end
+      end
+      puts 'Done'
+    RUBY
```
seven1m, line 41: This test started passing thanks to improved CLI flag parsing! 🎉

PR #1066, opened 2023-03-16
Refactor String#% into Kernel.sprintf, Fix print

nanobowers:
Somewhat related to #869

I realized that the string format is essentially printf and sprintf, though the interface is slightly different.
I decided that the cleanest place for the formatter (since it was in Ruby) was to move it into `Kernel.sprintf` and then have `Kernel.printf` and `String#%` call `sprintf`.  If this was all in C++, maybe a non-ruby exposed function would've made more sense.

In any event, the tests for sprintf/printf appear more comprehensive than the string-format tests and a low percentage of them currently pass (~20%), so it will take some time to `xit` / `NATFIXME` them, planning to do that in another PR or add an additional commit to this one depending on feedback.

One other thing of note: one of the tests exposed an issue whereby `$stdout` is set by one of the tests but is set to an object.  MRI has some special handling of global variables like this whereby `$stdout` and `$stderr` are checked upon assignment that they have a `write` method available.  Natalie does not implement this (yet), so some additional and temporary checks were added to the related method (`Kernel.print`).  Another thing I noted is that `IO.print` should be calling `IO.write`, also for another future PR.

Contains:
+ Refactor `String#%` into `Kernel.sprintf`.
+ Add `Kernel.printf`, `Kernel.sprintf`, and `Kernel.format`
+ Fix `Kernel.print` to add some additional checking and temporarily call `IO#write` instead of `IO#print` to prevent crashing.


seven1m (approved):
nice work

PR #1071, opened 2023-03-18
Add unicode casemap for single characters

seven1m:
#217

This only adds support for one-to-one mappings as specified by the [UnicodeData.txt](http://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt) file.

~WIP: I'm still fiddling with this. My initial naive approach was to generate a very large `switch()` statement, but the compilation time for such a file is BAD and the resulting file is very large:~

## First (Naive) Approach

```
→ ls -lh build/generated/casemap*

-rw-rw-r-- 1 tim tim 3.2M Mar 18 12:05 build/generated/casemap.cpp
-rw-rw-r-- 1 tim tim  17M Mar 18 12:05 build/generated/casemap.cpp.o
```

I saw another approach by the Wine project, where they compressed the entire casemap table into a sparse array: https://github.com/wine-mirror/wine/blob/e909986e6ea5ecd49b2b847f321ad89b2ae4f6f1/libs/port/casemap.c

...Unfortunately, that library is LGPL licensed and I don't think we can just drop their file into our project.

## Second Approach

See comment below about new approach, which is similar to how Wine compresses their casemapping array.


nanobowers:
I had thought about this some and wondered if some means of using gperf or similar tool could be used to transcribe the mappings into generated c-code?  As you show, the unicodeData.txt is fairly large and can cause large code and compile times. I took a really short look at this a while ago and it never went anywhere, i got scared and moved on to other things :laughing: 

I assume based on this PR you're interested in rolling your own, but you mentioned some other approaches, so i did a little digging and there are a few c-libraries, including one from unicode with a fairly permissive license (https://github.com/unicode-org/icu) and this other one (https://github.com/contour-terminal/libunicode) with an apache license, unsure if they offer anything of use or have compatible licenses.

i'm not sure if you wanted to add dependencies or

seven1m:
I'd like to implement this without a library if possible. That's what we've done for our encoding stuff so far, so it seems on-brand to continue that way. 😀

I have another approach in the works that is similar to Wine's compressed array...

seven1m:
Thanks to [this article](https://github.com/apankrat/notes/tree/master/fast-case-conversion) I was able to implement a similar compression approach as Wine's, but it's written from scratch. This results in much smaller files and faster compilation times:

```sh
→ ls -lh src/encoding/casemap.cpp
-rw-rw-r-- 1 tim tim 165K Mar 19 08:50 src/encoding/casemap.cpp

→ ls -lh build/encoding/casemap.cpp.o
-rw-rw-r-- 1 tim tim 305K Mar 19 08:56 build/encoding/casemap.cpp.o
```

The new object file takes 0.06s to compile on my machine. 🎉

PR #1072, opened 2023-03-18
Fixed README.md Compiling part

DanielcoderX:
if you try to compile by putting ***-c hello*** at the end, it won't return executable output :) (in my case that was a problem) 
now it's ok


seven1m:
Thanks! This was recently "fixed" in #1065.

PR #1081, opened 2023-03-26
Add new ASCII based encodings

nanobowers:
Add a variety of ascii-based encodings, namely IBM*, ISO8859-* and Windows-1250..1252
Also adds IBM037, which is EBCDIC (non-ascii) based and a dummy encoding.

Similar to other our existing extended ascii-encodings, the conversion of codepoints above 7F is marked as `NOT_YET_IMPLEMENTED`.  The IBM037 / EBCDIC is marked as non-yet-implemented for the entire codepoint conversion.

These encodings allow for a few more specs to pass/continue that were previously NATFIXME'd


nanobowers:
I don't grok what the workflow is saying about the formatter. `rake format` runs clean on my end.

seven1m:
I think different versions of clang-format can produce slightly different results. 😬 Don't worry about it.

PR #1085, opened 2023-03-28
Continue after SyntaxError in REPL

tekknolagi:
Without this change, a syntax error in the REPL causes it to exit. This does not match up with the upstream REPL, which continues.


tekknolagi:
:partying_face:

PR #1092, opened 2023-04-10
IO advise, fsync, getbyte, readline, etc

nanobowers:
Added:
+ `IO#advise` with specs
+ `IO#fsync` and `IO#fdatasync` with specs
+ `IO#getbyte` with specs
+ `IO#path` and `IO#to_path` ; removed `File#path` binding
+ `IO#readline`, which calls `IO#gets`.  Like `IO#gets` this is an incomplete implementation and does not support all extended cases but helpful because many tests use it

Reordered methods in io_object.hpp header to be in alphabetical order where it made sense to.

Cleanup: Replaced long-form exception calling of `SystemCallError` with `raise_errno()`
Moved `m_path` instance variable from `FileObject` to `IoObject` and changed the type from a `tm::String` to a `Value`.  I believe this is consistent with how MRI 3.2 works, and was trying to be forward-looking here.

Added additional IO constants for `SEEK_DATA` and `SEEK_HOLE`


lib/natalie/compiler/binding_gen.rb
```diff
@@ -850,23 +849,30 @@ def generate_name
...
 gen.binding('IO', 'close', 'IoObject', 'close', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('IO', 'closed?', 'IoObject', 'is_closed', argc: 0, pass_env: false, pass_block: false, return_type: :bool)
+gen.binding('IO', 'fdatasync', 'IoObject', 'fdatasync', argc: 0, pass_env: true, pass_block: false, return_type: :int)
 gen.binding('IO', 'fileno', 'IoObject', 'fileno', argc: 0, pass_env: true, pass_block: false, return_type: :int)
+gen.binding('IO', 'fsync', 'IoObject', 'fsync', argc: 0, pass_env: true, pass_block: false, return_type: :int)
+gen.binding('IO', 'getbyte', 'IoObject', 'getbyte', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('IO', 'gets', 'IoObject', 'gets', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('IO', 'initialize', 'IoObject', 'initialize', argc: 1, pass_env: true, pass_block: false, return_type: :Object)
 gen.binding('IO', 'isatty', 'IoObject', 'isatty', argc: 0, pass_env: true, pass_block: false, return_type: :bool)
+gen.binding('IO', 'path', 'IoObject', 'get_path', argc: 0, pass_env: false, pass_block: false, return_type: :Object)
```
seven1m, line 862: I don't think IO has a `path` method in MRI. From my experimentation, it lives on the `File` class.

Screenshot 2023-04-11 7 29 24 AM

PR #1098, opened 2023-07-27
Fix CI builds

TheGrizzlyDev:
Docker builds weren't working because of recent breaking changes in pip that now requires system-wide package installations to be very explicit and otherwise prefer the package manager of the distro you are using.


seven1m (approved):
Nice, thank you!!! <3

TheGrizzlyDev:
> Nice, thank you!!! <3

You're very welcome, I hope to be able to contribute more like I used to!

PR #1100, opened 2023-08-05
Implement Set#divide

herwinw:


herwinw:
The current spec is very minimal. I pushed some extra tests upstream (https://github.com/ruby/spec/pull/1049), I want to wait for that one before I move this PR out of draft status. And of course I have to fix the last NATFIXME in this one as well

lib/set.rb
```diff
@@ -151,6 +151,19 @@ def disjoint?(other)
...
+    ret = Hash.new { |h, k| h[k] = Set.new }
+    if block.arity == 2
+      values = to_a.product(to_a)
+      values.each(&block)
+    else
+      each { |val| ret[block.call(val)] << val }
+    end
+    ret.values
+  end
+
```
seven1m, line 166: Wow, this is way simpler than I ever thought it would be. I had an implementation that worked for arity 1 but not 2, but it was wayyy bigger than this. Great work.

herwinw: Just to be clear: the arity 2 version does not work yet.

seven1m: Oh ok! 😆

herwinw: It turned out to be even easier than the first implementation: There is a `Set#classify` which does pretty much all the preparation we need.

The two argument version now works as well. The code is pretty ugly though.

herwinw:
The updated `divide_spec` has been merged upstream and is now integrated in this PR. I had to disable one of the new specs due to a bug in `enum_for`, which is out of scope of this change.

PR #1104, opened 2023-08-06
Implement String#crypt

herwinw:


herwinw:
Another checkbox for #217

herwinw:
The test of MacOS currently has an error "ld: library not found for -lcrypt". I have never used MacOS, so if somebody could point me in the right direction for this

PR #1110, opened 2023-08-10
Start with BasicSocket#send

herwinw:
There are still a few NATFIXMEs in the spec, they are all related to UDP sockets, TCP seems to work fine.

A script like this now works:
```ruby
s = TCPSocket.new('natalie-lang.org', 80)
s.send("GET / HTTP/1.0\r\nHost: natalie-lang.org\r\n\r\n", 0)
puts s.read
```


seven1m:
Cool!

PR #1112, opened 2023-08-10
Implement Matchdata#regexp

herwinw:


herwinw:
There is something wrong with the GC code in this change, a few scripts (e.g. `test/natalie/zlib_test.rb`) now result in GC related segfaults.
We can't simply circumvent the issue by copying the RegexpObject, the spec has a check to see if the object ids are equal, and there is no copy function for `regex_t` in Onigmo.

PR #1124, opened 2023-08-12
Use crypt_r(3) instead of crypt(3)

herwinw:
crypt(3) uses a global buffer for its output, which may cause issues if we ever add multithreading to Natalie. crypt_r(3) is the reentrant version.

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -3,7 +3,7 @@
 #include "natalie/string_unpacker.hpp"
 #include "string.h"
 
-#include <unistd.h>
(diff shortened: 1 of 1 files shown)


herwinw:
Apparently this one is not available on MacOS

herwinw:
We could get around here with a bunch of `#ifdef` statements and use `crypt(3)` on Mac instead of `crypt_r(3)`, but that would defy the purpose.
It would be easier to add some mutex around the operation to fix concurrency issues, but that can wait until we actually have threads.

PR #1139, opened 2023-08-15
First steps of OpenSSL::Digest

herwinw:
This is good enough to make the specs pass, even though there are a couple more interfaces to these objects


seven1m:
Nice!!

PR #1144, opened 2023-08-18
Openssl digest improvements

herwinw:
This is still very much a work in progress. The final version of probably going to contain another 5 refactors of the C++ part of this implementation.

My 2 goals for this project (except for the obvious one of implementing the missing methods in Natalie):

* [x] Extend the ruby-specs for `OpenSSL::Digest`. The current specs are woefully incomplete (https://github.com/ruby/spec/pull/1054)
* [X] ~~Autogenerate the subclasses by looping through the available digests of OpenSSL. For example, we don't have a MD5 digest, because the specs don't include that. By looping through the data of OpenSSL we can ensure that all digest available on the system are available as subclasses. This might be an OpenSSL 3.0 feature (`EVP_MD_do_all_provided`)~~

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -8,6 +8,11 @@
 
 using namespace Natalie;
 
+static void OpenSSL_MD_CTX_cleanup(VoidPObject *self) {
+    auto mdctx = static_cast<EVP_MD_CTX *>(self->void_ptr());
+    EVP_MD_CTX_free(mdctx);
+}
+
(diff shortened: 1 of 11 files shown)


herwinw:
I am going to abandon the goal for the automatic creation of subclasses. To do that, I need OpenSSL 3.0 or a working `const_missing` implementation, so that's going to be too much of a scope creep.
According to Rubocop, those subclasses are deprecated anyway.

herwinw:
So, I guess this is it. I've included the specs in the `test/natalie/openssl_digest` folder, since they haven't been merged upstream yet.

PR #1148, opened 2023-08-20
Update specs to remove checks for Ruby 3.0 or below

herwinw:
This should be merged *after* #1147

Sync all the local specs with upstream to remove the checks for Ruby version 3.0 or below.


herwinw:
Apparently the CI already passes (I had expected the Ruby2.7 check to fail, since this assumes we have at least 3.0). That means there's no need to wait.

seven1m:
The hosting Ruby version shouldn't normally affect the specs. It might just affect the sort of syntax and libraries we have available in the compiler itself.

PR #1152, opened 2023-08-22
Implement ENV.invert

herwinw:
Just out of curiosity: what is the use case for this method? I really can't think of one

### DIFF
--- a/include/natalie/env_object.hpp
+++ b/include/natalie/env_object.hpp
@@ -22,6 +22,7 @@ public:
     bool has_key(Env *, Value name);
     Value has_value(Env *, Value name);
     Value inspect(Env *);
+    Value invert(Env *);
     bool is_empty() const;
     Value key(Env *, Value);
     Value keys(Env *);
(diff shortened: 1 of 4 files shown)


ryangjchandler:
The only real thing I can think of is looking up an environment key by value..

```ruby
inverted = ENV.invert
key = inverted.key?(some_value)
```

herwinw:
But then there are more specific methods for that too:
```ruby
ENV.value?(some_value)
key = ENV.key(some_value)
```

seven1m:
I don't know when `ENV#key` and `Hash#key` were added (Ruby 3.2 maybe?), but I have used `invert` to look up a key by value in the past, e.g.:

```ruby
h = { a: 1 }
h.invert[1] # => :a
```

I only learned about `Hash#key` because of your recent work. 🎉

herwinw:
https://github.com/ruby/spec/commit/d530cdcd0c9eeabb1bccee7093b037bc43b6ea60#diff-8e0c31c79bb135a02001f89bebf952af22ddd9c8aae849f09f2379a2fcdc8318

Looks like `Hash#key` has been added in 1.8.8. But to be fair, I never knew about it either.

PR #1155, opened 2023-08-24
Start over with Kernel#sprintf

seven1m:
I wasn't happy with the readability of the old formatting code, and decided to rewrite it with a little state machine for the parser. I hope this turns out better!

#869

### DIFF
--- a/src/kernel.rb
+++ b/src/kernel.rb
@@ -69,56 +69,138 @@ module Kernel
 
     attr_reader :format_string, :index, :chars
 
+    TRANSITIONS = {
+      literal: {
+        on_percent: :field_beginning,
+      },
+      field_beginning: {
+        on_percent: :literal,
+        on_alpha: :field_end,
+        on_number: :field_width,
+      },
+      field_width: {
+        on_number: :field_width,
+        on_zero: :field_width,
+        on_period: :field_precision_period,
+        on_alpha: :field_end,
+      },
+      field_precision_period: {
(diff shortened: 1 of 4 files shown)


src/kernel.rb
```diff
@@ -60,20 +60,31 @@ def warn(*warnings, **)
...
+      chars[index]
+    }
+    next_number = lambda {
+      c = chars[index]
+      raise ArgumentError, 'bad width' unless ('1'..'9').cover?(c)
+      num = [c]
+      while ('0'..'9').cover?(c = next_char.())
+        num << c
+      end
+      num.join.to_i
```
herwinw, line 81: I would make `num` a String instead of an Array, this only requires two changes:
```ruby
num = c
# ...
num.to_i
```
I would argue this is easier to read, and as a bonus it should be more efficient too (we now only use 1 String, instead of 1 Array where each item is a String, so it removes the object creation/allocation from the loop.

herwinw: It should even be possible to remove the String intermediate and just work with an int instead:
```ruby
num = c.to_i
while ('0'..'9').cover?(c = next_char.())
  num = num * 10 + c.to_i
end
num
```
This might be a bit harder to read, and efficiency will now depend on the number of characters we're parsing. I guess in general that would be 1 number, where this should be a little bit faster (it removes a string copy operation). For long numbers this would probably be slower due to the amount of `to_i` calls)

PR #1156, opened 2023-08-24
Fix encoding issues for ENV

herwinw:


herwinw:
Even though this does remove the last NATFIXMEs in the env folder, I'm pretty sure I haven't fixed every output of the ENV object. I'll try to have a look at the missing specs to push upstream.

herwinw:
This has been merged into #1158 (to prevent merge conflicts)

PR #1166, opened 2023-08-27
Implement Fiber.blocking? and Fiber#blocking?

herwinw:


seven1m:
TIL about non-blocking Fibers and `Fiber::Scheduler`. Could be a fun bit of work to figure out!

seven1m (approved):

herwinw:
Just to be clear: The current status of blocking fibers in Natalie is pretty limited. At first, you need a scheduler to actually use this (https://github.com/socketry/async is probably the most used one, since MRI doesn't include one either). And second, the blocking status should be updated as soon as something in the Fiber is blocking (like blocking IO, or a Mutex), so the scheduler can choose to switch to another fiber.
https://ruby-doc.org/3.2.2/fiber_md.html has a pretty nice high level overview of the feature.

PR #1167, opened 2023-08-27
Implement Fiber#storage and friends

herwinw:
Even though this is a Ruby 3.2 feature, it can be a very useful tool to handle some issues with cyclic data structures in our code. For example, the cycle detection of `Set#inspect` only works for 1 level depth:
```ruby
def inspect
  items = to_a.map { |e| equal?(e) ? '#<Set: {...}>' : e.inspect }
  "#<Set: {#{items.join(', ')}}>"
end 
```
This passes the specs because the specs are simply 1 level deep:
```ruby
it "correctly handles self-references" do
  (set = Set[]) << set
  set.send(@method).should be_kind_of(String)
  set.send(@method).should include("#<Set: {...}>")
end
```
It would result in an infinite loop if we had a depth of more than 1:
```ruby
a = Set[]
b = Set[a]
a << b
a.inspect
```
With Fiber storage, we could resolve this with something like this:
```ruby
def inspect
  if Fiber[:set_inspect_current]
    if Fiber[:set_inspect_current].include?(object_id)
      "<Set {....}>" # Cycle detected 
    else
      Fiber[:set_inspect_current] << object_id
      "<Set: #{each(&:inspect).join(', ')}>" # Current inspect
    end
  else
    # This is the first in a non-recursive call:
    f = Fiber.new(storage: { set_inspect_current: [object_id] }) do
      result = "<Set: #{each(&:inspect).join(', ')}>" # Current inspect
      Fiber[:set_inspect_current] = nil
      result
    end
    f.resume
  end
end
```

We have a number of this kind of issues in the current code base.

### DIFF
--- a/include/natalie/fiber_object.hpp
+++ b/include/natalie/fiber_object.hpp
@@ -145,6 +145,7 @@ public:
     bool is_blocking() const;
     static Value is_blocking_current();
     Value resume(Env *env, Args args);
+    Value storage(Env *) const;
     void yield_back(Env *env, Args args);
 
     void *start_of_stack() { return m_start_of_stack; }
--- a/include/natalie/fiber_object.hpp
+++ b/include/natalie/fiber_object.hpp
@@ -145,6 +145,7 @@ public:
     bool is_blocking() const;
     static Value is_blocking_current();
     Value resume(Env *env, Args args);
+    Value set_storage(Env *, Value);
     Value storage(Env *) const;
     void yield_back(Env *env, Args args);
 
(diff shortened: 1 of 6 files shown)


herwinw:
This has a hidden dependency on https://github.com/ruby/spec/pull/1058, since the current specs are a bit limited.

herwinw:
I think this is ready now. Our first steps towards Ruby 3.2 compatibility, and as a bonus the upstream specs for fiber storage now test a lot more corner cases.

seven1m:
Cool!

PR #1170, opened 2023-08-27
Add stub file fiber.rb to lib

herwinw:
This helps with the specs that require this file.

This feels a bit like a hack, but since support for Ruby 3.0 will probably be dropped less than a year from now, I would think this is good enough.
This really improves the overview at https://natalie-lang.org/specs, where most of the Fiber specs now simply crash on the require statement.

### DIFF
--- a/lib/fiber.rb
+++ b/lib/fiber.rb
@@ -1,0 +1,5 @@
+# This is a stub file to help the specs that have `require "fiber"` in them. This makes it possible to remove all
+# the work arounds we have in our specs folder, and helps with displaying the actual status on the page
+# <https://natalie-lang.org/specs>.
+# Starting with Ruby 3.1, this file is no longer required and does not do anything on MRI. As soon as the Ruby
+# Specs drop support for Ruby 3.0, we can remove this file again.
--- a/spec/core/fiber/blocking_spec.rb
+++ b/spec/core/fiber/blocking_spec.rb
@@ -1,9 +1,7 @@
 require_relative '../../spec_helper'
 require_relative 'shared/blocking'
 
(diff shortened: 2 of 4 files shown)


lib/fiber.rb
```diff
@@ -0,0 +1,5 @@
+# This is a stub file to help the specs that have `require "fiber"` in them. This makes it possible to remove all
+# the work arounds we have in our specs folder, and helps with displaying the actual status on the page
+# <https://natalie-lang.org/specs>.
+# Starting with Ruby 3.1, this file is no longer required and does not do anything on MRI. As soon as the Ruby
+# Specs drop support for Ruby 3.0, we can remove this file again.
```
seven1m, line 5: Minor nitpick: code in the wild could have `require "fiber"` in it, so we should keep this stub around as long as MRI has it (vs removing this whenever ruby/spec stops using it).

seven1m (approved):

lib/fiber.rb
herwinw: That is pretty much forever. You can still `require 'thread'` in Ruby, which doesn't do anything since Ruby 2.3 IIRC.

Ruby has the `$LOADED_FEATURES` list that includes all the files that have been included (it probably uses some other data structure internally), so if you doe `require 'fiber'` on Ruby >=3.1, it just returns `false` and doesn't do anything. Our current solution still loads the file.
So the perfect solution would be to keep a list of preloaded files somewhere, so we don't need this stub file anymore. But that doesn't really fit in the compiler stage of Natalie.

seven1m: OK cool, good to know!

PR #1173, opened 2023-08-28
More Kernel#sprintf fixes

seven1m:


src/string_object.cpp
```diff
@@ -2774,9 +2774,22 @@ Value StringObject::convert_float() {
     char *endptr = nullptr;
     String string = String(m_string);
 
+    // check for two consequtive underscores
+    for (ssize_t i = 1; i < (ssize_t)string.length(); ++i) {
+        auto c2 = string[i];
+        auto c1 = string[i - 1];
+        if (c1 == '_' && c2 == '_')
+            return nullptr;
+    }
```
herwinw, line 2785: Any specific reason for the casts to `ssize_t`? The minimum value of `i` is `1`, the minimal index is `i - 1 == 0` (line 2780), so a regular `size_t` works fine here.

seven1m: Just remnants of earlier code and I forgot to remove the cast after refactor. Good catch.

PR #1174, opened 2023-08-28
Update spec for Set#inspect

herwinw:
The parser error of the original code has been resolved, so we remove this local workaround.


herwinw:
As some extra information: this syncs the file with an older version of the upstream spec. It has been updated today to include a deeper cycle, which we don't handle yet. #1167 includes the latest version of the spec, combined with a fix for the code.
For now it made more sense to me to just import the old one and run the test, instead of importing the new one and having to disable it.

PR #1194, opened 2023-09-02
Add read check in IO#getbyte

herwinw:
This is similar to #1190 where we had to convert the exception in case of no write access. Two helper methods have been added to help with this check, I expect we'll be needing a lot more of them.

https://github.com/ruby/spec/pull/1068 has been added as an upstream request to add this to the specs. For the time being, a local test has has been added.


herwinw:
This needs a different approach. The value of `sb.st_mode` is the mode on the filesystem, so if I have a file that is writable for the user, and open that file in read only mode, it would still return this file as writable.

herwinw:
This one is good to go now

PR #1195, opened 2023-09-02
Fix is_writable checks in IoObject

herwinw:
`fstat` reads the file system access flags, not the access flags used for this IO object.

This fixes the issue mentioned in #1194


seven1m:
Nice solution

PR #1197, opened 2023-09-02
Move aliases definitions into the binding definition

herwinw:
This has a few advantages:
* Reduce duplication
* No differences between things like number of arguments (in the old version one might add an extra argument to one method en forget the aliases).
* Having information on what methods are created as aliases helps with some checks of `UnboundMethod`, where aliased methods are seen as equal <https://github.com/ruby/spec/blob/c91ccefa3eec64a8e11a44306b1776411c6fcb75/core/unboundmethod/name_spec.rb#L9>.


herwinw:
This change only includes one alias for now, I want to have the idea and framework approved before I spend my time rewriting the other aliases.

seven1m (approved):
Yes! This is great. Don't know why I didn't think of that. 😄

herwinw:
For my own reference: the following duplicates are kept:
```
Found alias on Array: `===` vs `==` (type: `define_method`)
Found alias on Array: `intersection` vs `&` (type: `define_method`)
Found alias on Array: `to_ary` vs `to_a` (type: `define_method`)
Found alias on Array: `|` vs `union` (type: `define_method`)
Found alias on ENV: `each_pair` vs `each` (type: `define_singleton_method`)
Found alias on ENV: `size` vs `length` (type: `define_singleton_method`)
Found alias on ENV: `to_hash` vs `to_h` (type: `define_singleton_method`)
Found alias on ENV: `value?` vs `has_value?` (type: `define_singleton_method`)
Found alias on FalseClass: `^` vs `|` (type: `define_method`)
Found alias on File: `realpath` vs `realdirpath` (type: `define_singleton_method`)
Found alias on File: `unlink` vs `delete` (type: `define_singleton_method`)
Found alias on File: `zero?` vs `empty?` (type: `define_singleton_method`)
Found alias on FileTest: `executable_real?` vs `executable?` (type: `define_singleton_method`)
Found alias on Dir: `pwd` vs `getwd` (type: `define_singleton_method`)
Found alias on Dir: `rmdir` vs `delete` (type: `define_singleton_method`)
Found alias on Dir: `unlink` vs `delete` (type: `define_singleton_method`)
Found alias on Float: `quo` vs `/` (type: `define_method`)
Found alias on Kernel: `eql?` vs `===` (type: `define_method`)
Found alias on Kernel: `kind_of?` vs `is_a?` (type: `define_method`)
Found alias on Kernel: `public_methods` vs `methods` (type: `define_method`)
Found alias on Kernel: `to_s` vs `inspect` (type: `define_method`)
Found alias on Module: `attr_reader` vs `attr` (type: `define_method`)
Found alias on NilClass: `^` vs `|` (type: `define_method`)
Found alias on Range: `step` vs `%` (type: `define_method`)
```
Most often this was because the Ruby docs did not describe an explicit alias (especially `ENV` and `File` were subject to this). Sometimes the method did not exist according to the Ruby documentation (`Array#===` and most of the leftovers on `Kernel`). `FileTest#executable?` is actually different from `FileTest#exectuable_real?` (Using effective or real user id), so that's a bug on our side.

herwinw:
For the stats: the generated `bindings.cpp` has shrunk from 8780 to 8136 lines

PR #1207, opened 2023-09-05
Move responsibility of GC visiting members into MethodInfo class

seven1m:
That way we can let our GC lint script do the work to catch mistakes.

### DIFF
--- a/include/natalie/method_info.hpp
+++ b/include/natalie/method_info.hpp
@@ -30,9 +30,9 @@ public:
 
     bool is_defined() const { return !m_undefined && m_method; }
 
+    void visit_children(Cell::Visitor &);
+
 private:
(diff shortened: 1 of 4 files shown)


include/natalie/method_info.hpp
```diff
@@ -30,9 +30,9 @@ class MethodInfo {
 
     bool is_defined() const { return !m_undefined && m_method; }
 
+    void visit_children(Cell::Visitor &);
```
herwinw, line 33: This would probably need a `virtual` and `override` added to it (although I'm not really sure what happens without them).

seven1m: MethodInfo doesn't subclass from Cell, so it isn't virtual in this case.

PR #1213, opened 2023-09-06
Integrate YARP

seven1m:
This will be my branch to integrate [YARP](https://github.com/ruby/yarp) instead of [NatalieParser](https://github.com/natalie-lang/natalie_parser).

NatalieParser was born out of my frustration with the state of Ruby parsers and lack of a library that was free of a dependency on Ruby itself. Our humble parser project proved you can build a fast Ruby parser free of the tight coupling to MRI, but we didn't have the resources to make it fully correct. There are just a lot of edge cases with Ruby parsing I didn't have time to suss out.

YARP came along after and did everything I hoped NatalieParser could do, but with the backing of a major corporation and support from the Ruby community.

NatalieParser was quite a labor of love, and we learned a lot. I'd like to think it showed some in the Ruby community what was possible, and I have no regrets in building it.

Now it's time to be a little more practical for Natalie, given that there are so few of us on the this project, and we don't want to make building a parser our full-time OSS job. 😄 (I will archive the NatalieParser project once YARP is fully working with Natalie.)


herwinw:
Using upstream code? What's next, switching to `std::string`? :wink:

seven1m:
Ha! Possibly! 😆

seven1m:
All the tests in `test/natalie` are parsing and passing on my local machine. Progress! 🎉

seven1m:
All existing specs passing! Now I would like to start removing the translation layer and have pass1 consume the YARP nodes directly...

Rakefile
```diff
@@ -4,6 +4,7 @@ task default: :build
 
 DEFAULT_BUILD_TYPE = 'debug'.freeze
 SO_EXT = RUBY_PLATFORM =~ /darwin/ ? 'bundle' : 'so'
+SO_EXT2 = RUBY_PLATFORM =~ /darwin/ ? 'dylib' : 'so'
```
eregon, line 7: FYI this is called DLEXT in RbConfig terms

seven1m: Thanks!

PR #1221, opened 2023-09-09
Miscellaneous fixes to IO 

herwinw:
It started by adding the mode argument to `IO.new`, and from there it went down the rabbit hole, yak shaving, turtles all the way down and possibly some more animal references in an attempt to remove some of the new NATFIXME blocks.

This should be good enough to start working on the keyword arguments for `IO.new` and related methods, which are currently the biggest blocker for IO specs.


seven1m (commented):
This is great! ❤️

PR #1225, opened 2023-09-10
Keyword arguments for IO.new

herwinw:
This code is a mess, I don't think I've ever written so many nested if-statements without a deadline. And it's only partially because the Ruby code has to keep backwards compatibility in the options and provides like 4 different ways to set the encoding, which may clash, or might just overwrite things.

Another issue is that we extract the keyword arguments in the bindings, which results in 8 extra arguments to `initialize`, and we'll have to repeat that for every other method that uses these keyword arguments. This should be refactored to pass along the complete keywords arguments structure. Next, the interpretation of these keyword arguments should be extracted or combined with `fileutil::flags_obj_to_flags`, so we can reuse this code.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -24,6 +24,7 @@ namespace ioutil {
         int flags { O_RDONLY };
         EncodingObject *external_encoding { nullptr };
         EncodingObject *internal_encoding { nullptr };
+        bool autoclose { false };
 
         flags_struct(Env *env, Value flags_obj, HashObject *kwargs);
     };
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -20,6 +20,7 @@ namespace ioutil {
     StringObject *convert_using_to_path(Env *env, Value path);
     int object_stat(Env *env, Value io, struct stat *sb);
     struct flags_struct {
+        bool has_mode { false };
         int flags { O_RDONLY };
         EncodingObject *external_encoding { nullptr };
         EncodingObject *internal_encoding { nullptr };
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -104,7 +104,7 @@ public:
     Value stat(Env *) const;
     static Value sysopen(Env *, Value, Value = nullptr, Value = nullptr);
     Value read(Env *, Value, Value) const;
-    static Value read_file(Env *, Value, Value = nullptr, Value = nullptr);
+    static Value read_file(Env *, Args);
     Value readbyte(Env *);
     Value readline(Env *) const;
     int rewind(Env *);
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -20,7 +20,9 @@ namespace ioutil {
     StringObject *convert_using_to_path(Env *env, Value path);
     int object_stat(Env *env, Value io, struct stat *sb);
     struct flags_struct {
-        enum class read_mode { none, text, binary };
+        enum class read_mode { none,
+            text,
+            binary };
 
         bool has_mode { false };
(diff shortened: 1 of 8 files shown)


include/natalie/io_object.hpp
```diff
@@ -61,9 +62,10 @@ class IoObject : public Object {
     int fsync(Env *);
     Value getbyte(Env *);
     Value gets(Env *) const;
-    Value initialize(Env *, Value, Value = nullptr);
+    Value initialize(Env *, Value, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr, Value = nullptr);
```
seven1m, line 65: Good grief, lol

seven1m: I'm open to an alternative way of doing the bindings if you have any ideas. 😅

herwinw: My plan is to change the bindings to `argc: 1..`, so the complete `Args` objects is forwarded to `IoObject::initialize`, and from there forward it to some helper function to validate the number of elements and extract the keyword arguments.

seven1m: I dig it

herwinw:
So, this works, I think the code is of acceptable quality, and it's relatively easy to reuse for other IO keyword entries (I added `IO.read` as an example of the next rewrite.
Next up should probably be adding keyword arguments to `File.new`, but I kind of want to take a break from IO now.

include/natalie/io_object.hpp
```diff
@@ -20,11 +20,18 @@ namespace ioutil {
     StringObject *convert_using_to_path(Env *env, Value path);
     int object_stat(Env *env, Value io, struct stat *sb);
     struct flags_struct {
+        enum class read_mode { none,
+            text,
+            binary };
```
herwinw, line 25: I did not write it with this indentation, clang-format did. We might want to tweak the settings a bit for this one, but I don't think that should be blocking for this feature.

src/io_object.cpp
```diff
@@ -72,71 +72,206 @@ namespace ioutil {
...
-    flags_struct::flags_struct(Env *env, Value flags_obj) {
-        if (!flags_obj || flags_obj->is_nil())
-            return;
-
-        if (!flags_obj->is_integer() && !flags_obj->is_string()) {
-            if (flags_obj->respond_to(env, "to_str"_s)) {
-                flags_obj = flags_obj->to_str(env);
-            } else if (flags_obj->respond_to(env, "to_int"_s)) {
-                flags_obj = flags_obj->to_int(env);
+    namespace {
```
seven1m, line 75: Would it match Natalie style a bit more to have a Flags `class` with member functions vs this `namespace` and functions that accept `flags_struct`? I believe this is the only place we use this sort of style...

seven1m (approved):
This was a huge bit of work. Very good.

My one comment is just a nit about style. Not a blocker. Merging as-is is an option, and I can re-style the struct stuff later.

src/io_object.cpp
herwinw: I split that into #1247, right now I just kind of want to see the results.
This change is a pretty isolated one, the calling code will not see the results.

And for what it's worth: I chose this approach to reduce the compilation time: every private method in C++ means a change in the header file, which means we have to recompile pretty much everything. Since I had no clear vision of where I was going with this change (as can be seen by the 3 rewrites and countless related cases that prepared a thing or two for this change), this did help a bit.

herwinw:
@seven1m  I merged this change, and added an issue to do that refactoring whenever I can stand to look at this file again :wink:

src/io_object.cpp
seven1m: Great, works for me!

PR #1226, opened 2023-09-10
Implement Signal.list and Signal.signame

herwinw:


seven1m (approved):
Approved, assuming my commit fixes the failed build on macOS.

PR #1233, opened 2023-09-20
Implement FileTest#grpowned? and Process#groups

herwinw:
The specs of  `FileTest#grpowned?` depend on `Process#groups`, so this combination makes more sense than one would guess at first glance.


herwinw:
I did just push a newer version, it turns out you can call `getgroups(2)` with an empty buffer to query the number of groups. This might fix the Mac OS issue (or it might just as well be something completely unrelated).

herwinw:
https://github.com/natalie-lang/natalie/actions/runs/6275680663/job/17043679620?pr=1233

I added some debug output to the Mac OS run:
```
id -G
20 0 12 61 79 80 81 98 264 399 701 33 100 204 250 395 398 400

Run /usr/bin/id -G
20 0 12 61 79 80 81 98 264 399 701 33 100 204 250 395 398 400

Run ruby -e 'p Process.groups'
[20, 0, 12, 61, 79, 80, 81, 98, 264, 399, 701, 33, 100, 204, 250, 395, 398, 400]

Run bin/natalie -e 'p Process.groups'
[20, 0, 12, 61, 79, 80, 81, 98, 264, 399, 701, 33, 100, 204, 250, 395]
```
This still feels like we pass in a max of 16 and therefor don't retrieve the last two values.

herwinw:
Next attempt: use `getgrouplist(3)` instead of `getgroups(2)`. If that doesn't work, I'm tempted to just disable the tests on Mac.

seven1m:
I believe MRI is setting `_DARWIN_C_SOURCE` [here](https://github.com/ruby/ruby/blob/master/configure.ac#L1021-L1021).

With the following C program:

```c
#include <stdio.h>
#include <unistd.h>

int main() {
  int ngroups;

  ngroups = getgroups(0, NULL);
  printf("num = %d\n", ngroups);

  return 0;
}
```

...I get 16 as the result:

```sh
→ clang -o wat wat.c && ./wat
num = 16
```

...but if I define that macro, I get:

```sh
clang -o wat -D_DARWIN_C_SOURCE wat.c && ./wat   2.6.10p210
num = 17
```

herwinw:
I guess I was so focused on `Process.groups` that I completely forgot to update `FileTest#grpowned?` to fetch the required size of the groups list too. I updated that one too, let's hope the sixth time is a charm

seven1m:
Yay!

PR #1244, opened 2023-09-24
Implement File.readlink

herwinw:


src/file_object.cpp
```diff
@@ -514,6 +515,19 @@ StringObject *FileObject::path(Env *env, Value pathname) {
     return ioutil::convert_using_to_path(env, pathname);
 }
 
+Value FileObject::readlink(Env *env, Value filename) {
+    filename = ioutil::convert_using_to_path(env, filename);
+    TM::String buf(128, '\0');
+    while (true) {
+        const auto size = ::readlink(filename->as_string()->c_str(), &buf[0], buf.size());
```
seven1m, line 522: `&buf[0]` 🙈 😆

herwinw: Hey, it works :rofl: 
Maybe if I ever get around implementing that alternative string buffer constructor I proposed we can get rid of it.

PR #1250, opened 2023-09-24
Implement IO#open

herwinw:
There was a basic implementation for `File.open`. That one has been made more generic and moved to `IO.open`. The implementation now works for `IO`, `File` and `Socket`.

### DIFF
--- a/include/natalie/file_object.hpp
+++ b/include/natalie/file_object.hpp
@@ -21,7 +21,6 @@ public:
 
     Value initialize(Env *, Value, Value, Value, Block *);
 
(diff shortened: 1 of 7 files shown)


herwinw:
```
Failed specs:
File
  .open
    closes the file when the block has an exception
      false should be == to true
      (test/natalie/file_test.rb:72:in `block in block in block in block')
```
Makes you wonder why `IO.open` does not test this.

herwinw:
I rewrote the code in Ruby and suddenly it works.
Note to self: improve upstream specs.

Issue #1184, opened 2023-08-31
Unblocked Fibers


herwinw:
This is a feature of Ruby 3.0 where a Fiber can mark itself as being blocked, notify a scheduler of this information, and let the scheduler change the execution to a different fiber. Given that we currently have Fibers but no Threads, this is probably the easiest way into getting some form of concurrency into Natalie.

I found some example scripts on https://blog.monotone.dev/ruby/2020/12/25/ruby-3-fiber.html and I want those to work with Natalie.

This does require the following steps:
* [x] Add blocking status to Fiber (#1166)
* [ ] Add a possibility to get and set the Fiber scheduler (#1183)
* [ ] Get the `kernel_sleep` example to work. This requires at least updates to the `sleep` method to make the callback
* [ ] Get the `io_wait` example to work
* [ ] Get the `block`/`unblock` example to work. This probably requires some more primitives in the code, like Mutex and Queue

seven1m:
@herwinw I need to contact you via email if possible. Can you email me at tim@timmorgan.org?

PR #1255, opened 2023-09-25
Implement IO.pipe

herwinw:
This is one of those things that is used in a lot of specs (I see 37 occurrences in the specs folder when I exclude the `pipe_spec.rb` file). Most relevant for me is `IO.select`, which is one of the things I need for  #1184.

For now the only issue in the specs is the encoding, but that is partially because we don't acknowledge the `Encoding.default_external`/`Encoding.default_internal` in the generic code of the IO object. Since character encoding hardly works in Natalie, this shouldn't be too much of an issue for the first version.
Another issue: if you run this with a subclass of IO, and that subclass has a `close` method that throws an error, your program will probably crash. The `Defer` code has been copied from the old implementation of `File.open`, which had this issue too when I moved that to `IO.open` and ran the specs that actually test that. I'll try to add those tests to the upstream specs soon-ish.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -91,6 +91,7 @@ public:
     bool is_closed() const { return m_closed; }
     bool is_eof(Env *);
     bool isatty(Env *) const;
+    static Value pipe(Env *, Args, Block *, ClassObject *);
     int pos(Env *);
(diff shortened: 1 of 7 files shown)


herwinw:
And another compilation error on Mac. I'm starting to consider getting a cheap second hand macbook from somewhere just to speed up the feedback process.

PR #1257, opened 2023-09-26
Implement StringIO#printf

herwinw:
This should get another 140 tests passing, which is not bad for a change of 4 lines of code


seven1m:
Wow! 🎉

herwinw:
So, in the past few days the number of passing specs of the daily tests was 12785, 12893, 13001, 13188. That were some good days.

PR #1259, opened 2023-09-26
Make Complex#to_s spec compliant

timcraft:


spec/core/complex/to_s_spec.rb
```diff
@@ -41,4 +42,13 @@
...
   end
+
+  it "treats real and imaginary parts as strings" do
+    real = NumericSpecs::Subclass.new
+    real.should_receive(:to_s).and_return("1")
+    imaginary = NumericSpecs::Subclass.new
+    imaginary.should_receive(:to_s).and_return("2")
+    imaginary.should_receive(:<).any_number_of_times.and_return(false)
+    Complex(real, imaginary).to_s.should == "1+2i"
+  end
```
seven1m, line 53: We need to either upstream this to [ruby/spec](https://github.com/ruby/spec) or move it to our own `test/natalie/complex_test.rb`, otherwise it will be lost when we update specs from ruby/spec later.

timcraft: It's from ruby/spec, I updated from here:
https://github.com/ruby/spec/blob/master/core/complex/to_s_spec.rb#L46-L53

src/complex.rb
```diff
@@ -32,7 +32,7 @@ def to_s
     if real.respond_to?(:nan?) && real.nan?
       s << "NaN"
     else
-      s << real.inspect
+      s << real.to_s
```
herwinw, line 35: This looks like the reversal of #1223, where I fixed `Complex#inspect`, so I guess this change is going to work for `Complex#to_s` but break `Complex#inspect`.

I think `real.send(__method__)` will work (and the same change on line 55)

spec/core/complex/to_s_spec.rb
seven1m: Oh, my bad. I checked a local copy of ruby/spec and failed to `git pull`. Thanks!

src/complex.rb
herwinw: Kind of my bad as well, when I fixed `Complex#inspect` and forgot to check `Complex#to_s`.

herwinw: That's probably the best solution. You could work around with Fiber storage (similar to `Set#inspect`), but I'm not sure if that's going to be much better.

timcraft: Done, not ideal but it works

herwinw: Another option that looks to be working:
```ruby
to_s_wrapper = lambda do |obj, method_name|
  real = obj.real
  imaginary = obj.imaginary
  # From here it's pretty much just the old to_s implementation
end

# Need to use define_method to create a closure to access to_s_wrapper
define_method(:to_s) do 
  to_s_wrapper.call(self, :to_s)
end
```
This makes the code a bit more convoluted, but at least it doesn't show up in the private methods of the object

PR #1260, opened 2023-09-26
Implement Warning module

timcraft:
This adds the `Warning` module and a few other bits of functionality required to get the specs passing:

* ~`IO#sync=` which doesn't really do anything yet but is used in `spec/core/warning/warn_spec.rb`~
* The `-w` command line option which is used in `spec/core/warning/element_reference_spec.rb`
* A couple of fixes for `test/support/spec.rb`

Let me know if you want those pulling off into separate pull requests to make this easier to review! I've used NATFIXME for a couple of the specs which also depend on specific warnings that aren't implemented yet

### DIFF
--- a/spec/core/warning/warn_spec.rb
+++ b/spec/core/warning/warn_spec.rb
@@ -1,0 +1,135 @@
+require_relative '../../spec_helper'
+
+describe "Warning.warn" do
+  it "complains" do
+    -> {
+      Warning.warn("Chunky bacon!")
+    }.should complain("Chunky bacon!")
+  end
+
+  it "does not add a newline" do
+    ruby_exe("Warning.warn('test')", args: "2>&1").should == "test"
+  end
+
+  it "returns nil" do
+    ruby_exe("p Warning.warn('test')", args: "2>&1").should == "testnil\n"
+  end
+
+  it "extends itself" do
+    Warning.singleton_class.ancestors.should include(Warning)
+  end
+
+  it "has Warning as the method owner" do
+    ruby_exe("p Warning.method(:warn).owner").should == "Warning\n"
+  end
+
+  it "can be overridden" do
+    code = <<-RUBY
+      $stdout.sync = true
+      $stderr.sync = true
+      def Warning.warn(msg)
+        if msg.start_with?("A")
+          puts msg.upcase
+        else
+          super
+        end
+      end
+      Warning.warn("A warning!")
+      Warning.warn("warning from stderr\n")
+    RUBY
+    ruby_exe(code, args: "2>&1").should == %Q[A WARNING!\nwarning from stderr\n]
+  end
+
+  xit "is called by parser warnings" do
+    Warning.should_receive(:warn)
+    verbose = $VERBOSE
+    $VERBOSE = false
+    begin
+      eval "{ key: :value, key: :value2 }"
+    ensure
+      $VERBOSE = verbose
+    end
+  end
+
+  it "is called by Kernel.warn with nil category keyword" do
+    Warning.should_receive(:warn).with("Chunky bacon!\n", category: nil)
+    verbose = $VERBOSE
+    $VERBOSE = false
+    begin
+      Kernel.warn("Chunky bacon!")
+    ensure
+      $VERBOSE = verbose
+    end
+  end
+
+  it "is called by Kernel.warn with given category keyword converted to a symbol" do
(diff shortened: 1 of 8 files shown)


src/io_object.cpp
```diff
@@ -616,6 +616,21 @@ Value IoObject::stat(Env *env) const {
     return new FileStatObject { sb };
 }
 
+bool IoObject::sync(Env *env) const {
+    if (m_closed) {
+        env->raise("IOError", "closed stream");
+    }
```
herwinw, line 622: There is a helper for this boilerplate:
```c++
raise_if_closed(env);
```

timcraft: Sweet, thanks!

herwinw:
I have one issue with the stubs for `IO.sync` and `IO.sync=`: the specs all pass for this incomplete implementation, which means it's very easy to forget we actually have to implement it.

I think there are three possibilities here
* Put a `// NATFIXME: Actually implement the logic for this field` somewhere in the code
* Create an issue to actually implement it
# Use a NATFIXME block as in https://github.com/natalie-lang/natalie/blob/c110b81a7fd9b2b98e8d85db6040e93da5ef03f9/spec/core/io/printf_spec.rb#L8 to disable the `IO#sync=` calls in the specs

I would think the third option would be the best indication that there is actually work to be done, and it also keeps this PR a bit more focused on the warnings module.

src/io_object.cpp
```diff
@@ -616,6 +616,21 @@ Value IoObject::stat(Env *env) const {
...
+    return m_sync;
+}
+
+Value IoObject::set_sync(Env *env, Value mode) {
+    if (m_closed) {
+        env->raise("IOError", "closed stream");
+    }
+    m_sync = mode->is_truthy();
```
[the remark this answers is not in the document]
timcraft: Makes sense, I've removed the IO#sync change

timcraft:
> Update: After reading the full change: `NATFIXME` is not going to work in this case. It's probably best to remove/outcomment those lines for the time being. The overview on https://natalie-lang.org/specs will still mark them as failed, so we can always see there is something that needs to be done.

Sure, I've commented out the lines for that spec

timcraft:
@herwinw Accidentally edited out your latest comment trying to quote it, sorry!

spec/core/warning/warn_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  #         puts msg.upcase
+  #       else
+  #         super
+  #       end
+  #     end
+  #     Warning.warn("A warning!")
+  #     Warning.warn("warning from stderr\n")
+  #   RUBY
+  #   ruby_exe(code, args: "2>&1").should == %Q[A WARNING!\nwarning from stderr\n]
+  # end
```
herwinw, line 42: `IO#sync=` is now implemented, this one can be reactivated.

spec/core/warning/warn_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+  # it "is called by parser warnings" do
+  #   Warning.should_receive(:warn)
+  #   verbose = $VERBOSE
+  #   $VERBOSE = false
+  #   begin
+  #     eval "{ key: :value, key: :value2 }"
+  #   ensure
+  #     $VERBOSE = verbose
+  #   end
+  # end
```
herwinw, line 54: It should be possible to use the `NATFIXME` block call instead:
```ruby
it "is called by parser warnings" do
  NATFIXME 'Warning for hash key duplicated and overwritten' do
    Warning.should_receive(:warn)
    # Rest of the test
  end
end
```
This way the tests expect a failure, and we'll get a notification as soon as it is fixed.

spec/core/warning/warn_spec.rb
timcraft: Done

spec/core/warning/warn_spec.rb
timcraft: I tried that originally but it fails with `Issue has been fixed, please remove or update the NATFIXME marker`

herwinw: I've seen that Schrödingers Fixme before, I would guess we're having problems with identifying the `should_receive` in the block.
In this case, it's easier to disable the block by replacing `it` with `xit`, that makes it show up as a skipped test. This time I actually tested it to see if it works ;)

timcraft: I've updated the code to use `xit`

PR #1268, opened 2023-09-29
Implement IO#select

herwinw:
This should be the final shaved yak before we can work on the `io_wait` example with unblocked Fibers (#1184)


seven1m (approved):
Nice!

PR #1270, opened 2023-09-30
Sync before close in IO.open

herwinw:
I was running into weird errors with the `File.open` spec, which fails at "opens a file that no exists when use 'a' mode" with a "Bad file descriptor (Errno::EBADF)", but the error was thrown in `touch` at the setup.
The man page of `close(2)` contains the following paragraph:

> A careful programmer who wants to know about I/O errors may precede
> close() with a call to fsync(2).

Adding a call to `fsync` appears to fix the issue.

### DIFF
--- a/src/io.rb
+++ b/src/io.rb
@@ -10,6 +10,7 @@ class IO
         yield(obj)
       ensure
         begin
+          obj.fsync unless obj.tty?
           obj.close
         rescue IOError => e
           raise unless e.message == 'closed stream'


seven1m:
Thank you! I saw this pop up a few times locally but when I went to investigate it stopped happening. 😆

herwinw:
> Thank you! I saw this pop up a few times locally but when I went to investigate it stopped happening. 😆

That happened to me too. "Let's just print something to stderr to see where it breaks" and suddenly it worked again.

PR #1273, opened 2023-09-30
Implement OpenSSL::KDF.pbkdf2_hmac

herwinw:
The next step towards encryption: the key derivation functions. The current code works for the happy path, but it needs a lot more tests for the failure path. And there are two more KDFs, which we can implement in separate pull requests.

https://ruby.github.io/openssl/OpenSSL/KDF.html

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -99,6 +99,72 @@ Value OpenSSL_Digest_digest_length(Env *env, Value self, Args args, Block *) {
     return IntegerObject::create(digest_length);
 }
 
+Value OpenSSL_KDF_pbkdf2_hmac(Env *env, Value self, Args args, Block *) {
+    auto kwargs = args.pop_keyword_hash();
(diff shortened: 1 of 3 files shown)


herwinw:
It turns out OpenSSL keeps an internal list with occurred errors, which you can use to find the relevant error message, and you're supposed to clean it up after an error occurred. So instead of all the "Internal OpenSSL error" we have in the digest code, we should use this functionality. See https://www.openssl.org/docs/man1.1.1/man3/ERR_get_error.html

Yes, I'm using this PR as a personal notebook/todo list :wink:

herwinw:
That should be all. The tests are a bit unstructured, sometimes behaviour and error messages depend on the version of OpenSSL.

herwinw:
Welcome to Natalie, we don't have threads, we don't know anything about character encodings, but you can interface with this unpronounceable acronym of OpenSSL :D

PR #1287, opened 2023-10-01
Add Fiber scheduler hooks for Kernel.sleep

herwinw:
This will be my work in progress for the coming time. I've gotten a very primitive (and incomplete) scheduler that switches Fibers in MRI. Next up is calling the scheduler from within Natalie, and we'll have our first actual checkbox for #1184.


herwinw:
This should do it. In hindsight I should have started with the limitless sleep, since most of the code is converting timestamps from and to various representations.

seven1m:
So cool!!! 😎

PR #1288, opened 2023-10-02
Run subfolders of test/natalie/ too

herwinw:
I was surprised to see the CI of #1287 passing, turns out we don't run tests in subfolders. So all those shiny new tests for OpenSSL::Digest are skipped now.


seven1m:
Yikes!

PR #1293, opened 2023-10-03
Add DLEXT and SOEXT to RbConfig::CONFIG

herwinw:
This should be the easiest way to fix the filename issue in  #1291

### DIFF
--- a/test/natalie/rbconfig_test.rb
+++ b/test/natalie/rbconfig_test.rb
@@ -1,0 +1,18 @@
+require_relative '../spec_helper'
+require 'rbconfig'
+
+describe 'RbConfig::CONFIG' do
+  platform_is :linux do
+    it 'uses the correct values for DLEXT and SOEXT' do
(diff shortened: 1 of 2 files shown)


test/natalie/rbconfig_test.rb
```diff
@@ -0,0 +1,18 @@
...
+      RbConfig::CONFIG['SOEXT'].should == 'so'
+    end
+  end
+
+  platform_is :darwin do
+    it 'uses the correct values for DLEXT and SOEXT' do
+      RbConfig::CONFIG['DLEXT'].should == 'bundle'
+      RbConfig::CONFIG['SOEXT'].should == 'dylib'
+    end
+  end
```
herwinw, line 17: This is the main reason for adding this test: this runs in MRI as well, so this way I can validate I didn't mix up those two values.

seven1m:
Thanks!

PR #1299, opened 2023-10-05
Add experimental support for the common Ruby idiom `$LOAD_PATH << stuff`

seven1m:
This sort of thing is at the top of a lot of Ruby files:

```ruby
$LOAD_PATH << File.expand_path('../../build/prism/lib', __dir__)
require 'prism'
```

In fact, that above is from our own `lib/natalie/parser.rb`. Everything in that code is determinable at compile-time, it just takes a bit of Ruby execution to figure it out. Luckily, we already have a VM that can do (a limited amount of) stuff, and I was able to teach it how to get the current `__dir__`, so that's cool.

I don't know if this is madness or genius, but... mu ha ha ha. I digress. (I am also a bit sick today, and probably shouldn't write code while lying in bed 😆)


lib/natalie/compiler/macro_expander.rb
```diff
@@ -189,6 +205,23 @@ def macro_include_str!(expr:, current_path:)
         end
       end
 
+      # $LOAD_PATH << some_expression
+      def macro_append_load_path(expr:, current_path:)
+        if @depth > 0
+          return drop_error(:LoadError, "Cannot maniuplate $LOAD_PATH at runtime")
```
nirvdrum, line 211:
```suggestion
          return drop_error(:LoadError, "Cannot manipulate $LOAD_PATH at runtime")
```

lib/natalie/compiler/macro_expander.rb
```diff
@@ -73,7 +79,17 @@ def macro?(node)
...
+          get_macro_name(node[1])
+        else
+          get_hidden_macro_name(node)
+        end
+      end
+
+      # "Hidden macros" are just regular-looking Ruby code we intercept at compile-time.
+      # We will try to support common Ruby idioms here that cannot be done at runtime.
+      def get_hidden_macro_name(node)
+        if node[0..2] == s(:call, s(:gvar, :$LOAD_PATH), :<<)
```
nirvdrum, line 91: I'm not familiar with Natalie's internals. Would this also catch `$: <<`?

test/natalie/load_path_test.rb
```diff
@@ -0,0 +1,15 @@
...
+$LOAD_PATH << File.expand_path('../support/load_path', __dir__)
+require 'works'
+
+describe '$LOAD_PATH' do
+  it 'has some basic support for compile-time appending' do
+    LoadPathWorks.should be_an_instance_of(Class)
+  end
+
+  it 'does not support conditional modification' do
+    -> { $LOAD_PATH << 'foo' }.should raise_error(LoadError, 'Cannot maniuplate $LOAD_PATH at runtime')
```
nirvdrum, line 13:
```suggestion
    -> { $LOAD_PATH << 'foo' }.should raise_error(LoadError, 'Cannot manipulate $LOAD_PATH at runtime')
```

lib/natalie/compiler/macro_expander.rb
seven1m: No, but I added a commit to support that one too. Thanks!

eregon:
Nice!

PR #1300, opened 2023-10-05
Rework Kernel.sleep

herwinw:
Make it spec compliant (except for the Thread stuff), and make the code a bit more generic. This last part should help with #1287.


herwinw:
Fun fact: the method returns the actual amount of time slept, floored down to seconds. MRI returns `1` about half the time I try `sleep(0.6)`, Natalie has a stable return value of `0`. So I guess we're more accurate in sleeping than MRI is :partying_face:

PR #1301, opened 2023-10-05
Use prism nodes directly for pass1

kddnewton:
This pull request is the beginning of work to have pass1 use prism nodes directly instead of going through a sexp translation layer. It purposefully does _not_ do every node, in an effort to make this easier to review, and to make it so that other contributors can do this incrementally. This pull request migrates 11 node types.

The first four are integer, rational, imaginary, and float. In the old AST these are all `:lit` nodes. In prism's AST they are all their own node types. For the parser we alias all their visit methods to `visit_passthrough` since they are all leaves in the old AST so that they get returned directly. For pass1 we implement our new `transform_` methods for each of the node types that is effectively a copy of one of the branches in `transform_lit`.

The next commit is false, true, nil, and self. These are also leaves and can be done the same as the previous examples. The only other issue is that some places explicitly create false, true, and nil nodes to use in place of missing nodes or as return values for macros. For these we manually create prism nodes. I don't love the interface to create these nodes (it seems really verbose to say `::Prism::NilNode.new(nil)` but that's probably something that can be addressed later.

Then we do 3 commits with and, or, and defined nodes. It's important to note that these nodes are inner nodes and not leaves. So we cannot just alias them to `visit_passthrough` in the parser because that would leave all of their children in place as well. Fortunately prism has a `copy` API we can use, that will reuse all of the attributes of a node that _aren't_ given to the copy method. So when we're parsing those nodes, we use the copy method with the visited output. So these now become prism nodes holding onto sexp instances. It's a little weird, but it works.

Hopefully this enables the rest of the tree to be migrated as well without too much pain. If this is a good direction after review, we can set up a tracking issue or something to make sure we don't step on each others toes when working on different nodes.

### DIFF
--- a/lib/natalie/parser.rb
+++ b/lib/natalie/parser.rb
@@ -390,9 +390,7 @@ module Natalie
           location: node.location)
       end
 
-      def visit_imaginary_node(node)
-        s(:lit, node.value, location: node.location)
-      end
+      alias visit_imaginary_node visit_passthrough
 
       def visit_interpolated_regular_expression_node(node)
         dregx = visit_interpolated_stringish_node(node, sexp_type: :dregx, unescaped: false)
--- a/lib/natalie/parser.rb
+++ b/lib/natalie/parser.rb
@@ -623,7 +623,7 @@ module Natalie
       end
 
       def visit_or_node(node)
-        s(:or, visit(node.left), visit(node.right), location: node.location)
+        node.copy(left: visit(node.left), right: visit(node.right))
       end
 
       def visit_parameters_node(node)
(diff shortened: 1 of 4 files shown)


seven1m (approved):
Brilliant!

lib/natalie/parser.rb
```diff
@@ -623,7 +623,7 @@ def visit_optional_parameter_node(node)
       end
 
       def visit_or_node(node)
-        s(:or, visit(node.left), visit(node.right), location: node.location)
+        node.copy(left: visit(node.left), right: visit(node.right))
```
seven1m, line 673: I like this `copy` interface.

PR #1302, opened 2023-10-05
Allow manipulating $LOAD_PATH from a required file

seven1m:
...also fixes a bug caused by untimely PR merges on my part.


seven1m:
I think the CI failure on macos was a fluke.

PR #1306, opened 2023-10-06
Implement Kernel.system (it's complicated)

herwinw:
This is simply a proxy into chaining `Kernel.spawn` and `Process.wait` with the arguments forwarded. This means this version has the same limitations as `Kernel.spawn`, most notably it only supports arguments to the executable as separate arguments to the method, which means `system("echo foo")` fails. For this reason, the specs have not been included.

Main rationale for adding this is the nightly spec runner. Currently pretty much every spec in `core/kernel` fails because `Kernel.system` does not exist, even though these specs don't need that one. Adding this method will fix most (or maybe all) of these crashes (currently 96), so we can see how well we actually do for the Kernel module.

### DIFF
--- a/spec/core/file/setuid_spec.rb
+++ b/spec/core/file/setuid_spec.rb
@@ -28,15 +28,15 @@ describe "File.setuid?" do
     it "returns true when the gid bit is set" do
       platform_is :solaris do
         # Solaris requires execute bit before setting suid
-        NATFIXME 'Implement Kernel#system', exception: NoMethodError, message: "undefined method `system'" do
-          system "chmod u+x #{@name}"
-        end
-        `chmod u+x #{@name}`
+        system "chmod u+x #{@name}"
+        # NATFIXME: Fix arguments for Kernel#system
+        system "chmod", "u+x", @name
       end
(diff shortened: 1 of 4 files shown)


seven1m (approved):

seven1m (approved):
lgtm

spec/core/file/setuid_spec.rb
```diff
@@ -28,15 +28,15 @@
...
+        # NATFIXME: Fix arguments for Kernel#system
+        system "chmod", "u+x", @name
       end
-      NATFIXME 'Implement Kernel#system', exception: NoMethodError, message: "undefined method `system'" do
+      NATFIXME 'Fix arguments for Kernel#system', exception: SpecFailedException do
         system "chmod u+s #{@name}"
+        File.setuid?(@name).should == true
       end
-      `chmod u+s #{@name}`
+      system "chmod", "u+s", @name
```
herwinw, line 39: This is a bit of a mess now, since we can run the `system` method, but it doesn't do what it is supposed to do. I added a second check for `File.setuid?` into this block, so at least we get a notification once this starts working.

I'll try to look into the arguments of `Kernel.spawnf` soon-ish, before this spreads.

Issue #1247, opened 2023-09-24
Cleanup of construction of flags_struct


herwinw:
Right now, this struct has a very simple constructor, which defers most of the work to a bunch of parser methods in an anonymous namespace. This could be rewritten to some private methods on the struct (or class).

(Related case: #1225)

herwinw:
And maybe split off the `ioutil` module into its own source file, that only gets included in the cpp files that use it.

PR #1310, opened 2023-10-07
Split ioutil into its own source file

herwinw:
Step 1 of #1247 (so I guess that means another 1246 steps to go)

This isolates the code a bit more, and removes the need to recompile everything after a change to this class has been made.


seven1m:
>Step 1 of #1247

😆

seven1m (approved):

PR #1311, opened 2023-10-07
Improve Kernel#spawn

herwinw:
Support the single string argument form, split that into separate arguments.

The specs will still fail, it looks like our output_to_fd capture is broken, since running this spec with MRI using our test library fails too.


seven1m (approved):
Nice.

I guess we get to implement this someday soon: https://ruby-doc.org/stdlib-2.6.5/libdoc/shellwords/rdoc/Shellwords.html 😅

PR #1313, opened 2023-10-07
Remove guards for Ruby version <= 3.0

herwinw:
Since Ruby 3.0 is the lowest version to be supported. These changes are in sync with the upstream version of the specs.

### DIFF
--- a/spec/core/warning/element_reference_spec.rb
+++ b/spec/core/warning/element_reference_spec.rb
@@ -1,11 +1,9 @@
 require_relative '../../spec_helper'
 
 describe "Warning.[]" do
-  ruby_version_is '2.7.2' do
-    it "returns default values for categories :deprecated and :experimental" do
-      ruby_exe('p [Warning[:deprecated], Warning[:experimental]]').chomp.should == "[false, true]"
-      ruby_exe('p [Warning[:deprecated], Warning[:experimental]]', options: "-w").chomp.should == "[true, true]"
-    end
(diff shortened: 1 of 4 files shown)


seven1m:
I wonder if we should change this to say 3.2: https://github.com/natalie-lang/natalie/blob/master/src/natalie.cpp#L405 (And around Christmas time, maybe change it to 3.3)

And if we do, does that mean we can eliminate more guards like you did in this PR?

(We support being hosted by MRI down to 3.0, but in terms of our own runtime, I think we can say we are are targeting 3.2 currently.)

herwinw:
> And if we do, does that mean we can eliminate more guards like you did in this PR?

I think we should look at the two commits in this PR separately. The first commit has some updates to the spec folder. These files were out of date with the upstream version (The block_spec is even more out of sync, so I might take another shot at trying to fix that one, since half of the checks are disabled for Natalie). As long as upstream specs support Ruby 3.0, we will have those version guards. So we won't be able to eliminate more guards in this folder
The second commit checks the test folder with our own test. We won't be able to eliminate more guards in this folder either, for the simple reason that this commit removes them all.

herwinw:
> I wonder if we should change this to say 3.2

I think that change is pretty separate from this PR. It will probably cause a few specs to fail, so it's not as easy as just bumping the version number.

seven1m:
>I think that change is pretty f/.4separate from this PR.

Ah yes, I didn't mean on this PR. I was just wondering aloud for the near future.

PR #1316, opened 2023-10-08
Prefer block version of File.open in io.rb

herwinw:
This removes the need to manually close the file descriptor, and still closes it if an exception occurs in the block.


seven1m:
Thanks friend!

PR #1320, opened 2023-10-09
Rework IO#gets

herwinw:
This fixes the following issues:
* Accept lines of more than 1024 bytes
* Accept files not ending with a newline
* Improve performance by reading chunks instead of separate bytes

My long term plan was to fix the remaining spec incompatibilities of `IO#gets`, then implement `IO#readline` and `IO#readlines` using `IO#gets`. I may have gotten sniped by #1314 for that one. I will still try to finish this method and have another look at all the duplication we have now.

### DIFF
--- a/spec/core/io/gets_spec.rb
+++ b/spec/core/io/gets_spec.rb
@@ -359,15 +359,13 @@ describe "IO#gets" do
 
   it "uses the default external encoding" do
     @io = new_io @name, 'r'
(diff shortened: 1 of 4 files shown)


seven1m:
>I may have gotten sniped by #1314 for that one.

that's what I'm here for 😂

seven1m (approved):

PR #1322, opened 2023-10-10
Bump Ruby target version to 3.2

herwinw:
This does not include any changes to the behaviour of the code, all failing tests have been wrapped in NATFIXME blocks.

### DIFF
--- a/spec/core/fiber/storage_spec.rb
+++ b/spec/core/fiber/storage_spec.rb
@@ -1,161 +1,160 @@
-# NATFIXME: Even though this is a spec for Ruby 3.2, we're going to implement parts of
-# it to help with detecting cycling data structures.
-
 require_relative '../../spec_helper'
 
-describe "Fiber.new(storage:)" do
-  it "creates a Fiber with the given storage" do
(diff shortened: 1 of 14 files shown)


herwinw:
We mentioned this briefly in #1313

spec/core/fiber/storage_spec.rb
```diff
@@ -1,161 +1,160 @@
...
-  it "creates a Fiber with the given storage" do
-    storage = {life: 42}
-    fiber = Fiber.new(storage: storage) { Fiber.current.storage }
-    fiber.resume.should == storage
-  end
-
-  it "creates a fiber with lazily initialized storage" do
-    Fiber.new(storage: nil) { Fiber[:x] = 10; Fiber.current.storage }.resume.should == {x: 10}
-  end
+ruby_version_is "3.2" do
```
herwinw, line 3: Just to be clear: we were missing one level of indentation compared to the upstream version, since we implemented fiber storage while still targeting Ruby 3.1. So even though it looks like we changed everything in this file, it's mostly just a new level of indentation.

PR #1324, opened 2023-10-10
Leave in place class variable nodes

kddnewton:
This PR attempts to leave in place all class variable -related nodes. This includes all 6 related nodes including:

* class variable read
* class variable target
* class variable write
* class variable and write
* class variable or write
* class variable operator write

Some logic could probably be shared in the future between the and/or/operator writes, but for now I thought it better to be explicit while in this transitionary period.

The target nodes are really the most difficult, because they appear in 3 places you might not expect. I fixed the first two, but the last one will require a bigger change than I thought made sense for this PR. Besides, it's already broken today, so I figured it probably wasn't as big of a deal. They are:

* `begin; rescue => @@foo; end` - This was working in the previous version, so this works now.
* `@@foo, = 1` - This was working in the previous version, so this works now.
* `for @@foo in bar; end` - This was never working, and it looks like the for loop generally assumes that it only receives local variables, so that will need a larger change.

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -135,6 +135,22 @@ module Natalie
         instructions
       end
 
+      def transform_class_variable_and_write_node(node, used:)
+        instructions = [
+          ClassVariableGetInstruction.new(node.name, default_to_nil: true),
+          IfInstruction.new,
+          transform_expression(node.value, used: true),
+          ClassVariableSetInstruction.new(node.name),
+          ClassVariableGetInstruction.new(node.name),
+          ElseInstruction.new(:if),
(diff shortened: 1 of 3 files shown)


lib/natalie/compiler/pass1.rb
```diff
@@ -119,6 +119,65 @@ def transform_array_node(node, used:)
         instructions
       end
 
+      def transform_class_variable_read_node(node, used:)
+        return [] unless used
+        ClassVariableGetInstruction.new(node.name)
+      end
+
+      def transform_class_variable_target_node(node, used:)
+        raise "Target nodes should not be marked as used" if used
```
seven1m, line 128: I like this guard

PR #1327, opened 2023-10-11
Leave Prism instance variable nodes in place for pass1

seven1m:


lib/natalie/compiler/instructions/send_instruction.rb
```diff
@@ -4,9 +4,9 @@ module Natalie
   class Compiler
     # use:
     # push(block) if block
+    # push(receiver)
     # push(*args)
     # push(argc)
-    # push(receiver)
```
seven1m, line 9: This comment has been wrong for quite a while apparently. 😅

PR #1334, opened 2023-10-13
Implement File#flock

herwinw:


seven1m (approved):

herwinw:
It looks like the `NATFIXME` blocks didn't really work well with one of the tests that produced a lot of stderr output. I rewrote that one with a `xit` instead. It's one of the many places where we don't have Threads.

seven1m:
Speaking of threads, I just read this yesterday: https://redmine.ruby-lang.org/issues/19842

I'm trying not to get nerd sniped on Threads right now, though, as I already have a few irons in the fire. But someday!

PR #1336, opened 2023-10-13
Enable specs for constant on a singleton class

herwinw:
The comment about time out is outdated.


herwinw:
Disclaimer: it works on my machine, we might still trigger time outs in the CI for other setups. In that case this PR should be closed.

seven1m (approved):

Issue #1340, opened 2023-10-14
Add read buffer to IoObject


herwinw:
It started with this observation in the specs of `IO#sysread`:
https://github.com/ruby/spec/blob/59bdcb4ea95c60159bb2bfc8c73022364da8ec0d/core/io/sysseek_spec.rb#L24-L27

It turns out that CRuby uses a read buffer when reading IO. This can also be observed by the following code (using `lines.txt` from the folder `spec/core/io/fixtures/`:
```ruby
ruby -e 'File.open("lines.txt") { |f|
  puts f.gets
  puts f.gets
}
```
Using `strace` on this code shows the following calls for the file and `stdout`:
```
openat(AT_FDCWD, "lines.txt", O_RDONLY|O_CLOEXEC) = 5
ioctl(5, TCGETS, 0x7ffdf8282b30)        = -1 ENOTTY (Inappropriate ioctl for device)
read(5, "Voici la ligne une.\nQui \303\250 la li"..., 8192) = 137
write(1, "Voici la ligne une.\n", 20Voici la ligne une.
)   = 20
write(1, "Qui \303\250 la linea due.\n", 21Qui è la linea due.
) = 21
close(5)                                = 0
```
Currently, Natalie does this by using `pread` (which is a read without advancing the file position), determining the length of the line, then setting the file position to the length of the line, and repeat that for the second line.
This current Natalie code does not work on IO without random access (like a socket). We could resolve this by getting back to reading a single byte at a time again, but that destroys the performance.

The biggest issue in implementing this is that we need to alter every file access method that uses the file position at the same time, otherwise things will get out of sync.  This probably means the writing methods need some alterations too.

herwinw:
Assigning this one to myself, with the disclaimer that it might take a while before I'll finish this one.

PR #1342, opened 2023-10-14
Implement IO#pwrite

herwinw:


herwinw:
Just to be clear: this explicitly bypasses all buffering of #1340.

PR #1344, opened 2023-10-14
Import specs for IO#write and IO.write and allow non-ascii variable names

herwinw:
The current version of Natalie is unable to compile this file: `bad var name: :ë`. This pull request does include a change to accept non-ascii variable names. I have no idea how well this works with the self hosted compiler mentioned in #1315, so this commit might have to be excluded


seven1m:
>I have no idea how well this works with the self hosted compiler

Should be fine! 🙏

PR #1347, opened 2023-10-14
Add Mutex class

herwinw:
We'll need this one for the block/unblock in #1184. Most of the upstream specs will probably fail, since they all depend on threads, we might have to duplicate a few of them using fibers.

Once we have this one, we can probably build the other synchronisation primitives (like Queue and ConditionVariable) in plain Ruby using this Mutex class, and get the Scheduler hooks for free.


herwinw:
I'm removing this one, in favour of  #1541

PR #1351, opened 2023-10-15
Use ensure_no_missing_keywords in OpenSSL::KDF

herwinw:
This removes a lot of boilerplate code.


lib/openssl.cpp
```diff
@@ -104,46 +104,16 @@ Value OpenSSL_KDF_pbkdf2_hmac(Env *env, Value self, Args args, Block *) {
...
-    } else {
-        iterations = iterations->to_int(env);
-    }
-    auto length = kwargs->remove(env, "length"_s);
-    if (!length) {
-        missing_keywords.push("length");
-    } else {
-        length = length->to_int(env);
-    }
+    env->ensure_no_missing_keywords(kwargs, { "salt", "iterations", "length", "hash" });
```
seven1m, line 107: I totally forgot about `ensure_no_missing_keywords`. We should probably move that to the `Args` class to match the other `ensure_*` methods...

PR #1354, opened 2023-10-15
Add read buffer to IoObject

### LINKED ISSUE #1340: Add read buffer to IoObject

herwinw:
Closes #1340


herwinw:
This adds the read buffer, implements `IO#ungetbyte`, uses the read buffer in `IO#gets`, adds the `sep` and `limit` arguments to `IO#gets`, and all the changes to `IO#gets` are implemented for `IO#readline` too. A few issues with `IO#gets` are still open, but I would think these are outside of the scope of this case.

PR #1362, opened 2023-10-19
Finish MatchData#begin and MatchData#end

herwinw:
#1359 contained a pretty basic implementation, these additions make it pass the tests.


seven1m:
Sniped! :laughing:

PR #1367, opened 2023-10-20
Add a basic, incomplete, kinda-wrong implementation of File.fnmatch

seven1m:
~I don't know if this is a stupid way to do this (using regexps). After writing the code I learned about `fnmatch(3)` which maybe does this for us? Though I'm not sure it handles `**` ... I will do a bit more research before I work on this further!~

Indeed this was dumb. But I'm shipping it, for now at least.

### DIFF
--- a/include/natalie/match_data_object.hpp
+++ b/include/natalie/match_data_object.hpp
@@ -70,6 +70,8 @@ public:
         visitor.visit(m_regexp);
     }
 
+    virtual String dbg_inspect() const override;
+
     /**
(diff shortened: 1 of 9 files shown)


herwinw:
I have looked into `fnmatch(3)` before, that one does not do everything we need here.

seven1m:
This was a mistake.

>Some people, when confronted with a problem, think, “I know, I’ll use regular expressions.” Now they have two problems.

But I am going to ship it as-is because I want this functionality, and it's better than nothing.

PR #1376, opened 2023-10-22
Implement Dir.glob

seven1m:
There are some edge cases I'm leaving unimplemented. This is good-enough for any case I've seen where `Dir[]` and `Dir.glob` were used.


seven1m:
FYI my changes to the `fnmatch` specs should hopefully be upstreamed here: https://github.com/ruby/spec/pull/1099

PR #1386, opened 2023-10-24
Implement String#count

richardboehme:
This is probably not complete (for example it does not respect escape characters yet), but it's better than nothing and it passes all the specs!

[#217]


seven1m (approved):
@richardboehme returns! ❤️

richardboehme:
> @richardboehme returns! ❤️

Heeey :wave: I wish I had more time to work on Natalie :disappointed:

PR #1392, opened 2023-10-25
Enable test in file/new_spec

herwinw:
It works fine on my system, the comment mentions MacOS, so let's see if the CI agrees.


herwinw:
https://github.com/natalie-lang/natalie/actions/runs/6638885218/job/18036153149?pr=1392

By the looks of it, it works in MacOS too

PR #1394, opened 2023-10-26
Build self-hosted compiler as part of CI

seven1m:
Would be nice for this to not break again :-)


tekknolagi:
Wow! Does this mean it is back up and running now?

seven1m:
Yes sir! There are bugs of course, and it is SLOW, but it does compile, and it can print "hello world" and do some other small things.

There is a bug with encoding somewhere when I run the test suite with the self-hosted compiler, so I need to figure that out. But progress!

```sh
→ bin/natalie -c bin/nat bin/natalie
→ bin/nat --version
Natalie 0.1
```

PR #1395, opened 2023-10-26
Rewrite buffered IO to use FILE* instead of own read buffer.

herwinw:
It might take a few days before this one is done.


seven1m:
I'd like to tidy up our PR list. Can we close this for now?

herwinw:
Yeah, this one has been staring me in the face for a while.

I expected this to be required to implement `IO.popen`, but we can probably work around that by using `fileno(3)`. That means we still need this for a few obscure specs to pass, but that might become an unbalanced ratio for effort and profit.
I will close this one, but save the branch so we can always work on it if we ever need this again.

Reading between the lines, I guess I should check on the other 20 in progress PRs too.

PR #1403, opened 2023-10-28
Added a back to top option

vaibhav-009:
Added a back to top option in the readme file. On clicking it the webpage will scroll to the top.


seven1m:
No thank you.

PR #1410, opened 2023-10-29
Implement OpenSSL::Cipher

herwinw:
This is going to be a long term project. The code itself is probably not that sophisticated (validate Ruby type, convert to C type, pass data to OpenSSL function), but all the specs need to be written as well.

Documentation for the gem: https://www.rubydoc.info/stdlib/openssl/OpenSSL/Cipher

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -62,6 +62,13 @@ Value OpenSSL_Cipher_initialize(Env *env, Value self, Args args, Block *) {
     return self;
 }
 
+Value OpenSSL_Cipher_block_size(Env *env, Value self, Args args, Block *) {
+    args.ensure_argc_is(env, 0);
+    auto ctx = static_cast<EVP_CIPHER_CTX *>(self->ivar_get(env, "@ctx"_s)->as_void_p()->void_ptr());
+    const nat_int_t block_size = EVP_CIPHER_CTX_block_size(ctx);
+    return Value::integer(block_size);
+}
+
 Value OpenSSL_Cipher_decrypt(Env *env, Value self, Args args, Block *) {
     // NATFIXME: There is deprecated behaviour when calling with arguments. Let's not try to reproduce that for now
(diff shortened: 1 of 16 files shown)


herwinw:
Just to set the expectation here: I want to get the most basic example from the docs to work. That is:
```ruby
data = "Very, very confidential data"

cipher = OpenSSL::Cipher.new('AES-128-CBC')
cipher.encrypt
key = cipher.random_key
iv = cipher.random_iv

encrypted = cipher.update(data) + cipher.final
# ...
decipher = OpenSSL::Cipher.new('AES-128-CBC')
decipher.decrypt
decipher.key = key
decipher.iv = iv

plain = decipher.update(encrypted) + decipher.final

puts data == plain #=> true
```

There are a lot of fancy additional features in this Ruby class (and the OpenSSL libs), but they will have to wait.

seven1m:
>I want to get the most basic example from the docs to work.

❤️

herwinw:
I made a small variation of the example script:
```ruby
data = "Very, very confidential data"
cipher = OpenSSL::Cipher.new('AES-128-CBC')
cipher.encrypt
cipher.key = '1' * cipher.key_len
cipher.iv = '2' * cipher.iv_len
encrypted = cipher.update(data) + cipher.final
puts encrypted.unpack1('H*')

ciphertext = ["f161a8f5829bb47089cf5c5b021515f68b1133495b29d8c1d8481474a4dd576d"].pack('H*')
cipher = OpenSSL::Cipher.new('AES-128-CBC')
cipher.decrypt
cipher.key = '1' * cipher.key_len
cipher.iv = '2' * cipher.iv_len
plaintext = cipher.update(ciphertext) + cipher.final
puts plaintext
```
This replaces the random key and iv with fixed values, and prints the ciphertext. The big binary text is the actual ciphertext. Running this in Natalie results in the same output as running this in MRI.

seven1m (approved):

PR #1411, opened 2023-10-29
Fix bug in local_variable_or_write transform

richardboehme:
The value that might be assigned to the local variable has to be duped. Otherwise, exiting the current scope of the conditional assignment would pop of any value of the stack that was there before, potentially erasing something from the stack that's important. This broke for instance the following example:

```ruby
case 1
when 1
  text ||= ""
end
```


seven1m (commented):
We have the same problem in `transform_local_variable_and_write_node` yes?

seven1m (approved):

richardboehme:
> We have the same problem in `transform_local_variable_and_write_node` yes?

Yes seems like it. Probably also for all class variable types. Maybe I can fix those tomorrow!

PR #1413, opened 2023-10-30
Fix variable ||= and &&= compilation

seven1m:
Following from the discovery in #1411, this should fix all the variable `||=` and `&&=` transforms.


seven1m:
This is wrong. I will fix tomorrow!

seven1m:
I taught `IfInstruction` to yell loudly whenever the two branches have different stack sizes , and I fixed another few bugs found as a result. 🎉

PR #1419, opened 2023-10-30
Make MatchData#regexp spec compliant

timcraft:
The `$~` variable is already implemented. There's a similar problem with MatchData#string which I can't figure out a fix for right now, but last match needs to be set somewhere for String#gsub calls.


timcraft:
> So the full implementation is a bit more complicated than this.

Agreed, I'll close this for now.

PR #1426, opened 2023-11-02
Eval locals

seven1m:
This fixes `eval()` so it can see local variables like this:

```ruby
four = 4
eval('four + 5') # => 9
```

As a side effect of this work, we can remove the REPL variable hack, which is nice.


seven1m:
Going to wait on this since Prism will soon have a way to pass in locals without need for my hack here. :tada: 

ruby/prism#1763

PR #1432, opened 2023-11-04
Copy PrettyPrinter from MRI

seven1m:


lib/pp.rb
```diff
@@ -0,0 +1,657 @@
...
+    end
+  end
+
+  module PPMethods
+
+    # Yields to a block
+    # and preserves the previous set of objects being printed.
+    def guard_inspect_key
+      if Thread.current[:__recursive_key__] == nil
+        Thread.current[:__recursive_key__] = {}.compare_by_identity
```
herwinw, line 169: It looks like we only use the Thread class for thread-local storage. I think we can replace this with `Fiber.current` to get the same functionality, which saves us from the bazillion other changes related to Thread in the specs.

seven1m: I'm ok with the changes. I think it's a milestone to copy PrettyPrinter as-is without any tweaks needed.

PR #1434, opened 2023-11-04
Implement OpenSSL::X509::Name.parse

herwinw:
I was just trying to finish up the remaining specs for OpenSSL. It turns out this test is pretty much an integration test, it needs a bunch of additional methods implemented before I can even start working on the spec itself. And the last OpenSSL spec looks even worse in terms of functionality required to even run the test.
But at least I'm learning a lot about OpenSSL this way :partying_face:

### DIFF
--- a/lib/openssl.rb
+++ b/lib/openssl.rb
@@ -117,8 +117,14 @@ module OpenSSL
         'emailAddress' => ASN1::IA5STRING
       }.tap { |hash| hash.default = ASN1::UTF8STRING }.freeze
 
+      __constant__('COMPAT', 'int', 'XN_FLAG_COMPAT')
+      __constant__('RFC2253', 'int', 'XN_FLAG_RFC2253')
+      __constant__('ONELINE', 'int', 'XN_FLAG_ONELINE')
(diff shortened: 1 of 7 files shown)


lib/openssl.rb
```diff
@@ -12,6 +12,35 @@ def self.secure_compare(a, b)
...
+    __constant__('IA5STRING', 'int', 'V_ASN1_IA5STRING')
+    __constant__('UTCTIME', 'int', 'V_ASN1_UTCTIME')
+    __constant__('GENERALIZEDTIME', 'int', 'V_ASN1_GENERALIZEDTIME')
+    __constant__('GRAPHICSTRING', 'int', 'V_ASN1_GRAPHICSTRING')
+    __constant__('ISO64STRING', 'int', 'V_ASN1_ISO64STRING')
+    __constant__('GENERALSTRING', 'int', 'V_ASN1_GENERALSTRING')
+    __constant__('UNIVERSALSTRING', 'int', 'V_ASN1_UNIVERSALSTRING')
+    __constant__('BMPSTRING', 'int', 'V_ASN1_BMPSTRING')
+  end
+
```
seven1m, line 43: :clap:

PR #1437, opened 2023-11-04
Add Object::to_hash

herwinw:
This is the same type of coercion method as to_ary or to_str. Use this to remove some code duplication.

This is some code of  #1434 split off into a separate branch.


seven1m (approved):
So clean

herwinw:
There are probably a lot of locations where we could simplify the calls to `to_str` or `to_int` as well (I'm pretty sure I've written a number of the using `send` in the past). I'll have a look into that tomorrow. It looks like the CI is stressed enough already ;)

PR #1438, opened 2023-11-04
Implement Complex#**

seven1m:
Finishes #608

I am not a math guy, so this was a struggle for me. But it does pass the specs and it allows us to close a big issue!

### DIFF
--- a/src/complex.rb
+++ b/src/complex.rb
@@ -168,6 +168,37 @@ class Complex
     end
   end
 
+  def **(n)
+    if n.is_a?(Float) && n == 0.0
+      return Complex(1.0, 0.0)
+    elsif n.is_a?(Integer) && n == 0
+      return Complex(1)
+    end
+
+    if n.is_a?(Complex)
+      r = Math.sqrt(real ** 2 + imaginary ** 2)
+      theta = Math.atan2(imaginary, real)
+      r_n = Math.exp(n.real * Math.log(r) - n.imaginary * theta)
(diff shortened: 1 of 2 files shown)


src/complex.rb
```diff
@@ -168,6 +168,37 @@ def *(other)
...
+      theta = Math.atan2(imaginary, real)
+      r_n = r ** n
+      theta_n = n * theta
+      Complex(r_n * Math.cos(theta_n), r_n * Math.sin(theta_n))
+    elsif n.is_a?(Integer)
+      # NOTE: this branch is so we don't get floats in the result.
+      # If you know a better formula for that, please fix!
+      result = Complex(1, 0)
+      n.times { result *= self }
+      result
```
herwinw, line 195: Generally, you can make exponentiation faster by grouping operations. For example `n ** 4` results in three multiplications:
```ruby
n * n * n * n
````
But if we group this, we can see that we're doing the same operations twice:
```ruby
(n * n) * (n * n)
```
So we can save this in a temporary value and reduce it to two multiplications:
```ruby
n_tmp = n * n
n_tmp * n_tmp
```
But to be fair, I'm not sure how this is going to work out with the `Complex(1, 0)` base value in this case.

I'm pretty sure this gets mentioned in more detail in [Elements of Programming](http://elementsofprogramming.com/), which to me very much was like "Yeah, it's interesting to know this exists, but there is no way I am ever going to use the information of this book in practice".

PR #1441, opened 2023-11-05
OpenSSL sockets

herwinw:
This might take a while before we get there, but I want to get the following code to work:

```ruby
TCPSocket.open('natalie-lang.org', 443) do |socket|
  ssl_context = OpenSSL::SSL::SSLContext.new()
  ssl_socket = OpenSSL::SSL::SSLSocket.new(socket, ssl_context)
  ssl_socket.connect
  ssl_socket.write("GET / HTTP/1.1\r\nHost: natalie-lang.org\r\nConnection: close\r\n\r\n")

  p ssl_socket.read
  ssl_socket.close
end
```
Fun fact: I have never used the SSL sockets directly (normally you would use something like net-http), so I had to look up how to do this.

Some documentation links:
* https://wiki.openssl.org/index.php/SSL/TLS_Client
* https://wiki.openssl.org/index.php/Simple_TLS_Server (shows the socket more clearly)
* https://www.rubydoc.info/stdlib/openssl/OpenSSL/SSL/SSLContext
* https://www.rubydoc.info/stdlib/openssl/OpenSSL/SSL/SSLSocket
* https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_new.html

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -27,6 +27,11 @@ static void OpenSSL_SSL_CTX_cleanup(VoidPObject *self) {
     SSL_CTX_free(ctx);
 }
 
+static void OpenSSL_SSL_cleanup(VoidPObject *self) {
+    auto ssl = static_cast<SSL *>(self->void_ptr());
+    SSL_free(ssl);
+}
+
 static void OpenSSL_X509_NAME_cleanup(VoidPObject *self) {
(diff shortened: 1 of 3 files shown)


herwinw:
I think this is the minimal viable version to get this script to work. It probably has about a hundred security issues (I have no idea what TLS/SSL versions are allowed by default), so please don't use it for any serious applications.

seven1m (approved):

PR #1442, opened 2023-11-05
FIx negative number raised to fractional power

herwinw:
The changes from #1438 made this one pretty easy. It did require a small fix to the logic in `complex.rb`, otherwise this code would end up with an assertion error:
```ruby
Complex(-8.0) ** Rational(1, 3)
```


seven1m:
Nice!

PR #1443, opened 2023-11-05
Normalize type coercion

herwinw:
And reduce a whole lot of code duplication.


seven1m (approved):
So nice and clean. Makes me wonder if we need a linter to keep ourselves (read: me 😅) from using the wrong conversion in the future. 🤔

herwinw:
If only things were that simple. There are a few occasions where `nil` is an accepted return value for these conversions, good luck explaining these exceptions to your linter :wink:

PR #1449, opened 2023-11-05
Implement Module#class_variable_get and Module#class_variable_set

richardboehme:


richardboehme:
Not sure if this CI failure is related to my change... Maybe somebody has an idea why this bad file descriptor error appears?

herwinw:
It looks like a random CI failure, especially since the Mac os test passes. Pushing the branch again will trigger a new run, I guess that will resolve it.
If there's a retry button in the CI, that might fix it as well, but I've never managed to find that button.

richardboehme:
Ok thank you! I triggered a rerun. The button is normally on the right hand side of each failed job.

image

src/module_object.cpp
```diff
@@ -362,6 +375,27 @@ Value ModuleObject::cvar_set(Env *env, SymbolObject *name, Value val) {
...
+    return val;
+}
+
+Value ModuleObject::class_variable_set(Env *env, Value name, Value value) {
+    assert_not_frozen(env);
+    if (name->is_string()) {
+        name = name->as_string()->to_symbol(env);
+    } else if (!name->is_symbol()) {
+        name = name->to_str(env)->to_symbol(env);
+    }
```
seven1m, line 394: Should we use `Object::to_symbol()` here? I only ask because @herwinw has done some recent work to DRY up all these conversion bits of code and I *think* `to_symbol()` makes sense for this. (Not 100% sure though.)

seven1m (approved):
You can take or leave the `Object::to_symbol()` suggestion -- it may not be right for the specs. Merge when ready!

src/module_object.cpp
richardboehme: I think that's actually more correct. For example calling `class_variable_set` with `nil` as the first argument raises a wrong error in natalie currently. Changing it to use `Object::to_symbol()` fixes this!

PR #1451, opened 2023-11-06
String char/byte index optimizations

seven1m:
There are a few places where we were operating with a character index and converting to a byte index, which is super inefficient, because we have to iterate from the start of the string over each character until we get to the right place.

My main goal with this branch is to improve speed of String#scan and String#gsub by using a byte index. These methods are super slow on large inputs -- hoping to fix that.

### DIFF
--- a/include/natalie/encoding/ascii_8bit_encoding_object.hpp
+++ b/include/natalie/encoding/ascii_8bit_encoding_object.hpp
@@ -33,6 +33,8 @@ public:
     virtual String encode_codepoint(nat_int_t codepoint) const override;
     virtual nat_int_t decode_codepoint(StringView &str) const override;
     virtual bool is_ascii_compatible() const override { return true; };
+
+    virtual bool is_single_byte_encoding() const override final { return true; }
 };
 
 }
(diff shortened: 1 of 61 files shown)


seven1m:
These optimizations make it so our self-hosted compiler can actually run a spec! 🎉 

```sh
→ rake bootstrap
→ bin/nat test/natalie/method_test.rb
......................................

38 spec(s) passed.
```

It's still dog-slow compared to the MRI-hosted compiler, but not glacial as it was before.

```sh
→ hyperfine 'bin/natalie -d cpp test/natalie/method_test.rb' 'bin/nat -d cpp test/natalie/method_test.rb'
Benchmark 1: bin/natalie -d cpp test/natalie/method_test.rb
  Time (mean ± σ):     324.3 ms ±   9.6 ms    [User: 265.1 ms, System: 58.9 ms]
  Range (min … max):   313.4 ms … 347.0 ms    10 runs
 
Benchmark 2: bin/nat -d cpp test/natalie/method_test.rb
  Time (mean ± σ):      8.984 s ±  0.165 s    [User: 8.655 s, System: 0.328 s]
  Range (min … max):    8.769 s …  9.298 s    10 runs
 
Summary
  bin/natalie -d cpp test/natalie/method_test.rb ran
   27.70 ± 0.97 times faster than bin/nat -d cpp test/natalie/method_test.rb
```

PR #1454, opened 2023-11-07
Only close IO object on destruction if autoclose is enabled

### LINKED ISSUE #1452: Occasional Errono::EBADF when running tests
On my local machine about 1 out of 5 times I get this error:

```
        11: from spec/core/file/new_spec.rb:4:in `<main>'
        10: from /home/tim/pp/natalie/test/support/spec.rb:1469:in `block in block'
        9: from /home/tim/pp/natalie/test/support/spec.rb:1416:in `run_specs'
        8: from /home/tim/pp/natalie/test/support/spec.rb:1416:in `each'
        7: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `block in run_specs'
        6: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `reverse_each'
        5: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `block in block in run_specs'
        4: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `reverse_each'
        3: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `block in block in block in run_specs'
        2: from /home/tim/pp/natalie/test/support/spec.rb:1452:in `call'
        1: from spec/core/file/new_spec.rb:13:in `block in block in block'
spec/core/file/new_spec.rb:13:in `close': Bad file descriptor (Errno::EBADF)
```

It's also occurring sometimes on CI: https://github.com/natalie-lang/natalie/actions/runs/6763302781/job/18381713844

seven1m:
...and default autoclose to true.

This fixes #1452.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -31,9 +31,9 @@ public:
     virtual ~IoObject() override {
         if (m_fileno == STDIN_FILENO || m_fileno == STDOUT_FILENO || m_fileno == STDERR_FILENO)
             return;
-        if (!m_closed && m_fileno != -1) {
-            ::close(m_fileno);
+        if (m_autoclose && !m_closed && m_fileno != -1) {
             m_closed = true;
+            m_fileno = -1;
         }
     }
 
@@ -125,7 +125,7 @@ private:
     int m_fileno { -1 };
     int m_lineno { 0 };
     bool m_closed { false };
-    bool m_autoclose { false };
+    bool m_autoclose { true };
     bool m_sync { false };
     StringObject *m_path { nullptr };
     TM::String m_read_buffer {};
--- a/src/io_object.cpp
+++ b/src/io_object.cpp
@@ -607,13 +607,14 @@ Value IoObject::pwrite(Env *env, Value data, Value offset) {
 Value IoObject::close(Env *env) {
     if (m_closed)
         return NilObject::the();
+
     int result = ::close(m_fileno);
-    if (result == -1) {
+    if (result == -1)
         env->raise_errno();
-    } else {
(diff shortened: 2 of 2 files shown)


herwinw:
https://github.com/natalie-lang/natalie/blob/master/spec/core/io/close_spec.rb#L34-L40

There is a spec for this behaviour, which is now disabled (that one did crash consistently on my system). I tried running that spec with this branch, but it still crashes.

seven1m:
Maybe `IoObject::close()` is supposed to not call `::close()` in that case. 🤔

herwinw:
Yes, that's a different code path (your fix only handled the CPP destructor). I tried a quick and dirty fix for this one:
```diff
 Value IoObject::close(Env *env) {
-    if (m_closed)
+    if (m_closed || !m_autoclose)
         return NilObject::the();
```
That fixes this issue. We might be able to make this a bit more generic (by making the destructor call close), but we could save that for later. I think right now preventing these crashes in the tests should be more important than making everything clean.

seven1m:
Even for an IO object with autoclose=false, it still pretends to close, so I'm going to add that behavior to this PR too...

```ruby
irb(main):005> other_io = IO.new(@io.fileno)
=> #<IO:fd 5>
irb(main):006> other_io.autoclose?
=> true
irb(main):007> other_io.autoclose = false
=> false
irb(main):008> other_io.closed?
=> false
irb(main):009> other_io.close
=> nil
irb(main):010> other_io.closed?
=> true
```

seven1m:
Well dang that change breaks all kinds of stuff in other specs, perhaps related to pipes. I think I'll just ship this as-is and we can circle back on this later.

PR #1456, opened 2023-11-08
Convert flags_struct into a real object

### LINKED ISSUE #1247: Cleanup of construction of flags_struct

herwinw:
This closes #1247 

This is one of those chores that had to be done some day, and it was getting a bit into the way of #1395.

### DIFF
--- a/include/natalie/ioutil.hpp
+++ b/include/natalie/ioutil.hpp
@@ -26,15 +26,8 @@ namespace ioutil {
         StringObject *path() const { return m_path; }
         bool autoclose() const { return m_autoclose; }
 
-        // NATFIXME: This should be made private, but we have to shave some yaks first
-        HashObject *m_kwargs { nullptr };
-        bool m_has_mode { false };
-        int m_flags { O_RDONLY | O_CLOEXEC };
-        read_mode m_read_mode { read_mode::none };
-        EncodingObject *m_external_encoding { nullptr };
-        EncodingObject *m_internal_encoding { nullptr };
-
     private:
+        void parse_flags_obj(Env *, Value);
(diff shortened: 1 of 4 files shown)


herwinw:
Note to self: make `flags_struct` a subclass of `Cell` and implement the GC visitor. And I think the gc_lint check should checks structs as well.

seven1m:
If we don't attach a `flags_struct` to another object, then I don't think we need to make it managed by the GC do we? Is there a case where `flags_struct` will live longer than a single call frame?

herwinw:
Probably not (although I'm not sure when the GC kicks in, I assume only when we leave a scope / pop an env).

I tried to exclude this class, but that still failed and I don't really know why. It doesn't really help that I run Clang 16 on my machine, which seems to break the gc_lint script, so I cannot run it locally.
If anybody else wants to have a try: just convert `flags_struct` to a class and put `public:` as the first line.

seven1m:
Sorry I didn't have time to look at this. If you want to ship this as-is and make a new issue, I can circle back to the struct/class thing later. I don't think it's a big hangup for me that it is a `struct` for now.

herwinw:
I'm only like 30% finished, so I don't wan to ship it as it is now :wink:

herwinw:
And this should fix the remaining 70%. I'll leave the conversion to a class and the changes to gc_lint as an exercise to the reader.

seven1m (approved):
:clap:

Issue #142, opened 2021-09-12
Issue when using the REPL with Fibers


richardboehme:
When working on `Enumerator::Lazy` I noticed, that the REPL does not support Fibers. The following Error will occur:

```
nat> [1,2,3].to_enum.each {}
ruby: /home/richard/repos/natalie/include/natalie/fiber_value.hpp:120: void Natalie::FiberValue::create_stack(Natalie::Env*, int): Assertion `comparison_ptr < Heap::the().start_of_stack()' failed.
```

I already debugged this and found the issue. The start of the stack (set by `Heap::the().set_start_of_stack()`) will not be set because it is being set in the main function (`src/main.cpp`) which is not invoked when using the REPL.

I'm not sure about how we can fix this because I'm a bit confused about the whole start_of_stack-thingy. I tried using a new extern function `setup_heap` and to call it with a `Fiddle::Function` from the REPL. The problem is that I do not know which address I should pass to this function. Is this the correct way or are we able to use a fixed address for the start of the stack instead of the arguments passed to the program?

```cpp
extern "C" void setup_heap(void *address) {
    Heap::the().set_start_of_stack(address);
}
```

seven1m:
Good find. We have to keep track of the actual stack start that was allocated to the program, so it cannot be fixed. I'm not sure exactly what we need to do, but I can investigate later...

richardboehme:
Mh.. If I understand correctly, the start of the stack needs to be set to the current top of the stack, so we need to change it each time a new operation will be run by the REPL, right? If this is correct we could expose the method `setup_heap` and change the start of the stack each time to the current stack address of the program. As far as I know there is no (portable?) method to determine the current stack address...

seven1m:
Actually I think when calling a function via Fiddle, the Ruby process' stack is utilized (could be wrong of course). So I think we would just set it up when the REPL process first runs. But how do we get the actual start from Ruby?? 🤔

richardboehme:
Just a thought: Could we fix this by implementing the REPL in C++ or is there something that speaks against that? When doing so, we could easily get the stack address, right?

seven1m:
That's a good point! I've been wanting to do that for awhile. Just needed our own Fiddle library, which I made recently. 🎉

PR #1460, opened 2023-11-11
Use Eduardo Bart's minicoro library for Fibers

seven1m:
This lets us remove some bespoke x64 and aarch64 assembly in favor of a library that is battle-tested. Also, the library already had support for AddressSanitizer, so I was able to build off of that for #1455.

https://github.com/edubart/minicoro

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "ext/prism"]
 	path = ext/prism
 	url = https://github.com/ruby/prism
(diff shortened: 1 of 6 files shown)


herwinw:
This resolves the error in #142, but the bad news is that the error is now  replaced by a segfault :grin:

seven1m:
Yeah I don't think that issue is resolvable the way our REPL works currently. I have plans to rework the REPL to fix that...

PR #1461, opened 2023-11-11
Start with YAML.dump

herwinw:
This is a very coarse start for the YAML module. It can dump strings, and that's about it.


spec/library/yaml/dump_spec.rb
```diff
@@ -0,0 +1,70 @@
+require_relative '../../spec_helper'
+require_relative 'fixtures/common'
+
+# TODO: WTF is this using a global?
```
seven1m, line 4: LOL

seven1m (approved):

PR #1471, opened 2023-11-13
Bytecode for the VM

seven1m:
This is just a start. Lots of instructions don't yet have bytecode representation yet.


lib/natalie/compiler/instructions/meta.rb
```diff
@@ -0,0 +1,103 @@
+module Natalie
+  class Compiler
+    INSTRUCTIONS_META = {
+      alias_global:              { num: 1 },
+      alias_method:              { num: 2 },
```
herwinw, line 5: I think this could use a bit of documentation on how to add the next instruction. Currently both the names and the numbers are ordered. Unless the next three instructions are going to be called `zumbooruk`, `zymotic` and `zztop`, we are not going to keep it this way.

Adding the next instruction ordered by name means the numbers are going to get mixed up, which means it will be hard to find the next available number. Reassigning the numbers looks tidy, but it means our bytecode output is not backwards compatible.

seven1m: Yeah good point. I'm considering adding a preamble and version number at the beginning of the bytecode so we can detect changes and reject them outright. Changes to the instruction set should probably break existing bytecode anyway.

...And once that is in place, I think I can remove the explicit numbering and just rely on auto numbering from top-to-bottom, alphabetically.

seven1m:
I kinda lost motivation to work on this (for now!), so I'm going to merge what we have before it gets too stale.

PR #1474, opened 2023-11-13
Refresh specs for Kernel.Integer

herwinw:
And remove our local test for a base argument that cannot be coerced into an integer, these have been accepted upstream (including version guards and a spec for the fix in Ruby 3.3).


herwinw:
I tried running our test suite with ruby-3.0-preview3, a few of our local specs fail. I'm trying to get those tests into the upstream specs with version guards, so we can at Ruby 3.3 to the CI as soon as it's released.

seven1m (approved):

PR #1489, opened 2023-11-15
Add Threads

seven1m:
This PR adds basic `Thread` support by using fine-grained locks around our internal data structures.

Well, some of the locks could be even finer. For example, I lock all Array gets and sets with a single lock, whereas we could store a lock on each Array itself. But our locks are still much finer than a GIL (or GVL as Ruby calls it), which should allow for proper parallelism.

Here is a stupid stress test to compare CRuby to Natalie. Notice that with Natalie I'm able to max out all my logical cores...

Ruby:

Screenshot 2023-11-27 7 50 04 AM

Natalie:

Screenshot 2023-11-27 7 54 25 AM

----

This of course does not guarantee thread safety in user code, but neither does Ruby.

I only implemented enough of the Thread methods to get something working. They are certainly not all correct and some are just stubs that do nothing, like `Thread.pass` and `Thread#wakeup`.


include/natalie/macros.hpp
```diff
@@ -50,27 +50,36 @@
...
-                    "Be sure to heap-allocate the object with `new`.",                        \
-                    ptr);                                                                     \
-                abort();                                                                      \
-            }                                                                                 \
-        }                                                                                     \
+#define NAT_GC_GUARD_VALUE(val)                                                                        \
+    {                                                                                                  \
+        NAT_GC_LOCK_GUARD();                                                                           \
+        Object *ptr;                                                                                   \
+        /* FIXME: we should be able to have this guard on all threads */                               \
```
seven1m, line 57: I definitely want to fix this before merge.

seven1m:
Yay CI is finally passing! 🎉 

There is still at least one flaky test I am working on...

seven1m:
I can keep hacking on this, but it's nearing 1000 lines of new code, and the CI seems to be consistently passing, so I'm just going to pull the trigger. There's much to improve and fix here, but at least it's a good start for Natalie. ❤️

PR #1491, opened 2023-11-16
Implement $+

herwinw:


herwinw:
This should resolve a few compile errors in the nightly specs (library/defined, library/predefined and English)

PR #1499, opened 2023-11-17
Add Digest::Instance, implement Digest.file

herwinw:


lib/openssl.rb
```diff
@@ -4,6 +4,29 @@
 __ld_flags__ '-lcrypto'
 __ld_flags__ '-lssl'
 
+# We have some circular dependencies here: our digest implementations are simply aliases to OpenSSL::Digest classes,
+# but OpenSSL::Digest depends on ::Digest::Instance. So define this one in openssl.rb instead op digest.rb
```
seven1m, line 8:
```suggestion
# but OpenSSL::Digest depends on ::Digest::Instance. So define this one in openssl.rb instead of digest.rb
```

herwinw: Fixed that one

seven1m (approved):

PR #1500, opened 2023-11-17
[VM] speed up BER decoding

herwinw:
The general case will only decode a single byte. Delay the right shift until we actually need it.


seven1m (approved):

herwinw:
Also: #1500 :partying_face:

PR #1503, opened 2023-11-18
Implement Digest#length and Digest#size

herwinw:
Which required some updates to the constants of the digests to mark them as binary (we don't use the magic comment with file encoding), because it was calculating the size wrong. This showed up that the digests we were returning were not binary either.


seven1m (approved):
Yeah, encodings... 😬 Gotta work on that.

PR #1506, opened 2023-11-18
Implement YAML.load

herwinw:


herwinw:
This should be good enough to be useful now.

PR #1507, opened 2023-11-18
Use Ubuntu 22.04 in actions

### LINKED ISSUE #1472: Upgrade clang-tidy and fix warnings
We currently run clang-tidy in the Ubuntu 20.04 image on GitHub Actions, which means it's kinda old. Newer versions of clang-tidy have been known to catch things in our code, e.g. #1462.

It would be cool to:

1. Run `rake docker_tidy` locally and fix any warnings.
2. Have our CI (GitHub Actions) run our `rake docker_tidy` task. I believe this will let us run a newer version of clang-tidy. (If not, maybe see about what it would take to upgrade our Docker image to newer Ubuntu version.)

herwinw:
Which will probably not fix the issue in #1472, it's the latest linux runner but still over a year old.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   clang-tidy:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v1
       - name: checkout submodules
@@ -20,7 +20,7 @@ jobs:
       - name: run clang-tidy
         run: rake docker_tidy
   gc-lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v1
       - name: checkout submodules
--- a/.github/workflows/run_all_specs.yml
+++ b/.github/workflows/run_all_specs.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   specs:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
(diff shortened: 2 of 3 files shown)


herwinw:
https://github.com/kgorking/ecs/blob/1dd4400589b0d7e2bd9a1aed147890f566041648/.github/workflows/clang.yml#L32-L39

I guess we could use this to run a newer toolchain.

herwinw:
https://github.com/natalie-lang/natalie/actions/runs/6915553353/job/18814621284?pr=1507#step:4:109

Yeah, this still uses 14

herwinw:
> https://github.com/kgorking/ecs/blob/1dd4400589b0d7e2bd9a1aed147890f566041648/.github/workflows/clang.yml#L32-L39
> 
> I guess we could use this to run a newer toolchain.

Yeah, that's not going to work if we put our own Docker image in between.

seven1m:
Dang, we were already running `rake docker_tidy`! I don't know why I thought we were using an off-the-shelf GitHub action for that. My bad.

I think the changes here are good to just freshen up the runner anyway.

seven1m:
...but now I'm confused. I wrote #1472 because I thought we had a clang-tidy warning that CI didn't catch. Did I just dream that up?

herwinw:
> ...but now I'm confused. I wrote #1472 because I thought we had a clang-tidy warning that CI didn't catch. Did I just dream that up?

No, that's correct (although the actual issue reported by clang-tidy has been). We run clang-tidy in the `ruby:3.2` image, which is based on Debian Bookworm. This contains clang 14.
My desktop is running clang 16, that's how I found the issue that led to #1472

PR #1508, opened 2023-11-18
Do not lint spec directory with Rubocop

richardboehme:


richardboehme:
@seven1m Also do we intend to lint all other files with rubocop? My editor just started auto correcting all offenses everywhere, so I had to disable the auto correction for now. Maybe we want to consequently lint all *.rb files that aren't from a third party in a Github Action?

seven1m:
I think it's fine to just do the `lib` and `src` directory for now? I'm open to suggestion here.

PR #1517, opened 2023-11-20
Switch nightly spec runner back to Ubuntu 20.04

herwinw:
[The current CI runs are crashing](https://github.com/natalie-lang/natalie/actions/runs/6924196633/job/18838675024)
I'm not sure if this is related to the Ubuntu version or something else, but it is kind of suspicious that it started crashing the day we bumped the version.

### DIFF
--- a/.github/workflows/run_all_specs.yml
+++ b/.github/workflows/run_all_specs.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   specs:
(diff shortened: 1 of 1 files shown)


herwinw:
Okay, I'm just going to be very rude and merge this one without a review, so I can retry that action runner and see if this was the issue, or if something else is broken.

PR #1526, opened 2023-11-26
Fix Regexp code generation for embedded `#` character

herwinw:
This is the second yak I had to shave to get the `uri` gem to work. With this it's possible to compile the gem. It still doesn't work, so there's still a herd of hairy yaks waiting.

### DIFF
--- a/lib/natalie/compiler/instructions/push_regexp_instruction.rb
+++ b/lib/natalie/compiler/instructions/push_regexp_instruction.rb
@@ -12,7 +12,8 @@ module Natalie
       end
 
       def generate(transform)
-        transform.exec_and_push(:regexp, "Value(RegexpObject::literal(env, #{@regexp.source.inspect}, #{@regexp.options}))")
+        source = @regexp.source.inspect.gsub(/\\#(\$|@)/, '#\1')
+        transform.exec_and_push(:regexp, "Value(RegexpObject::literal(env, #{source}, #{@regexp.options}))")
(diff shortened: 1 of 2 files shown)


lib/natalie/compiler/instructions/push_regexp_instruction.rb
```diff
@@ -12,7 +12,8 @@ def to_s
       end
 
       def generate(transform)
-        transform.exec_and_push(:regexp, "Value(RegexpObject::literal(env, #{@regexp.source.inspect}, #{@regexp.options}))")
+        source = @regexp.source.inspect.gsub(/\\#(\$|@)/, '#\1')
```
seven1m, line 15: I think there was a bug with Prism's `RegexpNode#unescaped` method so I didn't use it here. We should upgrade Prism and see if we can get this fix for free. I can look at that later today!

herwinw: That might work as well. We've got some test cases for it.

seven1m: Turns out it wasn't related to the Prism thing ... but I did remember we already have a `StringToCpp` module that handles this escaping for us, so I added a commit to your PR.

herwinw: Yeah, that looks so much better.

PR #1529, opened 2023-11-28
Improve Module#const_defined? and Module#const_get

herwinw:
Add the `inherited` argument. Two more yaks for the URI gem.


herwinw:
This combined with the added `String#delete` feels like we're getting somewhere:
```
bin/natalie -ruri -e 'p URI("https://natalie-lang.org")'
#<URI::HTTPS https://natalie-lang.org>
```

PR #1532, opened 2023-11-28
Hash key shorthand

seven1m:
This compiles the hash key shorthand, like this:

```ruby
foo = 1
h = { foo: } # => { foo: 1 }
bar(foo:) # => bar(foo: 1)
```


seven1m:
Dang I just realized this will fail CI when hosted with Ruby 3.0. Maybe I'll wait to merge this after Ruby 3.3 is released and we drop 3.0 from CI.

herwinw:
The diff check on the test/natalie folder only checks the output, so something like this should make it pass:
```ruby
ruby_version_is ...'3.1' do
  it 'does not accept hash key shorthand' do
    b = 2
    method_with_kwargs1(1, b: b).should == [1, 2]
  end
end

ruby_version_is '3.1' do
  it 'accepts hash key shorthand' do
    b = 2
    method_with_kwargs1(1, b:).should == [1, 2]
  end
end
```

seven1m:
Welp that doesn't work because Ruby 3.0 cannot even parse the file. I get the error:

```
test/natalie/hash_test.rb:25: syntax error, unexpected ','
        h = { foo:, bar: }
```

I'll just wait till we're done with 3.0.

herwinw:
I managed to fix these issues by extracting the code into separate files, end using `ruby_exe` on the correct file.

herwinw:
Captain hindsight: `eval` would probably have worked too.

PR #1535, opened 2023-11-28
Import ipaddr gem

herwinw:
This is a direct copy from the source, it now works with Natalie. The source file has been added to the copyright list (similar to `pp.rb`)


seven1m:
Cool! 🎉

PR #1550, opened 2023-12-03
Fix Sequence behaviour of Array#[]

herwinw:
#1543 has been written with an older local version of the spec. The latest upstream version contains a few specs that trigger assertion failures in our code.

### DIFF
--- a/src/array_object.cpp
+++ b/src/array_object.cpp
@@ -2000,7 +2000,7 @@ Value ArrayObject::slice_in_place(Env *env, Value index_obj, Value size) {
                 idx += step;
             }
         }
-        m_vector = result;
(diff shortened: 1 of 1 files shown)


herwinw:
To make things more complicated: this bug trigger is not stable: When I try to isolate the `it` block that triggers the bug, it simply bails out with a failed spec. Running the whole spec file does trigger the bug most times, but it does not crash when running in GDB.

The only thing I found to get a backtrace in GDB is to compile `spec/core/array/element_reference_spec.rb` into an executable, use `ulimit -c` to save core dumps, run the executable a few times until it crashes, then load the executable and the core dump in GDB.

This feels like memory corruption, but I'm not sure yet where it happens.

PR #1556, opened 2023-12-06
Encoding fixups

herwinw:
It turned out every "Conversion above Unicode Basic Latin" comment in the specs folder was outdated, and a lot of our own tests included small mistakes.

### DIFF
--- a/spec/core/string/shared/chars.rb
+++ b/spec/core/string/shared/chars.rb
@@ -21,7 +21,7 @@ describe :string_chars, shared: true do
   end
 
   it "returns characters in the same encoding as self" do
-    "&%".force_encoding('ASCII-8BIT').send(@method).to_a.all? {|c| c.encoding.name.should == 'ASCII-8BIT'}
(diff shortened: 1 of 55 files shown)


spec/core/string/shared/chars.rb
```diff
@@ -57,10 +57,12 @@
       [0xAD].pack('C').force_encoding('BINARY'),
       [0xA2].pack('C').force_encoding('BINARY')
     ]
-    #s.force_encoding('SJIS').send(@method).to_a.should == [
-      #[0xF0,0xA4].pack('CC').force_encoding('SJIS'),
-      #[0xAD].pack('C').force_encoding('SJIS'),
-      #[0xA2].pack('C').force_encoding('SJIS')
-    #]
+    NATFIXME 'Implement SJIS encoding', exception: ArgumentError, message: 'unknown encoding name - "SJIS"' do
```
herwinw, line 60: To save everybody else the need to look it up: `SJIS` is not the same as `Shift-JIS`, but an alias for `Windows-31J`

PR #1557, opened 2023-12-06
Add ISO-8859 encodings

herwinw:
I'm going to make this more generic as soon as I've added the second encoding.

The framework in `test/encodings/eucjp_tables.rb` is pretty useful here, this saves like 90% of the boring work :partying_face:


herwinw:
Stats: 1004 insertions, 1071 deletions, which includes 669 new lines of specs.

https://encoding.spec.whatwg.org/#legacy-single-byte-encodings has some more single byte encodings that can probably be added this way, but I'm kind of bored of these changes for now.

seven1m:
We have some flaky tests that I'll have to look at later. This looks fantastic.

PR #1569, opened 2023-12-07
Fix MatchData#[] with Range argument and nil values in the matches

herwinw:
Another URI yak.


herwinw:
Sneak preview: the next yak is in `MatchData#[]` as well. I feel like these URI specs are better specs for MatchData than the actual MatchData specs ;)
I'm going to save the new specs until the URI gem passes its specs, so I can push them all upstream in 1 go.

PR #1571, opened 2023-12-08
Pattern matching assoc

seven1m:
This is the very very beginning work on pattern matching. It does just a few small things:

```ruby
[1, 2] => [1, a]; a
# 2

{ a: 1, b: 2 } => { a: 1, b: b }; b
# 2

[1, 2, 3] => [x, y]
# => length mismatch (given 3, expected 2) (StandardError)
```


seven1m:
I don't really like these new instructions, because they won't work well within a `case ... in` expression. I'm going to think about this some more... :thinking:

seven1m:
Going back to the drawing board on this for now!

PR #1572, opened 2023-12-09
Fix behaviour with range in MatchData#[] when first result is nil

herwinw:
Just one more yak remaining for the uri gem, and that one is not related to matchdata.


seven1m (approved):

seven1m:
If the build fails on macOS for the same reason it's been failing, feel free to go ahead and merge. I haven't had time to look into that yet.

herwinw:
Yes, it's still the openssl specs.
I think they started around the time I've added the ISO-8859 encodings, although I don't really understand how that can be related.

PR #1574, opened 2023-12-09
Import URI gem

herwinw:


herwinw:
So there is one failing spec file left (but also: more than half of the specs are just placeholders, so just adding an empty `lib/uri.rb` would make those pass).

Also: I'll be away next week, so it might take a while for me to get back to this.

herwinw:
And this one should be ready for release. I had to make a small alteration to some named regexp captures that were refused by Onigmo (I guess MRI uses a forked version that supports uppercase characters in capture names).

PR #1576, opened 2023-12-09
Always reset digest

### LINKED ISSUE #1575: OpenSSL Digest reset issue on macOS (OpenSSL 3.2?)
I suspect this is due to some change in OpenSSL 3.2 ... but it happens regularly in our CI and on my local Mac. See failed build here: https://github.com/natalie-lang/natalie/actions/runs/7152408867/job/19477709455

I did some debugging and found that I can get this test to pass:

https://github.com/natalie-lang/natalie/blob/master/spec/library/digest/sha512/hexdigest_spec.rb

...if I comment out this line in the spec:

https://github.com/natalie-lang/natalie/blob/master/spec/library/digest/sha512/hexdigest_spec.rb#L8

```ruby
cur_digest = Digest::SHA512.new

# Commenting this line fixes the spec:
cur_digest.hexdigest.should == SHA512Constants::BlankHexdigest

# Maybe need to reset something before we can call this again...
cur_digest << "test"
```

...or if I comment out this line in our implementation:

https://github.com/natalie-lang/natalie/blob/master/lib/openssl.cpp#L264

```cpp
if (args.size() == 1)
    OpenSSL_Digest_reset(env, self, {}, nullptr);

// changing to this works:
OpenSSL_Digest_reset(env, self, {}, nullptr);
```

In both cases, it seems that after we call `EVP_DigestFinal_ex`, we should be calling `OpenSSL_Digest_reset` or perhaps `EVP_MD_CTX_cleanup` according to Google.

I don't know enough about how OpenSSL works to know which is correct. @herwinw any thoughts?

seven1m:
This seems to fix #1575, at least for me, on my Mac with OpenSSL 3.2.0.

I'll let @herwinw decide if this is the proper fix. ❤️

### DIFF
--- a/lib/openssl.cpp
+++ b/lib/openssl.cpp
@@ -261,8 +261,7 @@ Value OpenSSL_Digest_digest(Env *env, Value self, Args args, Block *) {
     if (!EVP_DigestFinal_ex(mdctx, buf, &md_len))
         OpenSSL_raise_error(env, "EVP_DigestFinal_ex");
 
-    if (args.size() == 1)
-        OpenSSL_Digest_reset(env, self, {}, nullptr);
+    OpenSSL_Digest_reset(env, self, {}, nullptr);
 
     return new StringObject { reinterpret_cast<const char *>(buf), md_len, EncodingObject::get(Encoding::ASCII_8BIT) };
 }


herwinw:
Issue has been created in #1504 (but that was before we were testing with this version of OpenSSL, so that's why the CI didn't see the issue)

herwinw:
I think this change is good enough for now. It still has some issues, but at least this fixes the annoying CI issue.

The Digest class has an incremental interface using `<<` or `update`. Some experiments when using MRI (using MD5 digest to keep the output relatively small):
```
irb(main):022:0> d.reset
=> #<OpenSSL::Digest: d41d8cd98f00b204e9800998ecf8427e>
irb(main):023:0> d << 'foo'
=> #<OpenSSL::Digest: acbd18db4cc2f85cedef654fccc4a4d8>
irb(main):024:0> d << 'bar'
=> #<OpenSSL::Digest: 3858f62230ac3c915f300c664312c63f>
irb(main):025:0> d.hexdigest
=> "3858f62230ac3c915f300c664312c63f" # == MD5('foobar')
irb(main):026:0> d.hexdigest
=> "3858f62230ac3c915f300c664312c63f" # == MD5('foobar')
irb(main):027:0> d << 'foo'
=> #<OpenSSL::Digest: c98cbfeb6a347a47eb8e96cfb4c4b890>
irb(main):028:0> d.hexdigest
=> "c98cbfeb6a347a47eb8e96cfb4c4b890" # == MD5('foobarfoo')
irb(main):029:0> d.hexdigest('foobarfoo')
=> "c98cbfeb6a347a47eb8e96cfb4c4b890"
```
So not every call to `hexdigest` (or `digest`, but the hex output is a bit easier too read) resets the state. Compare this to Natalie (with this change):
```
nat> d << 'foo'
#<OpenSSL::Digest:0x55738951bd60>
nat> d.hexdigest
"acbd18db4cc2f85cedef654fccc4a4d8" # == MD5('foo')
nat> d.hexdigest
"d41d8cd98f00b204e9800998ecf8427e" # == MD5('')
```
But honestly, master is even worse:
```
nat> d << 'foo'
#<OpenSSL::Digest:0x5560d62a3d60>
nat> d.hexdigest
"acbd18db4cc2f85cedef654fccc4a4d8"
nat> d.hexdigest
"182ca6a7967906c0816bf82e83a65c26"
nat> d.hexdigest
"c578759acccb127a5834a0ced775523e"
```
I have no idea what those last two values are supposed to be.

herwinw (approved):

seven1m:
Ahhh, good digging. Thanks!

Issue #1579, opened 2023-12-10
Return value of attr_writer like methods should be ignored


herwinw:
This is an issue found in  #1574:
```ruby
class Foo
  attr_reader :bar

  def bar=(value)
    @bar = value
    return 'CANARY'
  end
end

foo = Foo.new
x = foo.bar = :baz
p x
```
This prints `:baz` in MRI, but `CANARY` in Natalie.

PR #1581, opened 2023-12-10
Stringification of Regexp

herwinw:
This should be the last yak for the URI gem (I'm going to ignore #1579 for now, that issue is only related to the kind of weird/irrelevant specs of this module). It does need some more specs though.


herwinw:
This should be it, it has some additional tests, and the code got simplified too

PR #1598, opened 2023-12-21
Implement ConditionVariable

herwinw:


herwinw:
Pushing this early to clarify that I'm currently working on this. When we've got this one, the implementation of `Queue` should be very trivial. Once we've got the `Queue` class, we can import the `timeout` gem (probably, we'll probably see more road blocks), which is a requirement for the `net` gem.

seven1m:
🐢 
 🐢 
  🐢
   🐢 

...down you go 😆

herwinw:
I'm going to close this Pull Request. The `std::condition_variable` implementation of C++ requires a `std::unique_lock` of a mutex, that is not really accessible from this point in the code without breaking all the RAII setup of mutex we already have.
I might retry this, but that one will have a totally different setup, so this version is pretty useless now.

PR #1615, opened 2023-12-25
Switch to Ruby 3.3 compatibility

herwinw:
This currently results in a lot of failing tests, and the CI does not run yet either.


herwinw:
@seven1m Would you be able to take a look at this failing spec from `test/natalie/proc_test.rb`?
```ruby
ruby_version_is '3.3' do
  it 'cannot create a new Proc object from a block argument' do
    def create_lambda(&block)
      lambda(&block)
    end
    -> { create_lambda { 1 } }.should raise_error(ArgumentError, 'the lambda method requires a literal block')
  end
end
```
This one should have been added in commit 90163209c67ace8d4811bb81e93b00b4e625a1e7, but when I try to revert that fix everything in the compiler breaks down.

herwinw:
It still has a bunch of NATFIXMEs (I didn't really want to fix every single issue here), but I think this is good enough to start with Ruby 3.3 compatibility.

PR #1616, opened 2023-12-25
Add FileUtils.mkdir_p

seven1m:


lib/fileutils.rb
```diff
@@ -14,4 +14,22 @@ def self.rm_rf(path)
       File.unlink(path)
     end
   end
+
+  def self.mkdir_p(path, mode: 0775)
+    parts = File.expand_path(path).split(File::SEPARATOR)
+
+    dir = parts.shift
+    parts.each_with_index do |part|
```
herwinw, line 22: I guess `parts.each` would be better here, since we don't use the index argument

PR #1622, opened 2023-12-28
Update nightly all-specs runner to use Ubuntu 22.04 and to run in Docker

### LINKED ISSUE #1518: Issues with nightly specs and Ubuntu 22.04 action runner
https://github.com/natalie-lang/natalie/actions/runs/6924196633/job/18838675024

It looks like the upgrade of the spec runner to ubuntu-22.04 broke things. I have tried retrying the specs, and they broke at the same location.

I  have reverted the change to the nightly runner ( #1517), which serves as a band aid and things are running again, but it would be nice to actually fix the issue that causes it. To do that, we have to find that issue first.

seven1m:
This should fix #1518.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
           - gcc
           - clang
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


seven1m:
I guess we'll find out later if that works...

PR #1625, opened 2023-12-29
Implement SignalException

herwinw:


src/natalie.cpp
```diff
@@ -511,6 +511,18 @@ void handle_top_level_exception(Env *env, ExceptionObject *exception, bool run_e
...
+        if (signo->type() == Object::Type::Integer) {
+            nat_int_t val = signo->as_integer()->to_nat_int_t();
+            if (val >= 0 && val <= 255) {
+                clean_up_and_exit(val);
+            } else {
+                clean_up_and_exit(1);
+            }
+        } else {
+            clean_up_and_exit(1);
+        }
```
seven1m, line 525: I think this part is incorrect. When I run CRuby with gdb, the program actually terminates with the signal from the SignalException, which is what I would expect.

Screenshot 2023-12-30 7 24 00 AM

seven1m: Maybe we need to `::kill(::getpid(), signal)` or some such?

seven1m: After thinking about this some more, I think what you have here is fine for now. I have in mind to work on signal handling soon anyway, so I can work on it then...

herwinw: I just pushed a fix to get the correct exit value. It should be good enough to go now.
This change is very much a chicken and egg with Process.kill and Signal.trap, I guess we have to start somewhere.

seven1m (approved):

PR #1630, opened 2023-12-30
Implement Thread#native_thread_id

herwinw:


src/thread_object.cpp
```diff
@@ -106,6 +106,18 @@ bool ThreadObject::is_stopped() const {
     return m_sleeping || m_status == Status::Dead;
 }
 
+Value ThreadObject::native_thread_id(Env *env) const {
+    if (is_stopped()) return NilObject::the();
+#ifdef __APPLE__
+    uint64_t id;
+    if (!pthread_threadid_np(m_thread_id, &id))
+        env->raise_errno();
+    return Value::integer(id);
```
seven1m, line 115: Can we just return the `m_thread_id` here on macos too? Why do we need to get a different ID for macOS here?

herwinw: If only things were that easy.

The `pthread_t` data type should be treated as an opaque value, on Linux it's simply an int, but MacOS appears to be using a more complex value. The previous test result is in https://github.com/natalie-lang/natalie/actions/runs/7365511535/job/20046890959:
```
src/thread_object.cpp:111:27: error: static_cast from 'pthread_t' (aka '_opaque_pthread_t *') to 'Natalie::nat_int_t' (aka 'long long') is not allowed
    return Value::integer(static_cast<nat_int_t>(m_thread_id));
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I expected this change to be some low hanging fruit, but it looks like it isn't. I'm thinking about closing this PR, since I don't think many people are going to use this method, so it feels like the amount of effort required to get this working is not proportional to the expected usage.

seven1m: A C-style cast `(long long)` works. 😬

herwinw: That's just the address of the pointer on MacOS. It might work, so let's just try that.

herwinw: And now the Linux specs break too. I'm going to close this PR.

PR #1633, opened 2023-12-31
Thread#raise

seven1m:
~I'll work on the remaining skipped specs here before merging.~

~Hopefully the dust settles on #1631 and #1632 by then.~

~I need to figure out how to raise when a socket is waiting to connect.~ 

The tests pass yay 🎉


herwinw:
As a precaution: the changes in  #1632 still have a few NATFIXME statements left in the shared raise file, so either the implementation of `Kernel.raise` and `Thread#raise` should be kept in sync including the bugs and missing features, or we need a way to selectively disable NATFIXME blocks. Something like:
```ruby
NATFIXME 'lorum ipsum', condition: @object.is_a?(Thread) do
  # specs that fail on Thread#raise, but pass on Kernel.raise
end
```

lib/natalie/compiler/binding_gen.rb
```diff
@@ -1309,7 +1309,9 @@ def generate_name
 gen.static_binding('Thread', 'exit', 'ThreadObject', 'exit', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.static_binding('Thread', 'list', 'ThreadObject', 'list', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
 gen.static_binding('Thread', 'main', 'ThreadObject', 'main', argc: 0, pass_env: false, pass_block: false, return_type: :Object)
-gen.static_binding('Thread', 'pass', 'ThreadObject', 'pass', argc: 0, pass_env: false, pass_block: false, return_type: :Object)
+gen.static_binding('Thread', 'pass', 'ThreadObject', 'pass', argc: 0, pass_env: true, pass_block: false, return_type: :Object)
+gen.static_binding('Thread', 'default_report_on_exception', 'ThreadObject', 'default_report_on_exception', argc: 0, pass_env: false, pass_block: false, return_type: :bool)
+gen.static_binding('Thread', 'default_report_on_exception=', 'ThreadObject', 'set_default_report_on_exception', argc: 1, pass_env: false, pass_block: false, return_type: :bool)
```
herwinw, line 1314: The Ruby methods are named `report_on_exception` and `report_on_exception=`, without the `default_` prefix.

PR #1636, opened 2023-12-31
Add cause object to exceptions without explicit raise

herwinw:
Just move the code around a bit and we've fixed another issue ;)


seven1m (approved):
🎉

PR #1649, opened 2024-01-02
Implement Kernel#caller_locations

herwinw:
This includes the new class `Thread::Backtrace::Location`

### DIFF
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -105,6 +105,7 @@
 #include "natalie/string_object.hpp"
 #include "natalie/string_upto_iterator.hpp"
 #include "natalie/symbol_object.hpp"
+#include "natalie/thread/backtrace/location_object.hpp"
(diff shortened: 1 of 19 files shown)


herwinw:
Even though it looks like all but one specs are passing, this actually breaks the REPL with an assertion error.

seven1m (approved):

herwinw:
I would prefer to have a new review/second opinion for this change. The CI kept failing in the MacOS run. I solved it by adding https://github.com/natalie-lang/natalie/pull/1649/commits/5322e05aed68e4a7d7f297632c7a07d00cbf5329, but that is a terrible hack and doesn't solve the issue. But since Linux works just fine, I haven't been able to reproduce the failure so debugging it is kind of hard.

seven1m:
I think that works for now. I can investigate the macos bug later!

PR #1663, opened 2024-01-05
wip: signal threads and waiting IO

seven1m:
We need a better way to interrupt blocking IO. Currently the solution has been to do a tight loop with a 100 nanosecond select(2), checking if the thread should interrupt each time.

But it seems the current state of the art is to set up a pipe that we can add to the select() and then write a byte to that pipe, waking up the blocking call.

This is my first naive attempt, which naturally has problems. It sets up a pipe for every file descriptor and another pipe for every thread. This worked ok for small processes, but for a few tests with lots of IO, we hit the per-process open files limit.

This work will be redone, but I wanted to record this attempt, for future reference. :-)

### DIFF
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -206,4 +206,6 @@ void dbg(const char *fmt, Args... args) {
     puts(out->c_str());
 }
 
+int pipe2(int pipefd[2], int flags);
+
 }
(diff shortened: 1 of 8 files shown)


seven1m:
My next attempt will be to set up a single pipe that wakes up all blocking threads....

PR #1664, opened 2024-01-05
Set file/line information when creating a lambda/proc

herwinw:
This resolves some issues with the source location specs.

Just for fun, try this script on the current master branch:
```ruby
x = -> {}
p x.source_location
```
Because we never set the environment info before creating this lambda, the last info is used, so this is shown as if it were defined in `src/warning.rb`


seven1m (approved):
In the future I think I would like to somehow give every instruction this location info automagically so we don't have to manually pass it in all over the place. I haven't decided how to do that yet...

PR #1674, opened 2024-01-07
Implement stop-the-world for GC

seven1m:
Originally I intended to just fix `ensure` blocks in threads. From there my life spiraled and I learned more about Linux signals and processor state than I ever wanted to know. 😆 

This PR does a lot of things -- too many actually -- but the headline feature is a stop-the-world workflow for our GC so that thread state is always in a stable place when we do our conservative stack and context scan.

I found and fixed lots of edge cases around thread birth and death, too.

The secondary feature is that `Thread#raise` now works as expected, by raising an exception at the point in the stack where the thread is running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG IMAGE=ruby:3.3
 FROM $IMAGE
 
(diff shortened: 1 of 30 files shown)


seven1m:
OMG it's a miracle. I'm going to run CI again just to be sure. 😄

PR #1676, opened 2024-01-07
Implement Enumerator.produce

herwinw:
Why have I never heard of this method before? This looks like an amazing generator.

### DIFF
--- a/src/enumerator.rb
+++ b/src/enumerator.rb
@@ -38,6 +38,20 @@ class Enumerator
     end
   end
 
+  def self.produce(*values, &block)
+    if values.size > 1
+      raise ArgumentError, "wrong number of arguments (given #{values.size}, expected 0..1)"
+    end
+
+    new do |yielder|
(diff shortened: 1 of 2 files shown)


seven1m (approved):

src/enumerator.rb
```diff
@@ -38,6 +38,20 @@ def to_proc
...
   end
 
+  def self.produce(*values, &block)
+    if values.size > 1
+      raise ArgumentError, "wrong number of arguments (given #{values.size}, expected 0..1)"
+    end
+
+    new do |yielder|
+      value = values.empty? ? block.call(nil) : values.first
+      while true
```
seven1m, line 48: We could use `loop do` here.

herwinw: Maybe I'm just out of the loop (har har) regarding best practices, but those look pretty much like synonyms to me.

seven1m: Someday we can optimize away the truthyness check, then you'd be correct.

PR #1682, opened 2024-01-09
Add Thread local variables

herwinw:
And extended the specs for those a bit, I'll try to push them upstream later today.


seven1m (approved):
Nice!

herwinw:
For completeness (or closure): https://github.com/ruby/spec/pull/1127 has the upstream specs

PR #1685, opened 2024-01-09
Remove local tests for Thread variables

herwinw:
They crash on Mac. For now the best course of action is just to remove them. The additions to the specs have been proposed as an upstream Pull Request, we should probably wait for them to get accepted before we change anything related to this code.


seven1m:
I'm working on the thread bugs in #1674, where the macos build is currently passing. 🎉 

I'm fine with removing these, but I'm also fine with just ignoring CI errors for threads on macos for now. I'll get it all working eventually. 😅

seven1m (approved):

PR #1696, opened 2024-01-21
Fix missing return check on write() call (and fix missing warning in debug mode)

### LINKED ISSUE #1689: Compiler warning in release mode
```
src/thread_object.cpp:658:12: warning: ignoring return value of ‘ssize_t write(int, const void*, size_t)’

 declared with attribute ‘warn_unused_result’ [-Wunused-result]                                          
  658 |     ::write(s_interrupt_write_fileno, "!", 1);                                                   
      |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

We should be seeing that in debug mode too, so maybe we need to enable `-Wunused-result`?

seven1m:
Fixes #1689

### DIFF
--- a/lib/natalie/compiler/flags.rb
+++ b/lib/natalie/compiler/flags.rb
@@ -10,6 +10,7 @@ module Natalie
 
       DEBUG_FLAGS = %w[
         -pthread
+        -Og
         -g
         -Wall
         -Wextra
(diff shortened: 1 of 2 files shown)


seven1m:
Dang that optimization level shows the warning, but it also triggers other bugs. Closing for now.

PR #1704, opened 2024-01-21
Implement Kernel.test

herwinw:


seven1m:
BTW if you see that one thread_test.rb timeout issue in CI, just ignore. I'll work on fixing that when I get done with the household duties today. 😄

PR #1718, opened 2024-01-24
Support UNIX sockets in BasicSocket#getsockname

herwinw:
The method still skips some information, but this resolves the "NOT YET IMPLEMENTED" panic in the specs of `BasicSocket#connect_address`.
It does include some additional fixes regarding UNIX sockets.

### DIFF
--- a/lib/socket.cpp
+++ b/lib/socket.cpp
@@ -969,6 +969,7 @@ Value Socket_unpack_sockaddr_un(Env *env, Value self, Args args, Block *block) {
     memcpy(&addr, sockaddr->as_string()->c_str(), std::min(sizeof(addr), sockaddr->as_string()->bytesize()));
     if (addr.sun_family != AF_UNIX)
         env->raise("ArgumentError", "not an AF_UNIX sockaddr");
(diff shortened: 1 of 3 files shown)


seven1m (dismissed):
Not sure what's going on with the Mac build.

herwinw:
> Not sure what's going on with the Mac build.

The error message implies that is is something related to an IP socket, even though this change only adds UNIX sockets to that file.

herwinw:
```ruby
    it 'works with the result of BasicSocket#getsockname (which is often smaller than sizeof(struct sockaddr))' do
      addrinfo = Addrinfo.new("\x01\x00/tmp/sock\x00".b)
      addrinfo.afamily.should == Socket::AF_UNIX
    end
```
This includes the binary version of a `struct sockaddr`, I guess that one is different (either in layout or in constants) on Darwin.
I added this test to have something with a short filename to have less bytes than the struct, so I can't just use a random filename here.
I guess I'll just have to disable this spec on darwin.

herwinw:
And still issues, I'm going to split this into smaller commits so we can actually figure out what the issue is

herwinw:
The slight rewrite in  #1735 seems to have done the trick, this is just a small code change and a bunch of additional tests.

PR #1729, opened 2024-01-26
Implement Module#remove_const

herwinw:


src/module_object.cpp
```diff
@@ -293,10 +293,23 @@ Value ModuleObject::const_set(Env *env, Value name, Value val) {
     return const_set(name_as_sym, val);
 }
 
-void ModuleObject::const_remove(SymbolObject *name) {
+void ModuleObject::remove_const(SymbolObject *name) {
     m_constants.remove(name);
 }
 
+Value ModuleObject::remove_const(Env *env, Value name) {
+    std::lock_guard<std::recursive_mutex> lock(g_gc_recursive_mutex);
```
seven1m, line 301: I think this guard should be in the other (lower-level) `remove_const` method, yeah? That was my oversight previously.

seven1m (approved):

src/module_object.cpp
herwinw: That sounds legit, although I have to admit I haven't looked deep enough into the GC to understand why this is needed (I just copy-pasted the line from the other methods related to constants 🫣)

PR #1733, opened 2024-01-26
Implement transform_constant_{and,or}_write_node

herwinw:
This resolves compilation errors with code like this:
```ruby
Foo ||= 1
Foo &&= 2
```

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -844,6 +844,24 @@ module Natalie
         instructions
       end
 
+      def transform_constant_and_write_node(node, used:)
+        instructions = [
+          PushSelfInstruction.new,
+          ConstFindInstruction.new(node.name, strict: false),
+          DupInstruction.new,
+          IfInstruction.new,
+          PopInstruction.new,
+          transform_expression(node.value, used: true),
+          DupInstruction.new,
+          PushSelfInstruction.new,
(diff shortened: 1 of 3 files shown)


herwinw:
Rebased and marked as ready for review. This change does not cover the path constants (like `A::B ||= 1`). When trying those, it's probably easier to start with the `transform_constant_path_and_write_node` operation, since the specs of the `or` equivalent really hit the bug in  #1734

seven1m:
Dang I gotta look at that mac build

lib/natalie/compiler/pass1.rb
```diff
@@ -844,6 +844,44 @@ def transform_class_variable_write_node(node, used:)
...
+        instructions << PopInstruction.new unless used
+        instructions
+      end
+
+      def transform_constant_or_write_node(node, used:)
+        instructions = [
+          IsDefinedInstruction.new(type: 'constant'),
+          PushSelfInstruction.new,
+          ConstFindInstruction.new(node.name, strict: false),
+          EndInstruction.new(:is_defined),
```
seven1m, line 870: I wonder why this one needs the IsDefinedInstruction but the other transform does not 🤔

herwinw: https://kddnewton.com/2023/12/06/advent-of-prism-part-6#constants has some explanation for that

PR #1745, opened 2024-01-31
Implement File.absolute_path?

herwinw:


src/file_object.cpp
```diff
@@ -180,6 +180,11 @@ bool FileObject::exist(Env *env, Value path) {
     return ::stat(path->as_string()->c_str(), &sb) != -1;
 }
 
+bool FileObject::is_absolute_path(Env *env, Value path) {
+    path = ioutil::convert_using_to_path(env, path);
+    return path->as_string()->string()[0] == '/';
```
seven1m, line 185: Should we add a check for zero-length string before we index it?

seven1m (approved):

src/file_object.cpp
herwinw: It's null terminated, so even if it is an empty string, we can still read the first character safely and get a `'\0'`.

PR #1762, opened 2024-02-07
Implement Data#with and Data#to_h

herwinw:


test/natalie/data_test.rb
```diff
@@ -0,0 +1,19 @@
+require_relative '../spec_helper'
+require_relative '../../spec/core/data/fixtures/classes'
+
+ruby_version_is ''...'3.2' do
+  describe 'we need the same number of tests here to make the ruby comparison pass' do
+    it 'transforms the data object into a hash' do
+      1.should == 1
+    end
+  end
+end
```
seven1m, line 10: This is weird, maybe we need our Ruby<->Natalie comparison code [here](https://github.com/natalie-lang/natalie/blob/master/test/support/compare_rubies.rb#L31-L31) to be a little smarter... 🤔

seven1m (approved):

test/natalie/data_test.rb
herwinw: I think that is going to be a mess. When we compare the output of Ruby 3.1 and Natalie, Natalie still identifies itself as Ruby version 3.3. So we'd have to change the tests in `ruby_version_is` to run `ruby -v` and parse that output.

It's probably a whole lot easier to add some upstream specs for `Data#to_h` and get rid of this file.

PR #1775, opened 2024-02-12
Fix calls to _error in Zlib_do_inflate

herwinw:
This is a class method, not an instance method.
The current isuse is visible in the nightly specs of `library/zlib/inflate/set_dictionary_spec.rb`:
```
undefined method `_error' for an instance of Zlib::Inflate
```
With this change, we still raise the wrong exception, but at least the first step is fixed.


herwinw:
Actually, that issue regarding the wrong exception was easy to fix, the old code altered the return value for some reason.

seven1m (approved):
Nice

PR #1779, opened 2024-02-13
Implement $" / $LOADED_FEATURES

herwinw:
This currently is a compilation construct instead of a language construct (comparable to `$!`), which means we're slightly limited compared to MRI.


seven1m:
The gc_lint script seems to be confused. 🤔

herwinw:
That's a weird one.
I've actually been refactoring this change in my head (without the weird `files` argument), I guess that will fix this GC-Lint issue as well.

PR #1781, opened 2024-02-13
Import net-http gem

herwinw:
This is going to be a long project to get some HTTP connectivity working, mostly to add the essential functionality to OpenSSL. I've added a simple script that I want to get working:
```ruby
require 'net/http'

puts Net::HTTP.get(URI('http://natalie-lang.org/'))
puts Net::HTTP.get(URI('https://natalie-lang.org/specs'))
```

I've taken some shortcuts with some required modules. It tried to import the Timeout gem, but this far it only used the exceptions from this code, so I've just limited this file to the exceptions and removed all the other dependencies in timeout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ the respective file, in the same directory with a name like `LICENSE`, or both.
 | `prettyprint.rb` | Yukihiro Matsumoto                | BSD               |
 | `shellwords.rb`  | Akinori MUSHA                     | BSD               |
 | `spec/*`         | Engine Yard, Inc.                 | MIT               |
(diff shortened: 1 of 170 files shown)


herwinw:
Okay, the HTTP part is working (at least good enough for what I'm trying to do), time to move to the OpenSSL part
```
$ bin/natalie -rnet/http -e "puts Net::HTTP.get(URI('http://natalie-lang.org/'))"
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

herwinw:
Let's close this WIP-request for no, I will keep the branch and continue with this some day.

PR #1785, opened 2024-02-16
Use __constant__ macro in zlib to define constants

herwinw:
This results in code that is a bit easier to understand, less repetitive, and has some `#ifdef` checks for free.


seven1m (approved):
🎉

PR #1787, opened 2024-02-17
Implement IO#wait and IO#write_nonblock

herwinw:
And refactor `IO#wait_readable` and `IO#wait_writable` to use `IO#wait` to make the code more generic (which will become useful once we actually start working on the Fiber scheduler for IO). Include some minor refactors to reduce the code duplication when setting IO objects to nonblock.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -113,6 +113,7 @@ public:
     static Value try_convert(Env *, Value);
     Value ungetbyte(Env *, Value);
     Value ungetc(Env *, Value);
+    Value wait(Env *, Args);
     Value wait_readable(Env *, Value = nullptr);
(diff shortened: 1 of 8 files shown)


include/natalie/io_object.hpp
```diff
@@ -95,6 +96,7 @@ class IoObject : public Object {
     void set_fileno(int fileno) { m_fileno = fileno; }
     Value set_lineno(Env *, Value);
     Value set_sync(Env *, Value);
+    void set_nonblock(Env *, bool) const;
```
herwinw, line 99: I had to make this one `const` to make clang-tidy pass, but I can't say I agree with this analysis. Yes, from a compilers point of view we don't mutate the object, but it's still a mutation of something inside the object.

PR #1795, opened 2024-02-18
Implement StringIO#binmode

herwinw:


seven1m (approved):

herwinw:
It looks like I accidentally included these commits in  #1796 as well, which means they are merged via that pull request. I'm closing this one, since merging this won't have any effect.

PR #1808, opened 2024-02-19
Mark $" as read only

herwinw:
And provide a framework to check this.

### DIFF
--- a/src/global_env.cpp
+++ b/src/global_env.cpp
@@ -14,11 +14,8 @@ void GlobalEnv::add_file(Env *env, SymbolObject *name) {
 
     m_files.set(name);
 
-    auto loaded_features = new ArrayObject { m_files.size() };
-    for (auto [file, _] : m_files) {
-        loaded_features->push(file->to_s(env));
-    }
-    global_set(env, "$\""_s, loaded_features, false);
+    auto loaded_features = global_get(env, "$\""_s);
+    loaded_features->as_array()->push(name->to_s(env));
 }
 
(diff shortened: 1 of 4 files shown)


src/global_env.cpp
```diff
@@ -3,6 +3,13 @@
 
 namespace Natalie {
 
+static const auto read_only_globals = []() {
+    TM::Hashmap<TM::String> read_only_globals;
+    read_only_globals.set("$\"");
+    read_only_globals.set("$LOADED_FEATURES");
+    return read_only_globals;
+}();
```
seven1m, line 11: What do you think about putting a `bool` on the [GlobalVariableInfo](https://github.com/natalie-lang/natalie/blob/master/include/natalie/global_variable_info.hpp) class instead of having this additional hashmap? We already have to do the lookup on line 60 to get the `info` object, so we could check the readonly bool there.

herwinw: That does not solve the issue of getting the information there in the first place.
The code currently looks like this:
```c++
auto info = m_global_variables.get(name, env);
if (info) {
    info->set_object(val.object());
} else {
    auto info = new GlobalVariableInfo { val.object() };
    m_global_variables.put(name, info, env);
}
```
I guess you could add a lookup in the else branch of the name for the name being of a read only variable, but you would still need the lookup table to determine this.

herwinw: That makes sense, and it also removes the `$LOADED_FEATURES` alias from the global read only list.

seven1m: I have an idea to try to eliminate the additional hashmap, but I will give that a go in a separate PR...

PR #1815, opened 2024-02-21
Set $" read-only in build_top_env()

seven1m:


src/global_env.cpp
```diff
@@ -44,19 +38,18 @@ Value GlobalEnv::global_get(Env *env, SymbolObject *name) {
...
     if (info) {
-        if (check_readonly && info->is_readonly())
+        if (info->is_readonly())
             env->raise_name_error(name, "{} is a read only variable", name->string());
         info->set_object(val.object());
     } else {
-        auto readonly = readonly_globals.get(name->string()) != nullptr;
         auto info = new GlobalVariableInfo { val.object(), readonly };
         m_global_variables.put(name, info, env);
     }
```
herwinw, line 55: I think we should change line 49 to:
```c++
if (info->is_readonly() && !readonly)
```
Sometimes we have to reassign a global variable internally (e.g. `$?` after a subprocess has been run), that won't be possible with this code.

seven1m: Yeah good call.

PR #1817, opened 2024-02-22
Set $? global variable as read only for users

herwinw:
And fix an inversion in the readonly check.


src/global_env.cpp
```diff
@@ -49,7 +49,7 @@ Value GlobalEnv::global_set(Env *env, SymbolObject *name, Value val, bool readon
         if (!readonly && info->is_readonly())
             env->raise_name_error(name, "{} is a read only variable", name->string());
         if (readonly)
-            assert(!info->is_readonly()); // changing a global to read-only is not anticipated.
+            assert(info->is_readonly()); // changing a global to read-only is not anticipated.
```
seven1m, line 52: Ahhh yep 😅 I knew it was dumb to write code without a corresponding test 🤗

Issue #1551, opened 2023-12-03
Hooked globals


seven1m:
Recording this here so I don't forget. I saw this tweet:

https://x.com/eregontp/status/1731354734118797316?s=46

>TruffleRuby even defines these Regexp-related global variables with Ruby code: https://github.com/oracle/truffleruby/blob/4cd14c48852b7bdb51070b42b50b702dc5a2d7d2/src/main/ruby/truffleruby/core/match_data.rb#L158-L189

This would be a cool idea to steal so we don't keep setting globals on each match.

herwinw:
@seven1m I think this case can be resolved: we have access hooks, so it's just a matter of implementing the missing ones (the NATFIXME blocks in https://github.com/natalie-lang/natalie/blob/master/spec/language/predefined_spec.rb are a good reminder for those).

PR #1818, opened 2024-02-22
Add assignment hooks for global variables

herwinw:
Relates to #1551

This is a quick and dirty PoC to hook a function into the assignment of globals. This PoC only adds a validation, but we can extend this to make it set other globals too, or use the same idea to get the globals.


herwinw:
Deleting this experiment in favour of  #1853.

PR #1819, opened 2024-02-22
Implement throw and catch

herwinw:
For now this is a PoC in Ruby: it does pass the tests, but we probably shouldn't use an Exception object in the throw operations. I had to play around with it a bit at first, since I didn't really know how they were supposed to work.

I've seen this in use in many of the stdgems, so implementing this would simplify the process of importing those from MRI.

### DIFF
--- a/include/natalie/env.hpp
+++ b/include/natalie/env.hpp
@@ -86,6 +86,10 @@ public:
         raise(class_name, message);
     }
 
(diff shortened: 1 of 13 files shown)


herwinw:
And it's been rewritten to C++, the internals are completely abstracted from the user space, and the implementation is much leaner.

PR #1823, opened 2024-02-23
Add example scripts for OpenSSL sockets

herwinw:
This serves mostly as an easy way to test the sockets (related to  #1781, or extracted from that case). I don't want to turn this into a unit test, since it has dependencies on network and external servers. We can add that once we have SSL servers working.

The SSL sockets broke after the nonblock settings in Socket have been added, this change adds a reversal of that setting if the socket is being used in an SSL connection.


seven1m:
Cool!

PR #1825, opened 2024-02-24
Improve String#inspect, implement String#dump

herwinw:
These still have a few encoding issues, but they pass most of the specs.

A lot of specs of  #1781 currently fail due to a missing `String#dump`, so this change should improve that a bit.


herwinw:
Welp, those minor encoding issues break like half of the string specs.

herwinw:
I removed all the encoding changes from this branch to keep that as a future work, the only improvement to `String#inspect` is that it fixes the escapes in `#@` and `#$`. It still adds a very basic version of `String#dump`

PR #1833, opened 2024-02-25
Remove exclusion for spec of SSLSocket#initialize on Darwin

herwinw:
We're updated the Ruby version in the CI.


herwinw:
I'm not sure if this can be removed by now, so if this fails the CI test I'll just remove this branch.

PR #1835, opened 2024-02-25
Certificate validation in OpenSSL

herwinw:
I never knew there were this many hoops to jump through before OpenSSL starts to check your certificates.

Certificate validation has been added to the example script. We should be ready now to start looking at  #1781 again.


seven1m:
This was a monumental effort!

seven1m (approved):

PR #1839, opened 2024-02-26
Possible double free in YAML dump

herwinw:
A bit of background: I was trying to find out why some specs in the nightly runner (e.g. `core/string/dump_spec.rb`) appear as if they are completely skipped (0 passed, 0 failed, 0 errored). It turns out the YAML formatter can create invalid YAML due to a lack of character escaping in `String#inspect`. Even with the improvements of  #1838 it still has a number of issues.

Given that we manually print the YAML output, the easiest thing to do was using the YAML library, but the first implementation resulted in double free errors (at least on my system).

This branch currently is a research branch to dive deeper into that issue.


seven1m:
I was able to reproduce the double free on this branch:

Screenshot 2024-12-30 9 54 38 AM

I admit I haven't looked into how the YAML library works, but it appears we are calling `yaml_emitter_emit()` with the same event object over and over, and the header claims that the event object is deleted.

Screenshot 2024-12-30 9 55 29 AM

So I thought I was onto something, and I merged `master` into this branch, but the problem vanished. I cannot reproduce the double-free now. 😞 

I'm going to close this for now. If it comes up again, maybe we'll remember this info and have something to go on...

PR #1844, opened 2024-02-29
Add PlatformGuard.implementation?

herwinw:
This guards checks for specific Ruby implementations, we probably shouldn't try to be compatible with these corner cases.

Currently, it's only used for specific load paths in MRI.


herwinw:
This converts two errors in the command-line specs into failures.

seven1m (approved):

PR #1851, opened 2024-03-02
Import bubblebabble.c from MRI and run this in Natalie

herwinw:
Current state: it has been converted to use the Natalie API. We should add wrapper functions to emulate the MRI API.


herwinw:
So, this is like 80% of the lines converted and only 20% of the actual work done.

The main issue I'm encountering is the need to pass the `env` argument everywhere. There is no equivalent state object in MRI (mruby does have one, but since the C modules are not compatible that's not a solution for us)
I've started to use some macros instead of functions, but this feels very brittle and I feel like I'm reaching the ceiling for that.

So, the best structure I could come up with is this:
```c++
namespace {
  // Global variables, not thread safe, so every caller has to use the mutext
  Env *env;
  std::recursive_mutex mutex;

  // Translation functions, e.g.
  void rb_raise(VALUE klass, VALUE message) {
    env->raise(klass, message);
  }

  #include "original file from ruby.c"
  rb_define_method(rb_mDigest_Instance, "bubblebabble", rb_digest_instance_bubblebabble, 0); // Example line from the original file, which results in this code:
  auto bubblebabble_lambda = [](Env *env_arg, Value self, Args args, Block *block) -> Value {
    // Only one caller changing *env at a time
    std::lock_guard<std::recursive_mutex> lock(mutex);

    // Set current env
    env = env_arg;

    // Call the original method, which now has an env variable accessible.
    return rb_digest_instance_bubblebabble();
  };
  rb_mDigest->define_method(env, "bubblebabble"_s, bubblebabble_lambda, 0);
}
```

seven1m:
What a neat idea for getting the `env` to be available where we need it! If this works for this library, that could be not-bad.

Another idea I had was to see if we could write a little tool (e.g. `bin/translate_mri_c_to_natalie_c.rb` or some such) that adds the `env` to every call site of a MRI-type function. But that would result in a lot more altered code than your idea, so maybe what you have proposed is better.

herwinw:
I am closing the Pull Request. It served its purpose as a proof of concept, but I'm going to start with a new attempt using a C extension with a few less dependencies (in other words: just write a very basic one from scratch).

I will be developing at the branch https://github.com/herwinw/natalie/tree/run_ruby_c_ext in case anybody want to follow it, but it will be very infrequently.

PR #1852, opened 2024-03-02
Convert inline CPP source to an include statement

herwinw:
The preprocessor does not care about the file extension, and will happily include the C++ file. This preserves the file and line number of the original file, which makes debugging a lot easier.

### DIFF
--- a/include/natalie/file_object.hpp
+++ b/include/natalie/file_object.hpp
@@ -21,6 +21,7 @@ public:
 
     Value initialize(Env *, Args, Block *);
 
+    static Value absolute_path(Env *env, Value path, Value dir = nullptr);
     static Value expand_path(Env *env, Value path, Value root);
     static void unlink(Env *env, Value path);
     static Value unlink(Env *env, Args args);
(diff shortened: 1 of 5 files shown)


herwinw:
I tried this with the following diff:
```diff
diff --git a/lib/tempfile.cpp b/lib/tempfile.cpp
index 18099e1ce..9458cfda2 100644
--- a/lib/tempfile.cpp
+++ b/lib/tempfile.cpp
@@ -45,3 +45,7 @@ Value Tempfile_initialize(Env *env, Value self, Args args, Block *) {
     self->ivar_set(env, "@tmpfile"_s, file);
     return self;
 }
+
+Value Tempfile_crash(Env *env, Value self, Args args, Block *) {
+    NAT_UNREACHABLE();
+}
diff --git a/lib/tempfile.rb b/lib/tempfile.rb
index 7204f0e10..46364e5bb 100644
--- a/lib/tempfile.rb
+++ b/lib/tempfile.rb
@@ -18,6 +18,7 @@ class Tempfile
   end
 
   __bind_method__ :initialize, :Tempfile_initialize
+  __bind_static_method__ :crash, :Tempfile_crash
 
   def close(unlink_now = false)
     @tmpfile.close
```
(Because I always run into unreachable code until I actually need it)

The message on master:
```
$ bin/natalie -rtempfile -e 'Tempfile.crash'
panic: unreachable in /tmp/natalie.cpp20240302-47769-3ymk23#90
```

With this change:
```
$ bin/natalie -rtempfile -e 'Tempfile.crash'
panic: unreachable in lib/tempfile.cpp#50
```

seven1m (dismissed):
Good idea. If we can get that self-hosted test passing, this looks good to me!

herwinw:
My inner monologue during the fix:
Looks like it's using a relative path, let's just use `File.absolute_path` and fix that.
What do you mean method not found?
Well, I could just implement that method to fix it.

herwinw:
This includes some additional code, so I think a new review should be in order.

PR #1853, opened 2024-03-02
Add read hook to global variables

herwinw:
This implements the idea in  #1551 (or part of it, write variables are not yet implemented, and we still have to write it in C++ instead of Ruby as the tweet says).
For now, it is only implemented for `$$`, which adds the behaviour to cache the result. The following code:
```ruby
$stdout.syswrite("before first call")
$stdout.syswrite("#{$$}")
$stdout.syswrite("after first call, before second call")
$stdout.syswrite("#{$$}")
$stdout.syswrite("after second call")
```
Now results in an `strace` log like this for both MRI and Natalie:
```
write(1, "before first call", 17)       = 17
getpid()                                = 60093
write(1, "60093", 5)                    = 5
write(1, "after first call, before second "..., 36) = 36
write(1, "60093", 5)                    = 5
write(1, "after second call", 17)       = 17
```
There is no more need to fill this global at the start, instead we call `getpid` when needed, and save the result.

### DIFF
--- a/include/natalie/global_env.hpp
+++ b/include/natalie/global_env.hpp
@@ -83,6 +83,7 @@ public:
     Value global_set(Env *, SymbolObject *, Value, bool = false);
     Value global_alias(Env *, SymbolObject *, SymbolObject *);
     ArrayObject *global_list(Env *);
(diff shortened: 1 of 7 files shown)


seven1m (approved):
Dang, this is nice!

PR #1855, opened 2024-03-02
Write hooks for global variables

herwinw:
Once again, just a framework and an implementation for a single global variable.

With these two options combined, we can do some cool stuff like storing `$~` (the last regexp match) thread-local, and set read hooks for related variables like `$'` so they get updated automatically.


seven1m (approved):
you're on a roll!

PR #1868, opened 2024-03-04
Implement Module#ruby2_keywords

herwinw:
It's one of those things that I would rather not have, but a surprising number of gems depend on it (I tried to import `delegate` and this was the first error).

On the positive side: it showed two issues with `Hash#dup` that have yakshaved.


seven1m (approved):

herwinw:
Turns out the fixes to `Hash#dup` caused issues in `Hash.[]`. I've moved those to a separate branch, I'll have a look at that later.

PR #1875, opened 2024-03-05
Force initialize calls to return the object itself

herwinw:


seven1m (commented):
I hate that we have to do this symbol lookup and comparison on every method call. I will look to see how MRI does it... 🤔

herwinw:
Only the comparison (which is simply checking to see if two pointers point at the same location). The `static` results in a symbol creation/lookup only the first time this code is ran.

PR #1877, opened 2024-03-06
Import specs for Dir.chroot

herwinw:
Mostly because of the "it works on my pc", but the nightly spec appears to be crashing, so I'm very curious what the regular CI run would do.


herwinw:
Okay, so it breaks on the `as_root` section, which I don't trigger locally. The issue is the call to `Dir['/*']`, this returns the root dir in MRI, but it returns an empty list in Natalie. Changing it to `Dir['/**']` does not change anything. An attempt with `Dir['/home']` appears to get stuck in some infinite loop.

PR #1885, opened 2024-03-08
Add autoload for Set

herwinw:
At least in theory, in practice this means we always load the Set module. It has been written in a way that is fully compatible with an actual dynamic autoload.
The name `prelude.rb` is based on MRI:
```
$ ruby -e 'p Enumerable.instance_method(:to_set)'
#<UnboundMethod: Enumerable#to_set(klass=..., *args, &block) <internal:prelude>:28>
```

This should resolve all the current spec crashes for Enumerable

### DIFF
--- a/lib/set.rb
+++ b/lib/set.rb
@@ -390,13 +390,3 @@ class Set
   end
   alias >= superset?
 end
-
-module Enumerable
-  def to_set(&block)
-    if block
-      Set.new(map(&block))
-    else
-      Set.new(self)
-    end
-  end
-end
--- a/spec/core/enumerable/fixtures/classes.rb
+++ b/spec/core/enumerable/fixtures/classes.rb
@@ -342,4 +342,10 @@ module EnumerableSpecs
       @block.call(*args)
     end
   end
+
+  # Set is a core class since Ruby 3.2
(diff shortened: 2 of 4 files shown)


seven1m:
I need to think about this. I chose not to do this sort of thing for the `PP` module (`Kernel#pp` autoloads `PP` in MRI) because it will bloat every binary.

There are some things we just cannot support, being AOT compiled. This might be one of them.

I wonder if we can apply a patch to the specs... 🤔

herwinw:
Though experiment: can we delay the autoload until we know we need it? For example, this code:
```ruby
autoload :Set, 'set'
puts "Hello, world"
```
does not use a `Set`, so we could just skip the autoload.

So basically we'd be setting a hook to the code generator that if the constant `Set` would be seen, we'd be including the file mentioned in the autoload.

Of course this doesn't cover every scenario in the dynamic nature of Ruby, code like
```ruby
Object.const_get(:Set)
```
would not be recognized (which can take a dynamic value as well)

herwinw:
Thought experiment failed, there's code like this in the code lib:
```ruby
def Enumerable#to_set
  Set.new(self)
end
```
So this would cause the autoload to always load the set lib.

PR #1886, opened 2024-03-08
Explicitly require set in Enumerable specs

herwinw:
The alternative to #1885


herwinw:
Starting the nightly specs locally does not change the file `spec/core/enumerable/fixtures/classes.rb` (in other words: it adds the exact same require header as our local version does)

seven1m (approved):
This works for me

PR #1890, opened 2024-03-15
Fixes to Kernel#define_singleton_method

herwinw:
This resolves the crash in the nightly spec runner


herwinw:
I have to take a new look at this, since it results in seemingly unrelated crashes with asan enabled.

Issue #1893, opened 2024-03-15
Failures in asan test


herwinw:
Extracted from  #1891, since it does happen in the master branch as well.

Running `test/natalie/argument_test.rb` (and possibly other tests as well) with an asan build currently fails at this part of the code:
```c++
Method *Object::find_method(Env *env, SymbolObject *method_name, MethodVisibility visibility_at_least, Value sent_from) const {
    ModuleObject *klass = singleton_class();
    if (!klass)
        klass = m_klass;
    assert(klass);
```

This error originates in the implementation of `Tempfile`:
```ruby
  (File.public_instance_methods - public_instance_methods).each do |method|
    define_method(method) do |*args, **kwargs, &block|
      @tmpfile.public_send(method, *args, **kwargs, &block)
    end
  end
```

And is easy to reconstruct:
```sh
$ bin/natalie -rtempfile -e ''
natalie20240315-252571-nodkaz: src/object.cpp:953: Natalie::Method* Natalie::Object::find_method(Natalie::Env*, Natalie::SymbolObject*, Natalie::MethodVisibility, Natalie::Value) const: Assertion `klass' failed.
```

Rewriting this piece of code to implement `method_missing` instead does resolve this specific issue, but then the next call to `each` breaks:
```ruby
def reset
  @stubs.values.each do |stubs|
    stub = stubs.first
    stub.subject.singleton_class.remove_method(stub.message)
  end
```
This can be rewritten to `@stubs.each_value`, but then something else breaks, so this is obviously no solution.

I've managed to get the following output it gdb:
```
(gdb) p (Natalie::Object)(*this)
$10 = {<Natalie::Cell> = {_vptr.Cell = 0x5310003b7720, m_marked = 224}, m_klass = 0x0,
  m_type = Natalie::ObjectType::Nil, m_singleton_class = 0x0, m_owner = 0x0, m_flags = 0, m_ivars = 0x0}
(gdb) p (Natalie::Object)(*NilObject::the())
$11 = {<Natalie::Cell> = {_vptr.Cell = 0x555556352ae8 <vtable for Natalie::NilObject+16>, m_marked = false},
  m_klass = 0x531000042a20, m_type = Natalie::ObjectType::Nil, m_singleton_class = 0x531000042a20, m_owner = 0x0,
  m_flags = 1, m_ivars = 0x0}
```
This looks like there is a second instance of an object with `m_type == ObjectType::Nil`, which is not supposed to happen.

[...]

PR #1894, opened 2024-03-15
Run asan tests on Debian Bullseye

herwinw:
This is a completely blind attempt at a mitigation for  #1893.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -287,6 +287,16 @@ task :docker_build_clang do
      '.'
 end
 
+task :docker_build_clang_bullseye do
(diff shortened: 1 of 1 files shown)


seven1m:
Fantastic. That was a good hunch! 👍

herwinw:
@seven1m I pushed a new version with an extended commit message (the code itself is unchanged). I think the best action for now would be to merge this change so the CI stops with these false negatives, we can keep the related issue to actually fix the underlying issue (or just wait a few days until Debian pushes some updates)

seven1m (approved):

PR #1895, opened 2024-03-15
Run asan tests on Debian Bookworm

herwinw:
This is the development case related to  #1893


herwinw:
I tried this randomly again, and it appears to be working. I'm just going to assume this was an issue in the Debian based image, and not look further into it.

seven1m:
Weird!

PR #1902, opened 2024-03-16
Bytecode VM: Pack 4 boolean flags for send in a single byte

herwinw:
Ruby has a method `Integer#[]` to access the single bytes of an integer, which means we can reuse the code to split the flags back into separate values.


seven1m (approved):
Noice!

PR #1907, opened 2024-03-17
Bytecode VM: Support push_int

herwinw:
Well, this one took forever, why are integers only trivial on the surface level?

I didn't think BER encoding to be a suitable format here: it only works for positive values, so that would add a sign byte, which results in different representations of `0` and `-0`, even though it's actually the same value.

Instead, it uses Marshal encoding with a twist. The options for the first byte (interpreted as signed) of the encoded value are:
* `0`, if we encode a `0`
* `6..127`, if we encode a value in the range `1..122`, we can decode that one by subtracting `5` from the encoded value
* `1..4`, if we encode a value in the range `123..((1 << 30) - 1)`. This means we have to read this amount of extra bytes and use some shifts/ors on the binary value to retrieve the original numer

Negative values use in a similar way, but using negative values. This means the encoded value will never start with `5` or `-5`. We use those values as markers for positive or negative values out of the regular range. The number will be encoded as BER encoding of the absolute value.

### DIFF
--- a/lib/natalie/compiler/instructions/push_int_instruction.rb
+++ b/lib/natalie/compiler/instructions/push_int_instruction.rb
@@ -24,6 +24,76 @@ module Natalie
       def execute(vm)
         vm.push(@int)
       end
+
+      def serialize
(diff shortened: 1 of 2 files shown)


seven1m:
Cool!

PR #1910, opened 2024-03-17
Bytecode VM: Fix VariableGetInstruction deserialize

herwinw:
The old code resulted in the following disassembled instruction:
```
variable_get Natalie::Compiler::VariableGetInstruction
```

This fixes it to:
```
variable_get x
```

We can now run the following piece of code
```ruby
puts ['a', '0'].map { |x| x.ord }
```
This one is also included as a unit test.


herwinw:
And implement `variable_declare` as well. This can now run:
```ruby
a = 1
puts a
```
Which is included as a unit test as well.

PR #1922, opened 2024-03-21
Include a header for Bytecode files

herwinw:
This header consists of a magic 32 bit value (representing `NatX`), and two bytes for the version. The current version has been set to 0.0, meaning there is zero backwards compatibility guarantee and every git pull can cause your bytecode files to become invalid.

### DIFF
--- a/lib/natalie/compiler.rb
+++ b/lib/natalie/compiler.rb
@@ -52,6 +52,8 @@ module Natalie
     end
 
     def compile_to_bytecode(io)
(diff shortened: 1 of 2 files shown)


herwinw:
This is the first step of adding sections into the bytecode format (as discussed on Discord, https://discord.com/channels/868454594085281793/868454594085281796/1219370877981032498)

seven1m (approved):
Fancy!

PR #1929, opened 2024-03-24
Implement UDPSocket#recvfrom_nonblock

herwinw:


herwinw:
I'm skipping the buffer argument for now (in #1927 too). There is currently some discussion about the expected encoding of the result on the ruby-spec repo (do you use the encoding of the buffer argument, or the default binary encoding, both are kind of logical, depending on your use case).

PR #1934, opened 2024-03-26
Don't implement BasicSocket#getpeereid

herwinw:


herwinw:
```ruby
__ld_flags__ '-lbsd'
```
Very curious if this is going to work on OSX

herwinw:
```
/Users/runner/work/natalie/natalie/lib/socket.cpp:2:10: fatal error: 'bsd/unistd.h' file not found
#include <bsd/unistd.h>
         ^~~~~~~~~~~~~~
```
Turns out it doesn't even get to that point :D

herwinw:
So, the implementation that works on my machine breaks all the CI runs, the implementation that fixes these CI runs breaks compilation on my machine.

I replaced the implementation with a NATFIXME in the specs and a comment on why we should skip this one.

seven1m (approved):

PR #1940, opened 2024-03-29
Implement IntegerObject::convert_to_native_type

herwinw:
Add this to IoObject::read to get a user of the new code. We should probably convert most callers of IntegerObject::convert_to_nat_int_t to use this new code.

### DIFF
--- a/include/natalie/integer_object.hpp
+++ b/include/natalie/integer_object.hpp
@@ -13,6 +13,7 @@
 #include "natalie/object.hpp"
 #include "natalie/rounding_mode.hpp"
 #include "natalie/types.hpp"
+#include "nathelpers/typeinfo.hpp"
 
 namespace Natalie {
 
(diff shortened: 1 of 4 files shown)


include/natalie/integer_object.hpp
```diff
@@ -13,6 +13,7 @@
 #include "natalie/object.hpp"
 #include "natalie/rounding_mode.hpp"
 #include "natalie/types.hpp"
+#include "nathelpers/typeinfo.hpp"
```
herwinw, line 16: I could not get the gc_lint file to skip the structs in this file in a non-hacky way, so I resorted to moving this file into a different directory.

include/nathelpers/typeinfo.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <cstdint>
+
+namespace Natalie {
+
+template <class T>
+struct typeinfo {
+    constexpr const char *name() = delete;
```
herwinw, line 9: These are all done at compile time. When asked for a value that is not yet defined, it results in a compilation error that's easy to solve by adding a more specific `typeinfo` struct down below.

PR #1944, opened 2024-03-30
Refresh specs of io-wait after upstream changes

herwinw:


spec/library/io-wait/wait_spec.rb
```diff
@@ -111,7 +122,7 @@
     end
 
     it "accepts a list of modes" do
-      @io.wait(0, :r, :w, :rw).should == @io
+       @io.wait(0, :r, :w, :rw).should == @io
```
seven1m, line 125: Someone is messing up indentation upstream 😬

PR #1951, opened 2024-04-01
Make Object::is_a faster

### LINKED ISSUE #1874: Object#is_a? is not very efficient
The current code 
```c++
bool Object::is_a(Env *env, Value val) const {
   if (/* direct match (instance_of?), not relevant here */) {
    } else {
        ClassObject *klass = singleton_class();
        if (!klass)
            klass = m_klass;
        ArrayObject *ancestors = klass->ancestors(env);
        for (Value m : *ancestors) {
            if (module == m->as_module()) {
                return true;
            }
        }
        return false;
    }
}
```
In other words: unless we have a direct match, we first build a list of all the ancestors, then loop through the results to see if the wanted value is in it.

This could be done more efficient by removing the temporary array and directly looping with a method. The current code is pretty much like `enumerable.to_a.include?(val)`, which could be written as `enumerable.include?(val)`.

tekknolagi:
Don't allocate an intermediate ArrayObject.

Fix #1874

### DIFF
--- a/src/module_object.cpp
+++ b/src/module_object.cpp
@@ -663,8 +663,6 @@ ArrayObject *ModuleObject::ancestors(Env *env) {
 }
 
 bool ModuleObject::ancestors_includes(Env *env, ModuleObject *module) {
-    std::lock_guard<std::recursive_mutex> lock(g_gc_recursive_mutex);
-
     ModuleObject *klass = this;
     do {
         if (klass->included_modules().is_empty()) {
--- a/src/module_object.cpp
+++ b/src/module_object.cpp
@@ -662,6 +662,27 @@ ArrayObject *ModuleObject::ancestors(Env *env) {
     return ancestors;
 }
 
+bool ModuleObject::ancestors_includes(Env *env, ModuleObject *module) {
+    std::lock_guard<std::recursive_mutex> lock(g_gc_recursive_mutex);
+
+    ModuleObject *klass = this;
+    do {
+        if (klass->included_modules().is_empty()) {
+            // note: if there are included modules, then they will include this klass
+            if (klass == module) {
+                return true;
+            }
+        }
+        for (ModuleObject *m : klass->included_modules()) {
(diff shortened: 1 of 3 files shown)


tekknolagi:
I'm not sure if I need the lock anymore since I am not allocating.

herwinw:
> I'm not sure if I need the lock anymore since I am not allocating.

That one should no longer be needed. The code is very similar to `bool ModuleObject::is_subclass_of(ModuleObject *other)` just below your new code, and that one doesn't require it either.

And to be forward about this: my (undocumented) end goal here was to create an iterable interface for the ancestors. We now have `ModuleObject::ancestors` and `ModuleObject::ancestors_include`, so I kind of want to use things like `std::any_of` for that second one.
This does not mean that I don't like your change. Quite the opposite: if I were to implement the iterator interface from the current master, I would probably start with a commit like this one just to illustrate the similarities between these two methods, to make it easier to extract the common code.

src/module_object.cpp
```diff
@@ -662,6 +662,27 @@ ArrayObject *ModuleObject::ancestors(Env *env) {
     return ancestors;
 }
 
+bool ModuleObject::ancestors_includes(Env *env, ModuleObject *module) {
+    std::lock_guard<std::recursive_mutex> lock(g_gc_recursive_mutex);
```
herwinw, line 666: As described in a bit more detail in the comments: this line can be removed. Once it is gone, I would say it's good to merge.

herwinw (approved):

src/module_object.cpp
seven1m: I'm going to investigate this more in #1956. My gut tells me we *do* want the lock here (and in the method below). But I won't hold up this work!

seven1m (approved):
Thanks!

src/module_object.cpp
tekknolagi: Oh, why's that?

PR #1957, opened 2024-04-02
Lock methods that iterate over the m_included_modules vector

### LINKED ISSUE #1956: We may need a lock in ModuleObject#ancestors_includes and ModuleObject#is_subclass_of
Just a note for myself to investigate this more...

```ruby
module Bar; end
class Foo; include Bar; end

Thread.new do
  10000.times do
    mod = Module.new
    Foo.include(mod)
  end
end

1000.times do
  Foo.new.is_a?(Bar)
end

```

seven1m:
Fixes #1956

### DIFF
--- a/include/natalie/module_object.hpp
+++ b/include/natalie/module_object.hpp
@@ -85,7 +85,7 @@ public:
 
     void included_modules(Env *, ArrayObject *);
     Value included_modules(Env *);
-    Vector<ModuleObject *> included_modules() { return m_included_modules; }
+    Vector<ModuleObject *> &included_modules() { return m_included_modules; }
     bool does_include_module(Env *, Value);
 
     virtual Value cvar_get_or_null(Env *, SymbolObject *) override;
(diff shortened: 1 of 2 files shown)


seven1m:
Just for clarity, I actually got a _different_ error before adding cc34116 because we were returning a copy of the vector. But since copying a vector would be similarly racy, I just lumped these two bug fixes together.

include/natalie/module_object.hpp
```diff
@@ -85,7 +85,7 @@ class ModuleObject : public Object {
 
     void included_modules(Env *, ArrayObject *);
     Value included_modules(Env *);
-    Vector<ModuleObject *> included_modules() { return m_included_modules; }
+    Vector<ModuleObject *> &included_modules() { return m_included_modules; }
```
herwinw, line 88: This now returns a mutual reference, so this code:
```c++
module->included_modules().clear()
```
now resets the included modules, where before it operated on it's own copy.

I think this one should be `const` instead.

PR #1973, opened 2024-04-07
Rename PushComplexInstruction to CreateComplexInstruction

herwinw:
This instruction does not push anything on the stack, it converts the last two elements on the stack in a new element. This is similar to instructions as CreateArrayInstruction.


seven1m (approved):
🤷‍♂️ I struggle to know which ones to call what. They almost all push _something_.

herwinw:
> 🤷‍♂️ I struggle to know which ones to call what. They almost all push something.

My interpretation is that the push operations push something completely new on top of the existing stack. As soon as something is popped in the operation, it is some kind of conversion. That the result of the conversion is pushed back onto the stack is a side effect of a stack architecture.

PR #1988, opened 2024-04-19
Bytecode VM: Support PushLastMatchInstruction

herwinw:
There are some issues with using references in the interpreter mode (and thus in running the Bytecode VM). Compare:
```
bin/natalie -e '/(?<foo>bar)/ =~ "foobar"; p $~'
#<MatchData "bar" foo:"bar">

bin/natalie -i -e '/(?<foo>bar)/ =~ "foobar"; p $~'
nil
```

I expect the interpreter mode uses some regexp internally and overwrites the global


herwinw:
It needed a small update to accompany the changes in #1990.

PR #2006, opened 2024-04-26
Streamline the verbose warn idiom

### LINKED ISSUE #2001: Streamline the verbose warn idiom
In #2000 I needed to print a warning in case the `$VERBOSE` variable is set. This resulted in code like this:
```ruby
if (env->global_get("$VERBOSE"_s)->is_truthy())
    env->warn("Blahblahblah");
```
This code is quiet verbose (badum-tss), and I had to copy it from `module_object.cpp`. This can be improved:

First: we could combine these statements into a `env->verbose_warn(String)` to remove this duplication. This would make it a lot easier to use.

Second: we already got read/write hooks in use for this variable (because `$VERBOSE = :any_truthy_value; $VERBOSE == true`). We could use these hooks to set an instance variable on `GlobalEnv`, so we can just check a boolean and improve the speed of this code.

herwinw:
This closes #2001.
I'm not completely sure about the `verbose_warn` function name, it might be interpreted as "make your warning extra verbose", I'm open for suggestions on a better name.

### DIFF
--- a/include/natalie/env.hpp
+++ b/include/natalie/env.hpp
@@ -99,7 +99,7 @@ public:
 
     template <typename... Args>
     void verbose_warn(const char *format, Args... args) {
-        if (GlobalEnv::the()->is_verbose(this))
+        if (GlobalEnv::the()->is_verbose())
             warn(String::format(format, args...));
     }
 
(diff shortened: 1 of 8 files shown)


seven1m (approved):
Looks great

PR #2007, opened 2024-04-26
Improve complain checker in verbose mode

### LINKED ISSUE #2004: Improve verbose complain spec checker
Warning: don't start with this issue before #2001 is resolved, otherwise you'd end up with a lot of duplicated code

There are a number of specs that try to check if the code complains in verbose mode. An example from `array/new_spec.rb`:
```ruby
-> {
  -> { Array.new { raise } }.should_not raise_error
}.should complain(/warning: given block not used/, verbose: true)
```
This complain expectation works like this;
* Store the old value of `$VERBOSE` in a variable
* Set `$VERBOSE` to true
* Run the code
* Validate the warning with the given regexp
* Restore the value of `$VERBOSE`

This is only half of the test, since `verbose: true` also implies we shouldn't be seeing this message if we're in regular mode. And indeed, the given code does warn in Natalie in regular mode.
```
$ ruby -e 'Array.new { raise }'
$ ruby -e '$VERBOSE=1; Array.new { raise }'
-e:1: warning: given block not used
$ bin/natalie -e 'Array.new { raise }'
-e:1: warning: given block not used
```

If the verbose argument is set, we probably need to run the block twice, once in verbose mode which should print the warning, and once in regular mode which should not print the warning.

herwinw:
This fixes #2004

### DIFF
--- a/test/support/spec.rb
+++ b/test/support/spec.rb
@@ -903,6 +903,11 @@ class ComplainExpectation
     else
       puts "Expected a Regexp to complain but got #{@message.inspect}"
     end
+
+    # If the verbose argument is given, the code should *not* complain with verbose disabled
+    if @verbose
+      ComplainExpectation.new(@message, verbose: false).inverted_match(subject)
+    end
   end
 
   def inverted_match(subject)


herwinw:
This is just the first step: add the extra check. A number of checks will fail (i.e. the `Array.new` spec mentioned in the other case). The `Regexp.new` check passes.

This should be put on hold until #2006 is merged.

herwinw:
```ruby
  it "prints the message when VERBOSE is false" do
    -> { Warning.warn("foo") }.should complain("foo")
  end

  it "prints the message when VERBOSE is nil" do
    -> { Warning.warn("foo") }.should complain("foo", verbose: nil)
  end

  it "prints the message when VERBOSE is true" do
    -> { Warning.warn("foo") }.should complain("foo", verbose: true)
  end
```
So I guess this is not how the specs are supposed to be interpreted.

New plan: review the failed specs, manually check every offence to see the behaviour in MRI, update our behaviour if required, throw away this change in the specs.

herwinw:
I'm closing this one, since this is simply not how the specs are supposed to work.

PR #2017, opened 2024-05-01
Bytecode VM: Add with_singleton instruction

herwinw:
So, I had no idea what this instruction did, so I had to look up the usage of it in `pass1.rb`, then convert that method name back to the Prism class, and look in the Prism documentation to find an example :D


seven1m (approved):
For sure, it's confusing. I'm open to a better name!

PR #2018, opened 2024-05-10
Update Prism to v0.28.0 and remove our patches

seven1m:
It's nice we no longer need our patch hack. 🎉


include/natalie/integer_object.hpp
```diff
@@ -77,14 +76,14 @@ class IntegerObject : public Object {
     static T convert_to_native_type(Env *env, Value arg) {
         auto integer = arg->to_int(env);
         if (integer->is_bignum())
-            env->raise("RangeError", "bignum too big to convert into `{}'", typeinfo<T>().name());
+            env->raise("RangeError", "bignum too big to convert into `{}'", typeid(T).name());
```
herwinw, line 79: I have looked into `typeid` when I wrote this code, and this is not a suitable alternative. Instead of the name, it returns a one character name that may or may not differ between compilers/architectures.

For example: `int32_t` returns "i", `uint32_t` returns "j", `int8_t` return "a", `uint8_t` return "h".
I don't think I would understand what an error message "to big to convert into `j'" would mean.

include/natalie/integer_object.hpp
```diff
@@ -78,7 +78,7 @@ class IntegerObject : public Object {
         if (integer->is_bignum())
             env->raise("RangeError", "bignum too big to convert into `{}'", typeid(T).name());
         const auto result = integer->to_nat_int_t();
-        if (!std::numeric_limits<T>::is_signed && result < 0)
+        if (std::numeric_limits<T>::is_signed && result < 0)
```
herwinw, line 81: This change is incorrect. It might be a bit confusing, since we check for the type being signed and reverse that result (there is no `is_unsigned` property in `std::numeric_limits`).

Easiest way to test this:
```sh
ruby -e 'p File.open("/dev/zero").read(-1)'
bin/natalie -e 'p File.open("/dev/zero").read(-1)'
```
Both these lines (with Natalie master) throw `-e:1:in `read': negative length -1 given (ArgumentError)`

With the above fix, Natalie throws `-e:1:in `read': integer -1 too small to convert to `size_t' (RangeError)`, which is still correct, but not the expected error.

lib/ffi.cpp
```diff
@@ -220,34 +233,95 @@ Value FFI_Pointer_read_string(Env *env, Value self, Args args, Block *) {
...
+        } else if (sym == "short"_s || sym == "ushort"_s || sym == "int16"_s || sym == "uint16"_s) {
+            size = 2;
+        } else if (sym == "int32"_s || sym == "uint32"_s || sym == "float"_s) {
+            size = 4;
+        } else if (sym == "int64"_s || sym == "uint64"_s || sym == "long_long"_s || sym == "ulong_long"_s || sym == "double"_s) {
+            size = 8;
+        } else if (sym == "int"_s) {
+            size = sizeof(signed int);
+        } else if (sym == "uint"_s) {
+            size = sizeof(unsigned int);
```
herwinw, line 277: I'm not completely sure, but I would think the standards define that the signed and unsigned pairs have the same size, so these could be combined:
```c++
} else if (sym == "int"_s || sym == "uint"_s) {
    size = sizeof(int);
```

lib/ffi.cpp
```diff
@@ -220,34 +233,95 @@ Value FFI_Pointer_read_string(Env *env, Value self, Args args, Block *) {
...
+            env->raise("TypeError", "unknown size argument for FFI#initialize: {}", args.at(0)->inspect_str(env));
+        }
+        break;
+    }
+    case Object::Type::Integer: {
+        size = IntegerObject::convert_to_native_type<size_t>(env, args.at(0));
+        break;
+    }
+    default:
+        env->raise("TypeError", "no implicit conversion of {} into Integer", args.at(0)->klass()->class_name());
```
herwinw, line 294: These two options can be defined: `IntegerObject::convert_to_native_type` throws this error in case the input is not an Integer, so that would remove some duplication.

lib/ffi.cpp
```diff
@@ -220,34 +233,95 @@ Value FFI_Pointer_read_string(Env *env, Value self, Args args, Block *) {
...
+    }
+
+    size_t count = 1;
+    if (args.size() > 1) {
+        auto signed_count = args.at(1)->as_integer_or_raise(env)->to_nat_int_t();
+        if (signed_count < 0)
+            env->raise("ArgumentError", "negative allocation count");
+        count = static_cast<size_t>(signed_count);
```
herwinw, line 302: This adds the extra validation and bignum issue that I tried to prevent with `IntegerObject::convert_to_native_type`. I assume this is done because this has a different error message than the default "negative length (-xyz) given". I think the correct fix for that would be to add an optional template argument to `IntegerObject::convert_to_native_type`

include/natalie/integer_object.hpp
seven1m: Thanks. This PR is draft because I am still fiddling with it.

lib/ffi.cpp
seven1m: I don't mind the duplication in this case.

PR #2025, opened 2024-05-16
Add accessor for options to SSLContext

herwinw:


include/nathelpers/typeinfo.hpp
```diff
@@ -19,4 +19,11 @@ struct typeinfo<std::size_t> {
     constexpr const char *name() { return "size_t"; }
 };
 
+#ifdef __APPLE__
```
seven1m, line 22: Why only on Apple?

seven1m (approved):

include/nathelpers/typeinfo.hpp
herwinw: Every Linux system I've tried (my own system and the CI) uses the same underlying type for `std::size_t` and `uint64_t`, which causes compilation errors
```
nclude/nathelpers/typeinfo.hpp:24:8: error: redefinition of ‘struct Natalie::typeinfo<long unsigned int>’
   24 | struct typeinfo<uint64_t> {
      |        ^~~~~~~~~~~~~~~~~~
include/nathelpers/typeinfo.hpp:19:8: note: previous definition of ‘struct Natalie::typeinfo<long unsigned int>’
   19 | struct typeinfo<std::size_t> {
      |        ^~~~~~~~~~~~~~~~~~~~~
```
This is not the case on Apple (or at least: in the Apple CI we're using, see https://github.com/natalie-lang/natalie/actions/runs/9108336727/job/25038892473).

I tried to use `std::enable_if` to get this to work compile conditionally, which would be a bit more reliable, but I didn't manage to get that to work (but I'm always confused by the meta compilation things of C++, so that might as well just be me).

seven1m: Ahh ok, makes sense!

PR #2028, opened 2024-05-18
Add compatibility with Ruby 3.4.1

herwinw:
I'm hesitant whether we should merge this, it might get in the way of some new features later on.


seven1m (dismissed):

seven1m:
I'll defer to you as our compatibility expert 😂

herwinw:
I'm going to put this on hold for now. Most issues are related to frozen string literals, I'm well on my way of fixing those.

herwinw:
This is getting along nicely. A few remarks:
* Docker is running on 3.4.0-rc1, since the stable images are not released yet (they're always a few days late due to Christmas), once they're released we can simply drop the last commit
* Macos is not yet running, there is no ruby3.4 in brew yet
* The new formatting of `Hash#inspect` (extra space on both sides of `=>`) is now different from the output of `ENV.inspect`, so this required a new function instead of delegating it to the Hash object. I made an upstream report, since this looks weird and inconsistent (https://bugs.ruby-lang.org/issues/20982).
* The statement `defined?({1=>@baz})` where the ivar does not yet exist now differs from MRI, the Natalie tests have an ugly workaround to be able to run this test with MRI 3.3, MRI 3.4 and Natalie.

herwinw:
It's passing all tests now, so it's finally ready to merge.

.github/workflows/test.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           PKG_CONFIG_PATH: /usr/local/opt/openssl@3/lib/pkgconfig:/opt/homebrew/opt/openssl@3/lib/pkgconfig:/usr/local/opt/libffi/lib/pkgconfig:/opt/homebrew/opt/libffi/lib/pkgconfig
           NAT_BUILD_MODE: release
         run: |
-          export PATH="/opt/homebrew/opt/ruby@3.3/bin:$PATH"
```
seven1m, line 86: Hmmm, for a short period yesterday, the `/opt/homebrew/opt/ruby` path didn't exist. But I guess they fixed it. 🤷‍♂️

herwinw: The same path (without the explicit version) is used in an earlier debug step to print the ruby version, so this makes it more consistent as well.

test/natalie/enumerator_test.rb
```diff
@@ -184,8 +184,16 @@ def to_ary
...
+      ruby_version_is ''...'3.4' do
+        it 'raises if size is bignum and n is bignum' do
+          -> { (1..bignum_value).step.last(bignum_value) }.should raise_error(RangeError, "bignum too big to convert into 'long'")
+        end
+      end
+
+      ruby_version_is '3.4' do
+        it 'raises if size is bignum and n is bignum' do
+          -> { (1..bignum_value).step.last(bignum_value) }.should raise_error(RangeError, "bignum too big to convert into 'long'")
+        end
```
seven1m, line 196: There appears to be no difference between these two blocks -- the two ruby versions behave the same?

seven1m (approved):
That was a huge effort! Well done.

So. Many. Backticks. 😆

test/natalie/enumerator_test.rb
herwinw: No, this one resulted in an error when using MRI 3.3 as a host version. Just one more backtick

PR #2036, opened 2024-05-22
Support Prism::NoKeywordsParameterNode

herwinw:
This supports method definitions like this:
```ruby
def foo(*args, **nil) = args
```
Without the `**nil`, the keywords would sneakily end up in the `args` array.

Ruby uses a separate error message for this, this change uses the generic error message for extra keywords:
```
$ ruby -e 'def foo(*a, **nil) = a; foo(1, a: 1)
> '
-e:1:in `<main>': no keywords accepted (ArgumentError)

def foo(*a, **nil) = a; foo(1, a: 1)
                            ^^^^^^^
$ bin/natalie -e 'def foo(*a, **nil) = a; foo(1, a: 1)'
        1: from -e:1:in `<main>'
-e:1:in `foo': unknown keyword: :a (ArgumentError)
```

This resolves the compile error in https://github.com/ruby/spec/blob/master/core/method/parameters_spec.rb (which causes a bunch of extra errors in the specs, since we have not implemented that method)

### DIFF
--- a/lib/natalie/compiler/args.rb
+++ b/lib/natalie/compiler/args.rb
@@ -59,6 +59,8 @@ module Natalie
           transform_required_keyword_arg(arg)
         when ::Prism::OptionalKeywordParameterNode
           transform_optional_keyword_arg(arg)
+        when ::Prism::NoKeywordsParameterNode
+          transform_no_keyword_arg(arg)
         when ::Prism::ImplicitRestNode
           clean_up_keyword_args
           transform_implicit_rest_arg(arg)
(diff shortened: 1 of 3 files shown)


herwinw:
Note to self:
```ruby
def foo(*a, b: 1, **nil) = a
p foo(1)
```
This code now works in Natalie, but fails to compile in MRI (`syntax error, unexpected `nil', expecting ')'`). This might be a bug in Prism, but I'm going to wait until the update of  #2018 before investigating it deeper.

seven1m (approved):

PR #2037, opened 2024-05-23
Fixes to handling CallOrWriteNode

herwinw:
```ruby
a.foo ||= b
```
This should evaluate `a` only once.

This is just the beginning, there are a lot of calls similar to this. https://kddnewton.com/2023/12/06/advent-of-prism-part-6 and https://kddnewton.com/2023/12/16/advent-of-prism-part-16 describe most (or maybe all) of them.


seven1m (approved):
Yay!

PR #2043, opened 2024-05-24
Support ConstantPathOperatorWriteNode

herwinw:
This supports code like this:
```ruby
Foo::Bar += 1
```

TODO:
* [x] Rebase once the upgrade to Prism 0.29 is merged due to conflicting `operator`/`binary_operator` naming (#2042)
* [x] Save the temporary result of the path, to prevent duplicate calls (comparable to #2037). The included spec has a test for this behaviour.

This resolves he compile issues with https://github.com/ruby/spec/blob/master/language/assignments_spec.rb

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -906,6 +906,31 @@ module Natalie
         ]
       end
 
+      def transform_constant_path_operator_write_node(node, used:)
+        name, _is_private, prep_instruction = constant_name(node.target)
+        # FIXME: is_private shouldn't be ignored I think
+        instructions = [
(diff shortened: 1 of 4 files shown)


herwinw:
The code now kind of behaves like this:
```ruby
Foo::Bar::Baz += 1 # source

tmp = Foo::Bar
value = tmp::Baz = tmp::Baz + 1
return tmp::Baz if used
```
I tried to convert that last line to `return value if used`, but I could not find a way to get that to work with the stack, so we would either need something like a `swap2` instruction or a register based VM. I guess the extra const lookup is good enough.

seven1m (approved):

PR #2046, opened 2024-05-25
Import missing language specs with at least 1 passing spec

herwinw:


seven1m (approved):

seven1m:
We can remove the `numbers_spec.rb` from the self-hosted tests for now.

herwinw:
For now, I just disabled that single failing test. It contains a NATFIXME marker with more information, so it's easier to find than when we would just remove that whole spec suite.

PR #2047, opened 2024-05-27
Refactor handling of ConstantOrWriteNode

herwinw:
The old code worked like this:
```ruby
if !defined?(CONST) || !CONST
  CONST = value
else
  CONST
end
```
With a bit of DeMorgan and flipping the bodies, we can rewrite this as:
```ruby
if defined?(CONST) && CONST
  CONST
else
  CONST = value
end
```
This removes two operations from the resulting code.


seven1m (approved):
🎉

PR #2049, opened 2024-05-28
Faster bigint

### LINKED ISSUE #2040: Bignums are SLOW in the self-hosted compiler
```sh
time bin/natalie -e "p 2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
bin/natalie -e   0.65s user 0.16s system 99% cpu 0.809 total

time bin/nat -e "p 2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
bin/nat -e   26.19s user 4.23s system 99% cpu 30.426 total
```

I added a printf to the `BigInt(const TM::String &, const base = 10)` constructor, and it:

- is called once with `bin/natalie`
- is called 5,068,273 times with `bin/nat` 😱 (5,030,321 times after #2041)

I'll keep digging on this as I have time...

seven1m:
This library uses 32-bit int "words" instead of a string.

This should fix #2040.


seven1m:
>My gut feeling says that saving the absolute value of the number and adding a boolean signed field to the struct would be easier. But that's really a gut feeling, I haven't though about this low level stuff in ages.

That's how this library works: it has a magnitude and a sign flag. It's beautiful.

But we still have to mimic the quirkiness of two's complement representation when doing a bit shift and certain bitwise operations, as I already did a little bit. Right shift is broken for negative numbers because it's more correct than CRuby is, so now I get to make it have the same bugginess as CRuby. Yay.

PR #2050, opened 2024-05-28
Use Symbol as internal data type in ConstSetInstruction

herwinw:
This removes the differences in internal formatting with ConstFindInstruction, and can now use a shared symbol in the generated C++ code.


seven1m (approved):

herwinw:
I ran the whole test suite with a `raise unless symbol.is_a?(Symbol)` in the relevant code to check if there were more offenders. The only other issue that I saw was `LoadFileInstruction`, where we transform the full path of a require statement into a Symbol. These symbols should never occur anywhere else in the code, so that wouldn't cause any duplication.

PR #2054, opened 2024-05-30
Refresh array fixtures in specs

herwinw:
This does enable the `protected` statement again, this line was removed because it triggered an error in Natalie, but that appears to have been fixed.


seven1m (approved):
🎉

PR #2055, opened 2024-05-31
Produce better errors for bad compile time values

seven1m:


seven1m:
That macos regexp + threading error is becoming more frequent. 🤔

PR #2058, opened 2024-06-01
Raise SyntaxError for invalid regexp

herwinw:
This is one of those "do we really want this change" cases.

An invalid Regexp literal currently results in either a AOT compile error, or nothing at all in case the Regexp is not used. This changes that behaviour: an invalid Regexp literal will now generate an hardcoded exception in the output.
This fixes the compile issue with 4 language specs. Even though I kind of like the compile time error we're currently raising, I also like the fact that we're able to run more specs.


seven1m:
Works for me 🤷‍♂️

PR #2063, opened 2024-06-01
Add unicode casemap in String#capitalize

seven1m:
#217

This gets a few more specs passing.

### DIFF
--- a/include/natalie/encoding_object.hpp
+++ b/include/natalie/encoding_object.hpp
@@ -82,9 +82,9 @@ public:
     static EncodingObject *find_encoding_by_name(Env *env, String name);
     static EncodingObject *find_encoding(Env *env, Value encoding);
 
-    static nat_int_t codepoint_to_lowercase(nat_int_t codepoint);
-    static nat_int_t codepoint_to_uppercase(nat_int_t codepoint);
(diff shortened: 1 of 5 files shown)


include/natalie/encoding_object.hpp
```diff
@@ -82,6 +82,10 @@ class EncodingObject : public Object {
     static EncodingObject *find_encoding_by_name(Env *env, String name);
     static EncodingObject *find_encoding(Env *env, Value encoding);
 
+    static nat_int_t ascii_codepoint_to_lowercase(nat_int_t codepoint);
+    static nat_int_t ascii_codepoint_to_uppercase(nat_int_t codepoint);
+    static nat_int_t ascii_codepoint_to_titlecase(nat_int_t codepoint);
+
```
seven1m, line 88: I considered making `codepoint_to_lowercase` et al member functions on EncodingObject and letting the specific encoding subclasses override them. But, I don't see a place where the current string's encoding actually affects the behavior -- it's more about the flags passed into `String#downcase` et al. So I did this for now.

We can make these virtual member functions in the future when needed.

seven1m: Changed my mind a bit, and made `ascii_only` a boolean argument to the existing static case functions. Similar concept, but less code-smelly perhaps.

PR #2072, opened 2024-06-06
Use if_nametoindex to find the ifindex

herwinw:
This resolves an issue on my machine where the VPN interface didn't get an ifindex with the old implementation, but does get one with the new implementation. It reduces the complexity as well.


seven1m (approved):
Yay! This just started happening on my machine as well, with Tailscale, even though I've used it for years. Weird!

PR #2073, opened 2024-06-06
Import language specs for `for`

herwinw:
The compile time errors have been commented out

### DIFF
--- a/spec/language/for_spec.rb
+++ b/spec/language/for_spec.rb
@@ -1,0 +1,192 @@
+require_relative '../spec_helper'
+require_relative 'fixtures/for_scope'
+
+# for name[, name]... in expr [do]
+#   body
+# end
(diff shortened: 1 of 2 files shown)


spec/language/for_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+    end
+    j.should == 6
+  end
+
+  # NATFIXME: Support Prism::ImplicitRestNode
+  xit "iterates over a list of arrays and destructures with empty comma" do
+    #for i, in [[1,2]]
+      #i.should == 1
+    #end
+  end
```
herwinw, line 21: I've written all compile time errors with `xit` block, unless we could run some code that doesn't pass the test. This way the compile time errors still show up as skipped specs.

seven1m (approved):

PR #2087, opened 2024-06-09
Intern frozen string literals

herwinw:
A first attempt at interning frozen strings. This currently works for really basic scripts, since the garbage collector is not aware of these objects. For now, the GC is disabled in this branch.

### DIFF
--- a/include/natalie/global_env.hpp
+++ b/include/natalie/global_env.hpp
@@ -104,6 +104,8 @@ public:
     bool has_file(SymbolObject *name) const { return m_files.get(name); }
     void add_file(Env *env, SymbolObject *name);
 
+    void set_interned_strings(StringObject **, const size_t);
+
     friend class SymbolObject;
 
     virtual void visit_children(Visitor &visitor) override final;
@@ -138,6 +140,9 @@ private:
     Natalie::Object *m_main_obj { nullptr };
 
     ClassObject *m_Encodings[EncodingCount];
+
+    StringObject **m_interned_strings { nullptr };
+    size_t m_interned_strings_size { 0 };
 
     Env *m_main_env { nullptr };
     MethodMissingReason m_method_missing_reason { MethodMissingReason::Undefined };
--- a/include/natalie/global_env.hpp
+++ b/include/natalie/global_env.hpp
@@ -141,8 +141,11 @@ private:
 
     ClassObject *m_Encodings[EncodingCount];
 
-    StringObject **m_interned_strings { nullptr };
-    size_t m_interned_strings_size { 0 };
+    struct span {
+        StringObject **strs;
+        size_t size;
+    };
+    Vector<span> m_interned_strings {};
 
     Env *m_main_env { nullptr };
     MethodMissingReason m_method_missing_reason { MethodMissingReason::Undefined };
--- a/include/natalie/global_env.hpp
+++ b/include/natalie/global_env.hpp
@@ -9,6 +9,7 @@
 #include "natalie/global_variable_info.hpp"
 #include "natalie/method_missing_reason.hpp"
 #include "tm/hashmap.hpp"
+#include "tm/span.hpp"
 
 namespace Natalie {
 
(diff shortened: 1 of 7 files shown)


herwinw:
> Lookin good!

It looks even better on my machine, because that one has all the tests passing...

herwinw:
It looks like I got everything working now (my desktop has probably some more processing power/mem than the CI runners, which might have been the cause of issues not showing up on my machine). I will keep this one in Draft state for a little bit longer, just so I can run some tests to validate that the general case is not slower than before.

herwinw:
`rake test` looks like it slowed down with about 20 sec (22:01 of master, 22:24 with this change), but since I'm also using this machine for different things it's more likely to be unrelated to the change. I should probably repeat this test about 4 times more and take averages, but I'm not really waiting to have my machine run tests for another 3 hours.

I also tested the following script:
```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

str = 'abc'

def foo
  'abc'
end

a = nil # assign to a var to force it into used context
10_000_000.times do
  a = foo
end
```
I compiled this into a binary to isolate the run time: this dropped from 12 seconds in master to 8 seconds with interning.

herwinw:
I would say this works as intended. We could probably optimize a few more methods, for example `StringObject::operator==` could simply start with `if (this == other) return true`, that means in certain cases we can simply compare two addresses and don't have to fall back to the `strncmp` call, and with frozen strings the chance of using the same object increases.

include/natalie/global_env.hpp
```diff
@@ -139,6 +141,12 @@ class GlobalEnv : public Cell {
 
     ClassObject *m_Encodings[EncodingCount];
 
+    struct span {
```
seven1m, line 144: I think we should call this something more descriptive like `StringInternSpan` or some such. Your call on the name, but as is, fully-namespaced, this one is `Natalie::GlobalEnv::span`, which doesn't communicate what it's for.

lib/natalie/compiler/backends/cpp_backend.rb
```diff
@@ -285,6 +289,10 @@ def symbols_declaration
         "static SymbolObject *#{symbols_var_name}[#{@symbols.size}] = {};"
       end
 
+      def interned_strings_declaration
+        "static StringObject *#{interned_strings_var_name}[#{@interned_strings.size}] = {};"
```
seven1m, line 293:
```suggestion
        "static StringObject *#{interned_strings_var_name}[#{@interned_strings.size}] = { 0 };"
```

Can we skip the later `memset()` if we zero initialize the array here?

lib/natalie/compiler/backends/cpp_backend.rb
```diff
@@ -297,6 +305,30 @@ def init_symbols
         end
       end
 
+      def init_interned_strings
+        return [] if @interned_strings.empty?
+
+        # Start with setting the interned strings list all to nullptr and register the GC hook before creating strings
+        # Otherwise, we might start GC before we finished setting up this structure if the source contains enough strings
+        interned_strings = [
```
seven1m, line 313: I was confused for a sec because this `interned_strings` has the same name as the instance variable. Can we rename this to something like `lines` or `code` or whatever?

src/global_env.cpp
```diff
@@ -120,6 +124,9 @@ void GlobalEnv::visit_children(Visitor &visitor) {
         visitor.visit(m_Encodings[i]);
     for (auto pair : m_files)
         visitor.visit(pair.first);
+    for (auto pair : m_interned_strings)
+        for (size_t i = 0; i < pair.size; i++)
+            visitor.visit(pair.strs[i]);
```
seven1m, line 129:
```suggestion
    for (auto set : m_interned_strings)
        for (size_t i = 0; i < set.size; i++)
            visitor.visit(set.strs[i]);
```

I like `set` or `span` here rather than `pair`, since pair implies two.

seven1m (approved):
This is really cool! I just had some nits here and there, but nothing major.

include/natalie/global_env.hpp
herwinw: I think https://github.com/seven1m/tm/pull/11 is the more generic solution here, by extracting a generic Span class that is totally disjoint from StringObject.

herwinw:
I think all the issues have been addressed in the latest version.

PR #2097, opened 2024-06-15
Fix spelling

jbampton:


spec/core/string/shared/to_sym.rb
```diff
@@ -53,7 +53,7 @@
     sym.to_s.should == binary_string
   end
 
-  it "ignores exising symbols with different encoding" do
+  it "ignores existing symbols with different encoding" do
```
herwinw, line 56: I scooped this one in #2098.

The file in the `spec` folder are originally from https://github.com/ruby/spec, we import those file and disable the failing tests (mostly with NATFIXME blocks, visible in some other changes in this commit, sometimes we have to comment some code out due to compile errors). And of course, since we're keeping copies of files from a repository, we often get out of date.

This change has spelling fixes in two spec files that are not related to our local changes (this file and `truncate.rb`). I checked the upstream repository to see if the spelling fixes had been applied there, and both of them were (if they weren't, it would make more sense to submit them in the source repo instead of our copy). But this file had a few additional updates as well, so it seemed more logical to just import the file again.
It looks like the additional changes in the other file are a few lines apart, so I guess we can merge both of them without merge conflicts.

spec/shared/rational/truncate.rb
```diff
@@ -55,7 +55,7 @@
     end
   end
 
-  describe "with an invalid valud for precision" do
+  describe "with an invalid value for precision" do
```
herwinw, line 58: Just for reference: this one has been applied upstream as well, there are no other differences between this file and the upstream version.

PR #2104, opened 2024-06-15
Use new, better semantics for Hash#dup

seven1m:


test/natalie/hash_test.rb
```diff
@@ -110,6 +110,13 @@
       hash.default.should == 1
       hash.dup.default.should == 1
     end
+
+    it 'copies the compare_by_identity flag' do
+      hash = { a: 1 }
+      hash.compare_by_identity
+      dup = hash.dup
+      dup.compare_by_identity?.should == true
+    end
```
herwinw, line 119: https://github.com/ruby/spec/blob/master/core/hash/compare_by_identity_spec.rb#L94

This has an upstream spec (that is not in our local copy of this file)

herwinw (approved):

PR #2106, opened 2024-06-15
Refresh shared code for String#chars and String#each_chars specs

seven1m:
#217


herwinw (approved):

herwinw:
It looks like these are all in preparation for the default frozen string literals of Ruby 3.4, so this change should have no impact on the current behaviour.

PR #2110, opened 2024-06-16
Refresh specs for `for`

herwinw:
This adds two new compile errors


herwinw:
@seven1m  I took the liberty of merging this one without approval: it's mostly just importing this file from another repository (and adding the NATFIXME blocks), so it doesn't change any behaviour of our code.

PR #2111, opened 2024-06-16
Refresh specs for String#unpack and String#unpack1

herwinw:


herwinw:
Similar to  #2110:  a review of this change does not add anything.

PR #2114, opened 2024-06-16
Fix spelling

jbampton:


lib/linenoise/linenoise.c
```diff
@@ -75,7 +75,7 @@
  *
  * DSR (Device Status Report)
  *    Sequence: ESC [ 6 n
- *    Effect: reports the current cusor position as ESC [ n ; m R
+ *    Effect: reports the current cursor position as ESC [ n ; m R
```
seven1m, line 78: `linenoise` is an upstream source so I don't think we want to change anything in it. It would make harder to update it later.

seven1m (approved):

PR #2115, opened 2024-06-16
Make String#index mostly spec-compliant

seven1m:
#217 

This gets most of the specs passing, except a couple encoding ones.


seven1m:
I’m going to look at this again. I wonder if we should delegate to Regexp#=~ since it does basically the same thing? (Or does it?)

PR #2116, opened 2024-06-16
First step to support pattern matching

herwinw:
This supports the following syntax:
```ruby
expression => variable
```
Which is pretty much just an assignment written the other way around.

This is the first step in a very long process to add pattern matching to Natalie

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -1999,6 +1999,20 @@ module Natalie
         instructions
       end
 
+      def transform_match_required_node(node, used:)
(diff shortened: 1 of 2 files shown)


seven1m (approved):
I went down this road in #1571 and didn't like where I went with it. I remember thinking that the recursive nature of some of those patterns might require a whole new separate compiler just for patterns. 😬 

But you might have better thoughts on this!

herwinw:
To be fair, I'm probably going to refactor this another 50 times, then abandon it because there's something new and shiny :rofl:

PR #2118, opened 2024-06-16
Order `.rubocop.yml` file

jbampton:


herwinw:
Makes you wonder why there is no option in Rubocop to check this.

PR #2120, opened 2024-06-17
Support constants in pattern matching RHS

herwinw:
This includes the first refactor of the pattern matching code (as promised :sweat_smile: ), and I'm already convinced I should extract this method into its own class, so refactor 2 is coming up


seven1m (approved):
This is cool. I did have another thought about a ruby-to-ruby compiler, which I posted about in Discord, but this is fine for now.

PR #2124, opened 2024-06-18
Fixup of test/natalie folder

herwinw:
* Remove a test file that was probably used for MRI/Natalie comparison, that doesn't work anymore
* Rename the to_enum test to include it in the automated test suite.


seven1m (approved):
Whoops!

PR #2125, opened 2024-06-19
Extract transform_match_required_node into its own class

herwinw:
This will make it easier to rewrite this code, especially since we might be rewriting this code into something that works on a different level.

This is in preparation of the plans to generate Ruby code. Since that code will work at a totally different level than the instructions generated in Pass1, it will be cleaner to split that up. This would make it easier to add tests for the transformer as well.

### DIFF
--- a/lib/natalie/compiler/transformers/match_required_node.rb
+++ b/lib/natalie/compiler/transformers/match_required_node.rb
@@ -10,7 +10,7 @@ module Natalie
           @compiler = compiler
         end
 
-        def visit_constant_path_node(node)
+        def method_missing(_method, node)
           [
             compiler.transform_expression(value, used: true),
             DupInstruction.new, # Required for the error message
@@ -63,7 +63,10 @@ module Natalie
             PopInstruction.new,
           ]
         end
-        alias visit_constant_read_node visit_constant_path_node
+
+        def visit_array_pattern_node(node)
+          raise SyntaxError, "Pattern not yet supported: #{node.inspect}"
+        end
 
         def visit_local_variable_target_node(node)
           [
@@ -76,10 +79,6 @@ module Natalie
         def visit_match_required_node(node)
           @value = node.value
           node.pattern.accept(self)
-        end
-
-        def method_missing(_method, node)
-          raise SyntaxError, "Pattern not yet supported: #{node.inspect}"
         end
       end
     end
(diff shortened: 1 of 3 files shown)


herwinw:
And of course during the day, when you're doing something completely unrelated, you suddenly realize that with a slight change you can support many more cases.

lib/natalie/compiler/transformers/match_required_node.rb
```diff
@@ -0,0 +1,86 @@
...
+  class Compiler
+    module Transformers
+      class MatchRequiredNode
+        attr_reader :compiler, :value
+
+        def initialize(compiler)
+          @compiler = compiler
+        end
+
+        def method_missing(_method, node)
```
seven1m, line 13: I suspect you did this to find out what method Prism was calling... 😄

herwinw: No, that one is actually easier to do without a `method_missing`, since it will show the `NoMethodError` with all the information

I've been using the visitor pattern of Prism, every node type has an `accept` method that pretty much looks like this:
```ruby
def accept(visitor)
  visitor.send("visit_#{type}", self)
end
```
We always start with calling `visit_match_required_node`, where we use this pattern to visit the `accept` (RHS) part of the statement. We've got roughly three options here:
* ArrayPatternNode, for `expr => [1, y]`, this is not supported and raises an explicit exception
* LocalVariableTargetNode, for `expr => a`, this is handled as being an assignment with the visitor pattern in this class
* Anything else, which can be a class, an integer literal, a range, or pretty much anything: this is all rewritten from `expr => target` to something like `raise NoMatchBlah unless target === expr`, and the compiler ivar (the `Pass1` instance) is used to convert the `expr` and `target` to the relevant instructions (the calls to `compiler.transform_expression`)

The first commit has an explicit `alias visit_constant_read_node visit_constant_path_node` to handle both constants and constant paths, even though the method `visit_constant_path_node` has not a single reference to constants, everything is delegated to `Pass1` via `compiler.transform_expression`
The second commit introduces the `method_missing` handler, which handles the visitation of every not explicitly handled node type, which means we use this code in the general case and no longer have to worry about the node type.

it could probably be made a bit more secure, with `return super(method, node) if method != :"visit_#{node.type}"` and handling arities other than 1, but given that this code is close to being rewritten anyway, I think we're good for now.

PR #2129, opened 2024-06-20
Remove accidental copy of String#encode spec

herwinw:
This one has been added in  #2127, and is most likely the result of `:w'` in vim (at least: that's how I usually create them)


seven1m (approved):
Heh! Thanks.

PR #2141, opened 2024-06-25
Honor source encoding argument in String#encode

seven1m:
#217


include/natalie/encoding/ibm437_encoding_object.hpp
```diff
@@ -35,6 +35,11 @@ class Ibm437EncodingObject : public EncodingObject {
     virtual nat_int_t decode_codepoint(StringView &str) const override;
 
     virtual bool is_single_byte_encoding() const override final { return true; }
+
+private:
+    static TM::Hashmap<nat_int_t, nat_int_t> &to_unicode_map();
```
herwinw, line 40: Can we make this a const reference? I think callers of the method should not be allowed to modify its content.

PR #2145, opened 2024-06-26
Pattern match other than assignments

herwinw:
I think we're good now to start working on the `case in` syntax


seven1m (approved):

herwinw:
I'm running into multiple issues with the self hosted compiler, I created a bug issue for the first thing I encountered, but now the specs still fail. This needs some more investigation, and probably some more bug fixes to Natalie itself.

PR #2153, opened 2024-06-26
Fix behaviour of Enumerator#with_index with block

### LINKED ISSUE #2150: Calling with_index with a block on an enumerator results in the wrong output type
Observed in  #2145 where it broke our self hosted compiler. This snippet:
```ruby
p [:a].map.class
p [:a].map.with_index.class
p [:a].map.with_index { |value, idx| [value, idx] }.class
p [:a].map.with_index { |value, idx| [value, idx] }.to_a.class
p [:a].map.with_index { |value, idx| [value, idx] }
p [:a].map.with_index { |value, idx| [value, idx] }.to_a
```
The output of MRI (I used version 3.3.3, but it should be the same in any supported version):
```
Enumerator
Enumerator
Array
Array
[[:a, 0]]
[[:a, 0]]
```
The output of the same program ran with Natalie:
```
Enumerator
Enumerator
Enumerator
Array
#<Enumerator:0x564692b83c60>
[:a]
```
The call to `Enumerator#with_index` is supposed to return a new Enumerator if no block is given, but should return an array if a block is given. 
Additionally, the output of `[:a].map.with_index { |value, idx| [value, idx] }.to_a` is the original array, not the mapped version.

herwinw:
Fixes #2150

### DIFF
--- a/spec/core/enumerator/each_with_index_spec.rb
+++ b/spec/core/enumerator/each_with_index_spec.rb
@@ -1,0 +1,36 @@
+require_relative '../../spec_helper'
+require_relative '../../shared/enumerator/with_index'
+require_relative '../enumerable/shared/enumeratorized'
+
+describe "Enumerator#each_with_index" do
+  it_behaves_like :enum_with_index, :each_with_index
+  it_behaves_like :enumeratorized_with_origin_size, :each_with_index, [1,2,3].select
+
+  it "returns a new Enumerator when no block is given" do
+    enum1 = [1,2,3].select
+    enum2 = enum1.each_with_index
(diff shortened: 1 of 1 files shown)


herwinw:
Mental note: this added spec file is called `each_with_index_spec.rb`, but includes a few specs for `Enumerator#with_index`. This should probably be fixed upstream.

herwinw:
> Mental note: this added spec file is called `each_with_index_spec.rb`, but includes a few specs for `Enumerator#with_index`. This should probably be fixed upstream.

This has been fixed upstream, we will get to this at some point, but I don't really need this change any more than any other bugfix.

PR #2160, opened 2024-06-28
Make String#freeze spec-compliant

seven1m:
#217

Today I learned:

```ruby
'foo'.freeze.equal?('foo'.freeze) # => true
```

Interesting to note that it only seems to apply to string literals and doesn't work if you set a variable first:

```ruby
s = 'foo'
s.freeze.equal?('foo'.freeze) # => false
```

(Good thing too, because I have no idea how we'd do that with the compiler! 😅)

### DIFF
--- a/lib/natalie/compiler/backends/cpp_backend/transform.rb
+++ b/lib/natalie/compiler/backends/cpp_backend/transform.rb
@@ -158,7 +158,7 @@ module Natalie
 
         def interned_string(str, encoding)
           index = @interned_strings[[str, encoding]] ||= @interned_strings.size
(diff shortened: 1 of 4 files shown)


herwinw:
> (Good thing too, because I have no idea how we'd do that with the compiler! 😅)

Add a hashmap with the frozen strings, in case you freeze a variable that contains a string, look the actual value up in the vector, add it if it doesn't exist, and assign the address of the StringObject from the map to the variable.
I don't think this would be very good for your performance though.

seven1m:
This was really easy thanks to the work in #2087. 🤗

PR #2169, opened 2024-06-29
Raise SyntaxError for pattern match with pinned expression

herwinw:
Which is something I didn't even knew existed until just now.


seven1m (approved):
Fun!

PR #2172, opened 2024-06-29
Start work on String#byteindex

seven1m:
#217


include/natalie/object.hpp
```diff
@@ -376,7 +376,11 @@ class Object : public Cell {
     IoObject *to_io(Env *env);
     IntegerObject *to_int(Env *env);
     StringObject *to_s(Env *env);
+
+    // old error message style
     StringObject *to_str(Env *env);
+    // new error message style
```
herwinw, line 382: I would prefer an example of an error message with these comments, that way if I'm searching for the correct way to produce "no implicit conversion of {} into String" I know which one to choose.

PR #2174, opened 2024-06-29
Mass spec refresh

herwinw:
I used a script to find all fully passing upstream specs and copy them from a local checkout.


herwinw:
For future reference: I hacked this together:

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'json'

spec = JSON.load_file('current.json')['stats']['spec']

walk_spec = lambda do |part|
  part.flat_map do |key, value|
    if value.key?('compiled')
      if value['compiled'] && !value['crashed'] && value['errors'].zero? && value['failures'].zero? && value['success'].positive? && !value['timeouted']
        [key]
      else
        []
      end
    else
      result = walk_spec.call(value)
      if result.empty?
        []
      else
        result.map { |rest| "#{key}/#{rest}" }
      end
    end
  end
end

passing = walk_spec.call(spec)
passing.each do |file|
  `cp ../ruby-spec/#{file} spec/#{file}`
end
```

herwinw:
This one includes some changes likes:
```diff
-  # NATFIXME: Inconsistent results. We can probably fix this once we implement IO#close_read
-  xit "raises IOError on stream not opened for reading" do
+  it "raises IOError on stream not opened for reading" do
     -> { STDOUT.ungetbyte(42) }.should raise_error(IOError, "not opened for reading")
   end
```
So I probably need to get through all of them before I feel comfortable merging this.

seven1m (commented):
Dang!

herwinw:
Additional updates:
* Import a bunch of shared/fixtures that were required to run the updated specs
* Reverted a few changes, sometimes the NATFIXME block was in a darwin specific part, and is skipped in the nightly CI run
* Removed the files that no longer exist upstream

spec/library/set/shared/.collect.rb.swp
seven1m, line 1: Should we remove this or leave it to match upstream?

seven1m:
Sweet.

PR #2186, opened 2024-07-03
Even more self assignment fixes

herwinw:
```ruby
self.foo += bar
self[foo] += bar
```


herwinw:
I think these two are the last ones

seven1m (approved):

PR #2189, opened 2024-07-04
Typo fix: spallted => splatted

herwinw:


herwinw:
I'm merging this without a review, I don't think a review has any added value for a change like this.

PR #2191, opened 2024-07-04
Import some upstream typo fixes in specs

herwinw:
Mostly courtesy of @jbampton


herwinw:
https://github.com/ruby/spec/pull/1178

herwinw:
Since this is an import of some existing code, I don't think a review would add anything.

PR #2192, opened 2024-07-05
Remove unneeded duplicate words

jbampton:


src/dtoa.c
```diff
@@ -167,7 +167,7 @@
  *	FREE_DTOA_LOCK(...) are avoided; instead each thread with thread
  *	number < n has a separate copy of relevant data structures.
  *	After set_max_dtoa_threads(n), a call set_max_dtoa_threads(m)
- *	with m <= n has has no effect, but a call with m > n is honored.
+ *	with m <= n has no effect, but a call with m > n is honored.
```
herwinw, line 170: I would prefer to not alter this file. We simply use an import of https://www.netlib.org/fp/, so we are not the source for this. Adding changes only add confusion when we have to compare it to upstream.

seven1m: I had the same thought, but it's only a single word, so I merged it. If you want to change it back, I'm not opposed.

herwinw: Given that the last update of that file is listed as 2013, I think it's safe to expect that we are never going to update this file. So maybe we shouldn't be overthinking this :sweat_smile:

PR #2198, opened 2024-07-09
Make Kernel.rand a module_function

herwinw:


herwinw:
Since we're pretty much just forwarding the calls from `Kernel.rand` to `Random::DEFAULT.rand`, the failing specs have to be checked to see if it's specifically for `Kernel.rand` or if it's relevant for `Random#rand` as well.

PR #2200, opened 2024-07-09
Implement first version of IO.popen

herwinw:
For now its functionality is pretty limited, but a number of unrelated specs depend on this method.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -80,6 +80,7 @@ public:
     bool is_nonblock(Env *) const;
     bool isatty(Env *) const;
     int lineno(Env *) const;
+    Value pid(Env *) const;
     static Value pipe(Env *, Value, Value, Block *, ClassObject *);
     static Value popen(Env *, Args, Block *, ClassObject *);
     int pos(Env *);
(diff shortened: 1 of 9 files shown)


seven1m:
I wonder if there's a way to do popen without adding a FILE* to our IO object. I don't think CRuby has a FILE* on the IO struct. Maybe they implement their own popen instead of using the system's?

Not a blocker for me; progress is progress. I just worry about the juggling of both a fileno and FILE*.

herwinw:
I expect MRI does this the other way around: save a `FILE*`, instead of an `int fd`. (related: #1395)

herwinw:
https://github.com/ruby/ruby/blob/master/include/ruby/io.h#L143-L154

Looks like it has both

herwinw:
Another issue: `popen` is unidirectional by design, so we probably can't use that anyway for the `r+` access mode (we have a `popen2` in `src/natalie.cpp` that could be adapted to become bidirectional and solve this issue). That one now returns a `FILE*`, but does this by converting a file description using `fdopen(3)`. We can probably get around without the `FILE*` in the IO object (but we would still need to save the pid).

Marking this as draft for now, rewriting this to use a modified `popen2` might solve the Macos issues I'm seeing in the CI that I can't debug.

seven1m:
OK, thanks for explaining that to me and for linking to the CRuby source. I was looking at the wrong struct! Having both FILE* and fd makes sense now.

herwinw:
This now supports the basic functionality. I think this is good enough to get merged now, the next two steps are getting the additional arguments/argument types to work, and change the pipe to allow bidirectional communication.

seven1m (approved):
Nice!

Issue #2209, opened 2024-07-12
Proc#ruby2_keywords should not capture self when used as a method


seven1m:
Defining a method using a proc that had `ruby2_keywords` called on it should not capture `self`. I only found this because I'm working on adding [delegate.rb](https://github.com/ruby/ruby/blob/master/lib/delegate.rb) which uses something like this.

```ruby
describe '#ruby2_keywords' do
  it 'does not capture self when used as a method' do
    x = 9
    klass = Class.new do
      define_method(:plain, ->(*) { [self, x] })
      define_method(:wrapped, ->(*) { [self, x] }.ruby2_keywords)
    end
    obj = klass.new
    obj.wrapped.first.should == obj
    obj.wrapped.last.should == 9
    obj.plain.should == obj.wrapped
  end
end
```

PR #2210, opened 2024-07-12
Add Delegate library

seven1m:
I had to work around [one bug](https://github.com/natalie-lang/natalie/issues/2209), but mostly this just worked!


herwinw:
Nice, I remember looking into this one, it broke due to the lack of `Kernel#!~`, so that's what resulted in #1862.

seven1m:
Cool! 😎

PR #2222, opened 2024-08-26
Fix Regexp.escape/quote not escaping dollar signs

timcraft:
Dollar signs should be escaped/quoted:

    $ irb
    irb(main):001> Regexp.escape '$'
    => "\\$"

### DIFF
--- a/spec/core/regexp/shared/new.rb
+++ b/spec/core/regexp/shared/new.rb
@@ -56,9 +56,7 @@ describe :regexp_new_string, shared: true do
   end
 
   it "raises a RegexpError when passed an incorrect regexp" do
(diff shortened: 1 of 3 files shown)


herwinw:
The change looks good. Could you also remove the `NATFIXME` block in `spec/core/regexp/shared/new.rb` line 59 (just remove that line and the corresponding `end` line, and fix the indentation of the current line 60). You've accidentally fixed that bug too ;)

It would be even better if we could include the upstream spec fix too, but that's out of our control.

herwinw (approved):

timcraft:
Thanks! I mistakenly assumed the build was broken 😆

PR #2224, opened 2024-09-19
Refresh shared spec for Regexp.quote/Regexp.escape

herwinw:
This includes a spec for the escape of `$`, which means we can remove our local spec for this.

This finishes the leftovers of #2222, which depended on an upstream merge in ruby/spec.


seven1m (approved):

herwinw:
Rebased to fix the MacOS issue

PR #2225, opened 2024-09-20
Refresh spec for Thread#group

herwinw:
Now with actual tests

PR #2226, opened 2024-09-20
Add some motivation to the CI job on MacOS

herwinw:
It currently fails with this message:

    Error: Refusing to uninstall /usr/local/Cellar/ruby/3.3.5
    because it is required by asciidoctor, which is currently installed.
    You can override this and force removal with:
      brew uninstall --ignore-dependencies ruby


herwinw:
This should fix the failing CI of #2224 and #2225

PR #2227, opened 2024-09-20
Throw exception in Regexp.new/compile for trailing single backslash

herwinw:


herwinw:
The CI will fail on MacOS due to #2226

seven1m (approved):

herwinw:
Rebased with the MacOS fix

PR #2228, opened 2024-09-20
Fix invalid hex escape in Regexp

herwinw:


seven1m (approved):

herwinw:
Welp, it turns out `\xa` and '\xap` are valid escaped, which get interpreted as `\x0a` and `\x0ap`. Guess I learned something today.

PR #2229, opened 2024-09-20
Raise a RegexpError if the \u{} escape is empty

herwinw:


herwinw:
This one had a rewrite after the changes of #2249. The Natalie regexp test now works with the self hosted compiler too.

PR #2231, opened 2024-09-20
Enable commented specs in regexp_test

herwinw:
This has probably been fixed in #2058.


seven1m (approved):

herwinw:
I'm going to close this one. Things break in the self hosted compiler and this really takes a different code path there. Fixing this is rather complex, and not something I see myself doing anytime soon.

PR #2234, opened 2024-09-21
Enable the formerly crashing spec of Regexp#encoding again

herwinw:
This has been fixed somewhere in the past (at least: that what it looks like).


herwinw:
This might be a "works on my machine", so I still need to check if the CI agrees with my observation

PR #2236, opened 2024-09-24
Update Set#merge for Ruby 3.3

herwinw:
This does include the changes of https://github.com/ruby/spec/pull/1194, marking this as draft for now, the spec might need to be imported again if upstream requires changes.


herwinw:
Verbatim accepted upstream, so no updates are required.

PR #2249, opened 2024-10-05
Additional Unicode escape validation in Regexp.new/Regexp.compile

herwinw:


herwinw:
Note to self: https://github.com/ruby/spec/pull/1199 adds the missing spec for `\u{abX}`, if that one is accepted/merged upstream we can remove our local test too.

PR #2262, opened 2024-10-11
Import the CGI module 

richardboehme:
Apart from the CGI module (and friends) this changes the following things:

* Make Array#initialize more correct
* Implement Kernel#__callee__
* Make Kernel#__method__ spec compliant
* Implement the equal_element spec matcher

Apart from those changes the CGI module only needed one small modification in cgi/util.rb. It feels a bit hacky to use `__inline__` like that but I could not think of a different easy way of fixing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ the respective file, in the same directory with a name like `LICENSE`, or both.
 | file(s)          | copyright                         | license           |
 | ---------------- | --------------------------------- | ----------------- |
 | `bigint.*`       | 983                               | Unlicense         |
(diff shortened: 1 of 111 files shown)


richardboehme:
I had to change quite some things with my change to the Method object. Not sure if my solution is ideal but I have no other idea on how to implement Kernel#__callee__ without keeping track of the original method and creating a new method object for aliases.

seven1m (approved):
This looks ok to me. I might take another peek at it on my desktop tomorrow before merging. Great work.

PR #2263, opened 2024-10-12
Switch to Prism 1.0

herwinw:
This is going to take a while, the interface for creating the nodes has changed.

Once we've finished this one, the upgrade to Prism 1.2 should be easy again.


herwinw:
This one now passes the full test suite with the self hosted compiler (only took about 90 minutes to run that). I would say it's good to go.
I will look at the upgrade to 1.2 somewhere next week, but I've have seen enough Prism for today ;)

PR #2265, opened 2024-10-13
Upgrade to Prism 1.1

herwinw:
The yak of the day: adding [enum](https://github.com/ffi/ffi/wiki/Enums) support in our FFI lib. I didn't even know FFI supported this.


herwinw:
Tested it locally with `rake test_self_hosted_full` as well, and it works. Added bonus: two specs for invalid syntax now pass.

Upgrading to Prism 1.2.0 did not work out of the box, so I'm going to look at that another time.

PR #2270, opened 2024-10-25
Reopen module as class

herwinw:
```ruby
class Kernel; end
```
This code currently crashes with an assertion error, it should raise the same TypeError as MRI.


herwinw:
It looks like there are no upstream specs for this code (at least the error message cannot be found in the specs repo), so it might be worth it to add those upstream as well.

PR #2271, opened 2024-10-25
Import ERB::Util

richardboehme:
We will probably not support erb itself (at least not with dynamic strings), but the Erb::Util module is used a lot to escape HTML content and seems to run perfectly fine with Natalie.


seven1m:
Nice!

PR #2272, opened 2024-10-25
Implement BasicObject#instance_exec and improve #instance_eval

richardboehme:
I feel like the implementation to implement the scoping requirements of instance evaling code is quite brittle now, but I didn't have a better idea on how to implement this properly.

If you have any ideas about this let me know.

### DIFF
--- a/include/natalie/block.hpp
+++ b/include/natalie/block.hpp
@@ -48,8 +48,12 @@ public:
     void set_calling_env(Env *env) { m_calling_env = env; }
     void clear_calling_env() { m_calling_env = nullptr; }
 
-    void set_self(Value self) { m_self = self; }
+    void set_self(Value self) {
+        m_original_self = m_self;
+        m_self = self;
+    }
     Value self() const { return m_self; }
+    Value original_self() const { return m_original_self; }
 
     void copy_fn_pointer_to_method(Method *);
 
@@ -57,6 +61,7 @@ public:
         visitor.visit(m_env);
         visitor.visit(m_calling_env);
         visitor.visit(m_self);
+        visitor.visit(m_original_self);
     }
 
     virtual void gc_inspect(char *buf, size_t len) const override {
@@ -69,6 +74,7 @@ private:
     Env *m_env { nullptr };
     Env *m_calling_env { nullptr };
     Value m_self { nullptr };
+    Value m_original_self { nullptr };
     BlockType m_type { BlockType::Proc };
 };
 
--- a/include/natalie/block.hpp
+++ b/include/natalie/block.hpp
@@ -49,11 +49,9 @@ public:
     void clear_calling_env() { m_calling_env = nullptr; }
 
     void set_self(Value self) {
-        m_original_self = m_self;
         m_self = self;
     }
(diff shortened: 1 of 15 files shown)


seven1m (approved):
The `caller()->caller()->module()` thing is a bit weird, but I like that it's called out in the comment and we can improve in the future. I wanted @herwinw's eyes on this just in case he sees a better way to handle that...

herwinw:
My instinct tells me this is going to bite us in the butt very soon, but I have absolutely nothing to back that up.

I will try to get a thorough read of the change tomorrow, because I don't think I've ever seen this part of the code before

herwinw:
https://github.com/natalie-lang/natalie/commit/7631c2826573a9c562a154aa1f570aa8f9d1b71e

I found the first commit a bit hard too read, since it tried to do two things. This commit extracts the arity change, which means the next commit will be a bit shorter. I found this easier to read.

```c++
// block.hpp
void set_self(Value self) {
    m_original_self = m_self;
    m_self = self;
}

// object.cpp, in instance_eval
block->set_self(this);
// run the block
block->set_self(block->original_self());
```
This last line effectively swaps the values of `m_original_self` and `m_self`. This means the old context get preserved, and if we call a block a second time in an `instance_eval` context, we might end up with the wrong context:
```ruby
class KlassWithSecret
  def initialize(secret)
    @secret = secret
  end
end
k99 = KlassWithSecret.new(99)
k77 = KlassWithSecret.new(77)
block = lambda do |obj|
  if obj == k99
    k77.instance_eval(&block)
  else
    @secret
  end
end
p k99.instance_eval(&block)
k77.instance_eval { p block.call(self) }
```
The output in MRI, JRuby and Truffleruby (all the latest stable versions) is `77` and `nil`, but Natalie with this branch returns `77` twice. If you remove the `k99.instance_eval` line, the output is `nil` for both runtimes.

My first  thought was to reset the `m_self` value instead:
```c++
// block.hpp
void restore_original_self() {
    m_self = m_original_self;
    m_original_self = nullptr;
}
```
And replace the line `block->set_self(block->original_self());` with `block->restore_original_self();`. This prints 77 only once, but that's because it terminates with a segfault before it can print the second value. So I guess we need a stack of selfs.

richardboehme:
Mh I feel like having nested `instance_eval`s will break the current implementation nonetheless, because after executing the inner `instance_eval` we will reset `GlobalEnv::the()->instance_evaling()`  and the code executed afterwards thinks it will not be in an instance evaluation. Maybe one way to refactor this is implementing a stack of instance eval contexts on the global env. One of those context objects could hold the env in which `instance_eval` was called and the original self of the block. This way we could eliminate the "original self" on the block and the `caller()->caller()` hack. What do you think?

richardboehme:
@herwinw @seven1m I quickly implemented the idea that I described earlier. Now the script from @herwinw prints the same value in MRI and in Natalie! Let me know what you think about this solution!

seven1m:
Nice!

herwinw (approved):
That looks like the better alternative to me, approved.

PR #2273, opened 2024-10-26
Add support for anonymous block forwarding

richardboehme:
I've implemented this in the simplest way I could think of. Let me know if adding new instructions makes sense or if there are different approaches on how we could implement anonymous forwarding. I'd also like to use this PR as a template for anonymous splat and keyword forwarding.


seven1m:
Nice!

PR #2274, opened 2024-10-26
Add devbox config

seven1m:
I've been using [Devbox](https://www.jetify.com/devbox) for some other projects and I wanted to see if using it here could simplify the setup process for new devs.

TODO: there are few failing specs (when run within a devbox shell) I need to figure out.

- [x] `::crypt()` gives `EINVAL`
- [ ] `spec/core/file/flock_spec.rb` segfault
- [ ] `spec/core/io/write_spec.rb` `Errno::ENOENT`


src/string_object.cpp
```diff
@@ -1085,7 +1085,9 @@ Value StringObject::crypt(Env *env, Value salt) {
     if (strlen(salt_str->c_str()) < 2)
         env->raise("ArgumentError", "salt too short (need >=2 bytes)");
 
-    return new StringObject { ::crypt(c_str(), salt_str->c_str()) };
+    auto crypted = ::crypt(c_str(), salt_str->c_str());
+    assert(crypted);
```
herwinw, line 1089: According to `man crypt(3)`: `set errno when they fail.`, so this could be turned into:
```ruby
if (!crypted)
  env->raise_errno();
```
This might give a bit more information about the source of the error.

seven1m:
I'm closing this agian.

There were two issues:

1. Clang compiled for Nix is just too slow. From my limited understanding, it's due to reproducibility -- certain optimizations are disabled. On my machine, the following command (part of our build) goes from 32 seconds to 2:53. The test suite takes 4-5x longer to run. It's painful. GCC is even worse... I didn't even measure it.
2. There is a bug in our GC exposed by these versions of Clang and GCC. I'll file that separately...

There is some good work in this PR that I plan to ship as separate PRs, just without the devbox config.

PR #2285, opened 2024-11-04
Support fixed string argument with instance_eval

herwinw:
Marking as draft for now, the code works but I'm not that satisfied with how it looks.

Transform instance_eval with a fixed string into a block at compile time, similar to the eval macro handling.


herwinw:
Moved the code to `macros.rb` and shared as much code with the `eval` macro as possible. This should be ready now

PR #2286, opened 2024-11-04
Run CI on MacOS latest

herwinw:
> The macOS-12 environment is deprecated, consider switching to
> macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see
> https://github.com/actions/runner-images/issues/10721

Using the old version now occasionally fails in CI, depending on what CI runner is selected.


herwinw:
The `export path` command has been copied from the output of the installer. I replaced the `ENV` section in the version check with an extra command, this makes it consistent with running the actual checks, and this way we can use the old `$PATH` instead of having to repeat its value.

Issue #1734, opened 2024-01-26
defined? should not call const_missing


herwinw:
I ran into this issue when working on  #1733.

The specs of https://github.com/ruby/spec/blob/master/language/optional_assignments_spec.rb have the following code (simplified):
```ruby
class ClassA
  def self.const_missing(const)
    const
  end

  OR_ASSIGNED_CONSTANT1 ||= :assigned
end

p ClassA::OR_ASSIGNED_CONSTANT1
```
This prints `:assigned` in MRI, but ``:OR_ASSIGNED_CONSTANT1` in Natalie. This code depends on `defined?`
```ruby
p defined?(ClassA::NONE)
```
This prints `nil` in MRI, but `"constant"` in Natalie.
`defined?` is not supposed to call `const_missing`.

herwinw:
https://github.com/ruby/spec/blob/master/language/defined_spec.rb shows some more issues. For example:
```ruby
defined?(retry).should == "expression"
```
This currently results in invalid generated code:
```
$ bin/natalie -e 'defined?(retry)'
/tmp/natalie.cpp20240529-23773-i613wq: In lambda function:
/tmp/natalie.cpp20240529-23773-i613wq:96:17: error: ‘should_retry_’ was not declared in this scope
   96 |                 should_retry_ = true;
      |                 ^~~~~~~~~~~~~
/tmp/natalie.cpp20240529-23773-i613wq:97:17: error: continue statement not within a loop
   97 |                 continue;
      |                 ^~~~~~~~
```
I guess we need a kind of special scope inside a `defined?` argument, where the last expression needs a completely different set of evaluation rules.

herwinw:
https://github.com/ruby/spec/blob/517f06f66d533175d8586d222156323b89a43909/language/defined_spec.rb#L865-L868

The original issue is literally an existing spec (which doesn't show up in the nightly results due to unrelated compilation errors). I think I managed to implement  #1733 and related cases as good as possible, the failures due to this bug are marked as NATFIXME blocks.
In other words: I don't think there is much added value in keeping this explicit case opened, the specs take care of not forgetting this issue. I'm going to close this one

PR #2289, opened 2024-11-05
Make `defined?` spec compliant

richardboehme:
There are probably lots of cases where this still fails and that the specs don't cover. Also I noticed an interesting bug with global regexp matches which is why we will have some NATFIXMEs in the spec.

### DIFF
--- a/lib/natalie/compiler/instructions/global_variable_defined_instruction.rb
+++ b/lib/natalie/compiler/instructions/global_variable_defined_instruction.rb
@@ -15,7 +15,6 @@ module Natalie
 
       def generate(transform)
         transform.exec("if (!env->global_defined(#{transform.intern(@name)})) throw new ExceptionObject")
-        transform.push_nil
       end
 
       def execute(vm)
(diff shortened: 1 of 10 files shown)


herwinw:
I tried this one with the script of #1734, there are still some issues with the const-missing logic. This might have some deeper causes, and it's not blocking for this update, so see it as a FYI.

richardboehme:
> I tried this one with the script of #1734, there are still some issues with the const-missing logic. This might have some deeper causes, and it's not blocking for this update, so see it as a FYI.

Interesting, I can have a closer look into it. All in all `defined?` behaves quite weird so I wouldn't be suprised if there are more cases like this that are not covered in the `defined_spec`

seven1m:
The failing spec is just a few "This issue is fixed" NATFIXMEs 🎉

herwinw:
@richardboehme I took the liberty to push an extra commit to fix those last few issues with the specs, the branch should be ready to merge now.

richardboehme:
> @richardboehme I took the liberty to push an extra commit to fix those last few issues with the specs, the branch should be ready to merge now.

That's fine, thank you! I just haven't been at my computer yet 😅

PR #2290, opened 2024-11-05
Implement Process.times

herwinw:
This was required to get the `Benchmark` module from CRuby to work. After copying this file, the following benchmark works (the first example in the file, have not yet tried anything else):
```ruby
require 'benchmark'

puts Benchmark.measure { "a"*1_000_000_000 }
```
Also, on my machine this takes around 0.5 seconds for CRuby and 13 seconds for Natalie


herwinw:
All the other examples appear to work too, so I added the Benchmark module as well.

seven1m (approved):

PR #2294, opened 2024-11-06
Improve String#* with large multiplier

herwinw:
Create a large enough buffer from the start, prevent all the resizes/reallocs in the old code.

    puts Benchmark.measure { "a"*1_000_000_000 }

On my system, this went from 7.2 seconds to 1.5 seconds (in release build). For comparison: MRI runs this in 0.4 seconds, JRuby in 0.3 seconds, TruffleRuby in 4.5 seconds.

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -970,14 +970,12 @@ Value StringObject::mul(Env *env, Value arg) const {
     if (nat_int && (std::numeric_limits<size_t>::max() / nat_int) < length())
         env->raise("ArgumentError", "argument too big");
 
-    StringObject *new_string = new StringObject { "", m_encoding };
-    if (this->is_empty())
-        return new_string;
-
-    for (nat_int_t i = 0; i < nat_int; i++) {
-        new_string->append(this);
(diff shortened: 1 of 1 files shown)


herwinw:
We could probably improve even more if we would group some operations. For example: `'a' * 4` is now handled like this:
```c++
// c_str is the source object, buf is the result
char buf[5] = { 0 };
memcpy(buf, c_str, 1);
memcpy(buf + 1, c_str, 1);
memcpy(buf + 2, c_str, 1);
memcpy(buf + 3, c_str, 1);
```
This is like saying `a + a + a + a`. But we can use the intermediate results, so another option would be:
```c++
char buf[5] = { 0 };
memcpy(buf, c_str, 1);
memcpy(buf + 1, buf, 1);
memcpy(buf + 2, buf, 2); // buf == "aa", so copy two bytes at once.
```
Or: `(a * 2) * 2`. With `n = 4`, this only removed 1 operation, with `n = 8` this will remove 4 operations. Complexity roughly goes down from `O(n)` to `O(log n)`.

seven1m (approved):

PR #2298, opened 2024-11-07
Implement IPSocker#recvfrom

herwinw:
I was playing around with the [ruby-radius](https://rubygems.org/gems/ruby-radius) gem. This one hasn't been updated in over 10 years, and uses some non-idiomatic code. With a few recent changes (the fsync-fix for reading /dev/urandom with a block, the encoding fix for `String#<<`) and this addition, I can now perform a RADIUS authentication with a user/password.


herwinw:
The method appears to cause errors in the macos CI, I resorted to disabling the whole spec when running on macos, since it works fine on my linux machine and I have no way to debug it.

PR #2311, opened 2024-11-08
Add Process.spawn

herwinw:
By binding it to KernelModule::spawn. This adds a boatload of additional failing specs because there are approximately 2344239 ways to call it.


seven1m (approved):

herwinw:
For those who like turtles all the way down, I looked into some of the failing specs that break on the options parser. This is caused by trying to run some code with spaces in it:
```ruby
Process.wait Process.spawn(ruby_cmd("print Process.getpgid(Process.pid)"))
````
We first feed this code to `ruby_cmd:
```ruby
def ruby_cmd(code, options: nil, args: nil)
  binary = ENV['NAT_BINARY'] || 'bin/natalie'
  "#{binary} #{options} -e #{code.inspect} #{args}"
end

# Resulting string: 'bin/natalie  -e "print Process.getpgid(Process.pid)" '
```
This string is being fed to `KernelModule::spawn`, where we split it on spaces:
```c++
auto arg = args.at(0);
arg->assert_type(env, Object::Type::String, "String");
auto splitter = new RegexpObject { env, "\\s+" };
auto split = arg->as_string()->split(env, splitter, 0)->as_array();
const char *cmd[split->size() + 1];
for (size_t i = 0; i < split->size(); i++) {
    cmd[i] = split->at(i)->as_string()->c_str();
}
cmd[split->size()] = nullptr;
```
The result behaves like this:
```c
#include <spawn.h>
#include <stdio.h>
#include <sys/wait.h>

extern char **environ;

int main(const int argc, const char **argv) {
    char *const cmd[] = { "bin/natalie", "-e", "\"print", "Process.getpgid(Process.pid)\"", "", NULL };
    int pid;
    int result = posix_spawnp(
        &pid,
        cmd[0],
        NULL,
        NULL,
        cmd,
        environ); 
    if (result != 0)
        perror("posix_spawnp");
    int status;
    result = waitpid(pid, &status, 0);
    if (result < 0)
        perror("waitpid");
    return 0;
}
```
The ruby code string has now been split into separate parts, so `-e '"print'` is the actual code run, the rest gets thrown into the argument parser, and that breaks on the empty string that is the last argument. If we remove that one, it results in a parser error because the code only has a single quote.

The things you discover when you're trying to do something completely else

PR #2316, opened 2024-11-09
Implement Process.wait2

herwinw:


herwinw:
With this method added, the specs for Kernel.fork/Process.fork can be run, so these have been added as well.

seven1m (approved):

Issue #2310, opened 2024-11-08
Add flag to raise SyntaxErrors in `eval` at runtime


richardboehme:
We implement eval with constant strings by compiling those strings during compile time. If an error occurs we fail compilation. This is a pretty nice feature of Natalie to ensure that code in evals is actually valid. However this also deviates from the behavior of ruby which only raises if the eval is actually called. 

This is especially noticeable in our nightly spec runner. Currently there are at least 3 spec files that cannot compile because the spec explicitly tests that specific code raises a SyntaxError by using eval. One example:

```ruby
it "raises a SyntaxError if using the argument in its default value" do
  a = 1
 -> {
    eval "proc { |a=a| a }"
  }.should raise_error(SyntaxError)
end
``` 

What do you think about this?

herwinw:
https://github.com/natalie-lang/natalie/blob/master/lib/natalie/compiler/pass1.rb#L2361-L2380

We do something similar with regexp: parse it at compile time, and generate instructions to raise an exception if it cannot be parsed (#2058).
It's one of those things where we are stuck between being a precompiled language while still trying to be like Ruby.

richardboehme:
Interesting! I think we should unify this and either raise at compile time or at runtime depending on the proposed flag?

seven1m:
I'm cool with a CLI flag to change behavior. Should we default to the most-Ruby-like behavior and have the flag enable "extra" checks? (In reality, the checks would be free, due to how we precompile, but that's probably counterintuitive for a user.)

- Default: runtime error for bad eval
- `--extra-compiler-checks` (or some such): compile time error for bad eval

I also like the idea of using the same approach for the regex check.

richardboehme:
I like this approach, staying as close to "normal" ruby behavior as possible!

PR #2317, opened 2024-11-09
Implement flip flop operator

richardboehme:
TIL this existed :^)

We now also pass the whole language/if_spec!

This whole thing uses some interesting hacks, where I'm not sure if they are nice:
* I reuse the ElseInstruction to fetch the instructions of the two conditions (taken from IfInstruction)
* I hoist the flip flop states to the top of the generated C++ file, not sure if this is nice but we need to store the state somewhere...


herwinw:
This was one of the operators that I wanted to add to #2310 as "we are never going to support this, so maybe we should still be able to run those specs" :sweat_smile:

herwinw:
I added a commit to enable the specs in `precedence_spec.rb` as well, since these now work too.

seven1m (approved):

PR #2321, opened 2024-11-10
Make Args const in a couple of places

tekknolagi:
This otherwise copy-constructs on many calls which calls malloc and so
on and that's not good for function call performance.

Side by side of fib(25) callgrind instructions:

Before:
  summary: 2554791411
After:
  summary: 2167984106

Or about 15% faster.


seven1m:
This is fantastic. Thanks so much. 🤗

PR #2326, opened 2024-11-11
Don't set redo flag on NilObject singleton instance

herwinw:
Use a dedicated temporary object instead. Changing the flags on nil can cause nil to become invalid.


herwinw:
This should probably be done via the environment instead, and we still have issues when running redo outside of a block context, but those are future works.

seven1m (approved):

PR #2329, opened 2024-11-12
Refresh specs in Time

herwinw:
The specs for `Time#<=>` have some additional checks related to time zones. The shared spec for `Time#gmtime` has related updates, but since we don't yet use this shared spec these could be updated without failing tests.

### DIFF
--- a/spec/core/time/comparison_spec.rb
+++ b/spec/core/time/comparison_spec.rb
@@ -55,6 +55,38 @@ describe "Time#<=>" do
     }.should_not complain
   end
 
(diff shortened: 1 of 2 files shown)


herwinw:
This was a terrible yak shave when I was looking at the Marshal dump of Time to use that as a simple use case to support the `#_dump` method. That spec depends on `Time#utc`, so I just stubbed that one with `def utc = self`, and most of the specs for `Time#utc` were passing.

seven1m (approved):

PR #2331, opened 2024-11-14
Reduce process stack usage during GC mark phase

### LINKED ISSUE #2280: Segfault: EXC_BAD_ACCESS in Natalie::Cell::unmark
Hello. I have created Conway's Game of Life in multiple languages, including Ruby. I then like to benchmark the script on different runtimes, like CRuby, JRuby, and TruffleRuby. You can find the Ruby version here: https://github.com/KieranP/Game-Of-Life-Implementations/tree/master/ruby

I have downloaded and compiled Natalie, and can run some basic ruby scripts. But when I try to compile and run my GOL Ruby version with Natalie (`~/natalie/bin/natalie -c play play.rb`), I am getting segfaults.

I understand that Natalie is still WIP, there are many complex features of Ruby that won't be implemented yet. But my GOL implementation is using mostly basic things (conditionals, loops, hashes, arrays, classes, and string concatenation would be about 95% of the code).

I want to help by filing better bug reports for areas aren't working as expected. But all I can see when I run the play.rb script is "zsh: segmentation fault  ./play". Any suggestions on how I might be able to find the cause of the segfault, so that I might be able to file a proper bug report?

My environment info if it helps: Mac OSX Sequioa 15.1 - M3 Macbook Pro

seven1m:
Fixes #2280

For a large tree/graph of Ruby objects, our GC can really blow out the process stack during its marking phase. Here we change the code to _enqueue_ children to walk by adding them to our own `std::stack` data structure, and then walking them in while loop.

I also learned about the `mutable` keyword, which allows us to make these `visit_children` methods `const` again, since they only thing they need to change is the `m_marked` boolean on `Cell`s. Cool!

### DIFF
--- a/include/natalie/abstract_method_object.hpp
+++ b/include/natalie/abstract_method_object.hpp
@@ -19,7 +19,7 @@ public:
     Method *method() { return m_method; }
     int arity() { return m_method->arity(); }
 
-    virtual void visit_children(Visitor &visitor) override {
+    virtual void visit_children(Visitor &visitor) const override {
         Object::visit_children(visitor);
         visitor.visit(m_method);
     }
--- a/include/natalie/args.hpp
+++ b/include/natalie/args.hpp
@@ -66,6 +66,7 @@ public:
     const Value *data() const { return m_data.data(); }
 
     Vector<Value> &vector() { return m_data; }
+    const Vector<Value> &vector() const { return m_data; }
 
     bool has_keyword_hash() const { return m_has_keyword_hash; }
     HashObject *keyword_hash() const;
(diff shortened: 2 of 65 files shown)


include/natalie/abstract_method_object.hpp
```diff
@@ -21,7 +21,7 @@ class AbstractMethodObject : public Object {
 
     virtual void visit_children(Visitor &visitor) override {
         Object::visit_children(visitor);
-        visitor.visit(m_method);
+        visitor.enqueue(m_method);
```
herwinw, line 24: I'm not really sure about this rename. A visitor visits the nodes. That it enqueues the data during the visit is more of an implementation detail, that's now leaking through all of the code.

Another issue: if we want to implement `ObjectSpace.count_objects`, we will probably need another visitor to visit all the nodes just to count them instead of cleaning up. Suddenly that one has to work with `enqueue` too, even if the implementation works without a queue.

seven1m: I'm open to suggestions. Constructive ones.

PR #2338, opened 2024-11-16
Convert method arguments to rvalue references

herwinw:
This is far from finished, so this PR is juts me claiming this rewrite and trying to see if it works and is more efficient.

### DIFF
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -183,7 +183,7 @@ int pclose2(FILE *, pid_t);
 
 void set_status_object(Env *, pid_t, int);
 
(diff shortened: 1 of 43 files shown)


herwinw:
End results for `fib(25)` in callgrind:

Master: 4,908,779,692
This branch: 4,156,365,607

That's another 15% gain.

The class `ArrayObject` has been prepared to completely use rvalue references. The other classes still have to be updated, and will get slightly more efficient once they are converted, but I'm going to leave that as a future work since that are still about 80 files and it's so boring.

seven1m (approved):

tekknolagi:
Very cool!

PR #2341, opened 2024-11-17
The rest of the rvalue refence args

herwinw:
It's not even that much, so we should be able to clean this up completely.


herwinw:
This is mostly a cleanup, the actual performance gain with our fib test was about 0.001 percent, so that's not worth to think about.

seven1m (approved):

PR #2352, opened 2024-11-22
Fix behaviour for circular arguments

herwinw:
Implement the Ruby 3.4 behaviour instead of 3.3, this is easier and more future proof. The spec has a small extra check to restore the NATFIXME blocks once we've fully switched to Ruby 3.4 compatibility.


herwinw:
This fixes the first compile issue in the specs of language/block, language/def and language/lambda. But don't worry, we've got plenty more compile errors to go before we can run these specs.

seven1m (approved):

Issue #2369, opened 2024-12-03
Limit performance test uploads to branch master with the stats key


herwinw:
https://github.com/natalie-lang/natalie/actions/runs/12145595081/job/33867596932
This is a branch in a different repository, that does not have the API envvar.  This should just skip the upload, and not make the test fail.

https://github.com/natalie-lang/natalie/actions/runs/12141067176/job/33852277664#step:6:11611
This is a PR in the main repository, so it does have the envvar, and it uploads the stats to the server. We should probably limit this to the master branch, because unmerged MRs are not relevant for the progress over time.

PR #2370, opened 2024-12-03
Don't fail in perf_test with missing envvar

herwinw:
Still run the tests, but just print the output and don't try to upload.

This fixes the first issue of #2369


herwinw:
This PR is made from a branch in a forked repo, so it doubles as a test for the change.

seven1m (approved):

PR #2371, opened 2024-12-04
Don't push perf stats on branches other than master

herwinw:
This should fix the second issue of #2369.


seven1m:
My intention was to create a web page where one could compare a recent branch to master. Thoughts?

seven1m:
I think I want to give the web page idea a try at least. I think it'd be cool to be able to compare a branch to master and to itself over time, if it's something I'm actively working on. And I'll make a cleanup task that deletes old branch stats so they don't clutter up over time.

PR #2378, opened 2024-12-10
Run tests on GitHub CI without using Docker

seven1m:
I want to see if this will make the build any faster...


seven1m:
Well that was a can of worms and didn't achieve what I wanted. 🤷‍♂️

PR #2381, opened 2024-12-13
Support _dump/_load in Marshal

herwinw:
And add a stub method `Time#_dump`. This resolves the crashes in the nightly specs for Marshal.load/Marshal.restore.


seven1m (approved):
Cool!

PR #2386, opened 2024-12-14
Remove some unused variables

herwinw:
And fix some additional compiler warnings

### DIFF
--- a/lib/natalie/compiler/backends/cpp_backend/transform.rb
+++ b/lib/natalie/compiler/backends/cpp_backend/transform.rb
@@ -201,7 +201,7 @@ module Natalie
 
         def value_has_side_effects?(value)
           return false if value =~ /^Value\((False|Nil|True)Object::the\(\)\)$/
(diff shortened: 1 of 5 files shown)


lib/natalie/compiler/backends/cpp_backend/transform.rb
```diff
@@ -201,7 +201,7 @@ def add_file(name)
 
         def value_has_side_effects?(value)
           return false if value =~ /^Value\((False|Nil|True)Object::the\(\)\)$/
-          return false if value =~ /^Value\([\w_]*symbols\[\d+\]\/\*\:[^\*]+\*\/\)$/
```
seven1m, line 204: I think we set a `var_prefix` with something like "exception_object_" so I'm thinking this should match something like `Value(exception_object_symbols[7])` or whatever.

`var_prefix` is set here: https://github.com/natalie-lang/natalie/blob/master/lib/natalie/compiler/backends/cpp_backend.rb#L177

seven1m (approved):

PR #2395, opened 2024-12-16
Simplify macro_instance_eval

herwinw:
It turns out there is a `copy` method on Prism nodes that removed the need to look into its internals.


seven1m:
Way better!

PR #2399, opened 2024-12-16
Support Exception#cause in Marshal.dump and Marshal.load

herwinw:
This required a bit of inline C++, since there is no public interface to set the cause of an exception.

### DIFF
--- a/src/marshal.rb
+++ b/src/marshal.rb
@@ -271,6 +271,7 @@ module Marshal
       elsif message.ascii_only?
         message = message.b
       end
+      ivars.prepend([:cause, value.cause]) if value.cause
       ivars.prepend([:bt, value.backtrace])
       ivars.prepend([:mesg, message])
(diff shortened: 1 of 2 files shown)


src/marshal.rb
```diff
@@ -606,6 +607,11 @@ def read_object
       elsif object.is_a?(Exception)
         object = object_class.new(ivars_hash.delete(:mesg))
         object.set_backtrace(ivars_hash.delete(:bt)) if ivars_hash.key?(:bt)
+        if ivars_hash.key?(:cause)
+          exception = object
+          cause = ivars_hash.delete(:cause)
+          __inline__ 'exception_var->as_exception_or_raise(env)->set_cause(cause_var->as_exception_or_raise(env));'
+        end
```
seven1m, line 614: Fascinating!

seven1m (approved):

PR #2401, opened 2024-12-16
Don't store values of Struct classes as instance variables

herwinw:


seven1m (approved):
Interesting, that's some wild metaprogramming there. 😅

PR #2425, opened 2024-12-26
Enable UndefinedBehaviorSanitizer and rename "asan" build to "sanitized"

seven1m:


seven1m:
Hmmm, this is taking way too long. I may just have to run this locally from time to time.

PR #2431, opened 2024-12-28
Cancel in-progress tests when new commits are pushed

seven1m:


seven1m:
Of course I made a typo which allowed me to test this out right away! 😄

PR #2445, opened 2025-01-01
Upgrade to Prism 1.3

herwinw:
The retry of #2266, but an upgrade to 1.3 instead. The deduction of the string encoding feels kind of hacky, but I have no idea how else to fix this.

### DIFF
--- a/lib/natalie/parser.rb
+++ b/lib/natalie/parser.rb
@@ -77,7 +77,9 @@ module Natalie
     end
 
     def encoding
-      result.source.source.encoding
+      encoding = result.magic_comments.find { |e| e.key == 'encoding' }
+      return result.encoding if encoding.nil?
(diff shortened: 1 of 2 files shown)


lib/natalie/parser.rb
```diff
@@ -77,7 +77,9 @@ def ast
     end
 
     def encoding
-      result.source.source.encoding
+      encoding = result.magic_comments.find { it.key == 'encoding' }
+      return result.encoding if encoding.nil?
+      Encoding.find(encoding.value)
```
seven1m, line 82: This seems like an OK approach. I have to assume the change in https://github.com/ruby/prism/pull/3172 was for users of Prism as a library and not for language implementations. I'd like to see what TruffleRuby does with this. :popcorn:

PR #2450, opened 2025-01-03
Support /o modifier on regexp

herwinw:


spec/language/regexp/modifiers_spec.rb
```diff
@@ -16,10 +16,8 @@
   end
 
   it "supports /o (once)" do
-    NATFIXME 'it supports /o (once)', exception: SpecFailedException do
-      2.times do |i|
-        /#{i}/o.should == /0/
-      end
+    2.times do |i|
+      /#{i}/o.should == /0/
```
seven1m, line 20: This seems like a silly spec to me because it doesn't indicate _why_ anyone would want the `/o` flag, i.e. surely this isn't the use case for it. 🤔

herwinw: I don't think the Ruby specs are the right place to find the *why*, more the *what* and the *how*.

And there is not much use for this modifier. It only makes sense if your interpolated code is slow or has side effects. A really far fetched example:
```ruby
def re_fib25 = /\A#{fib(25)\z/o

p "1234".match?(re_fib25) # Slow, this calculates fib(25)
p "1234".match?(re_fib25) # Fast, it just returns the regexp
```

PR #2456, opened 2025-01-05
Happy new year

herwinw:
Which raises a bunch of new questions. Most of these have just been "copyright 2023" with a single year, but the GTK example had a "copyright 2019-2022". So I guess everything should be "copyright 2019-2025" now.


seven1m:
Hrmmm, the range is weird. Not sure why I did that. Maybe just to pad out the content of the gtk box haha.

PR #2458, opened 2025-01-05
Support /e option in Regex

herwinw:


herwinw:
I expected this to fix more specs than just these three :shrug:

src/regexp_object.cpp
```diff
@@ -403,7 +403,8 @@ void RegexpObject::initialize_internal(Env *env, const StringObject *pattern, in
     options &= ~RegexOpts::NoEncoding;
 
     UChar *pat = (UChar *)tweaked_pattern.c_str();
-    OnigEncoding enc = fixed_encoding && !no_encoding ? ruby_encoding_to_onig_encoding(pattern->encoding()) : ONIG_ENCODING_ASCII;
+    const auto enc = encoding ? ruby_encoding_to_onig_encoding(encoding) : fixed_encoding && !no_encoding ? ruby_encoding_to_onig_encoding(pattern->encoding())
+                                                                                                          : ONIG_ENCODING_ASCII;
```
seven1m, line 407: Holy nested ternary batman! 😆

seven1m (approved):

src/regexp_object.cpp
herwinw: Yeah, it was more readable than this:
```c++
const auto enc = encoding ?
    ruby_encoding_to_onig_encoding(encoding) :
    fixed_encoding && !no_encoding ? ruby_encoding_to_onig_encoding(pattern->encoding()) :
    ONIG_ENCODING_ASCII;
```
But clang-format disagreed with me on that one.

On the bright side: there are a few other options related to encoding, so we might have to refactor this soon.

PR #2465, opened 2025-01-07
Implement String#append_as_bytes

herwinw:


herwinw:
Kind of related to #217, it's a Ruby 3.4 method so it's not yet in the list.

I remember seeing this in the release notes and thinking there was a certain spot where I needed this. Of course I can't remember where that was (might have been something work-related)

PR #2466, opened 2025-01-07
Fix String#inspect for Shift_JIS encoding

herwinw:


seven1m:
Haha, I have this commit somewhere, but somehow I left it out of my Shift JIS branch. D'oh! Thanks for spotting that.

herwinw:
> Haha, I have this commit somewhere, but somehow I left it out of my Shift JIS branch. D'oh! Thanks for spotting that.

Now it's just missing `prev_char` for Shift_JIS and we're able to run all the tests in `test/natalie/string_test.rb`

PR #2470, opened 2025-01-07
Allow short names as FFI libraries

herwinw:
Turns out you don't have to write `"libm.#{SO_EXT}"`, but `"libm"` and `"m"` work just as fine.

I tried ffi-mysql again (see #2404 for context). This is enough to load the lib. Next step: `FFI::Struct`, which resembles a C struct and needs padding.

### DIFF
--- a/lib/ffi.cpp
+++ b/lib/ffi.cpp
@@ -12,14 +12,12 @@ Value init_ffi(Env *env, Value self) {
     return NilObject::the();
 }
 
-static void *dlopen_wrapper(Env *env, Value name) {
-    name->assert_type(env, Object::Type::String, "String");
-    const auto &str = name->as_string()->string();
-    void *handle = dlopen(str.c_str(), RTLD_LAZY);
+static void *dlopen_wrapper(Env *env, const String &name) {
+    void *handle = dlopen(name.c_str(), RTLD_LAZY);
     if (!handle) {
-        auto error = [&str]() {
+        auto error = [&name]() {
             // dlerror() clears the error message, so we need to set it again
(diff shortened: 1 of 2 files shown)


herwinw:
```
$ bin/natalie -Ivendor/bundle/ruby/3.4.0/gems/ffi-mysql-1.1.6/lib/ -rffi-mysql/mysql -e 'p Mysql.client_version'
80040
$ ruby -Ivendor/bundle/ruby/3.4.0/gems/ffi-mysql-1.1.6/lib/ -rffi-mysql/mysql -e 'p Mysql.client_version'
80040
```
I had to uncomment about half of that file, but at least we can use the FFI lib now.

herwinw:
> Making this change fixes the issue on my Mac.
> 
> ```diff
> -            if (name.length() > so_ext->length() && name[0] != '/' && (!name.ends_with(so_ext->string()) || name[name.length() - so_ext->length() - 1] != '.')) {
> +            if (name.length() > 0 && name[0] != '/' && !name.ends_with(so_ext->string())) {
> ```
> 
> In particular the `name.length() > so_ext->length()` bit was false on my machine, I guess because the extension is `dylib`?

That one was kind of flipped: if the length of the name is shorter than the file extension, it cannot contain the file extension, so we don't need to check for it. We still need to check for the length thanks to the extra check for a dot (for example `liblasso` ends with the file extension text, but not the extension).

herwinw:
Okay, I just refactored that code completely:
```c++
if (name.find('/') != 0 && !name.ends_with(so_ext)) {
```

PR #2473, opened 2025-01-09
Remove Ruby 3.1 as a supported host Ruby version

seven1m:


herwinw:
Could you update `test/natalie/data_test.rb` as well? That test has a workaround, since Data is only introduced in Ruby 3.2, so we could get rid of half of that file.

herwinw (approved):

PR #2476, opened 2025-01-09
Make fewer throwaway Envs for Blocks

seven1m:
There were a few places we were doing this:

```cpp
auto block_env = new Env {};
// ...
new Block { block_env, ... }
```

Since Block makes a copy of the passed Env anyway -- it has to because the parent scope can go away -- let's take a reference here to encourage call sites not to allocate a new Env on the heap, since it would be thrown away and collected by the GC on the next collection.

### DIFF
--- a/lib/ffi.cpp
+++ b/lib/ffi.cpp
@@ -329,12 +329,12 @@ Value FFI_Library_attach_function(Env *env, Value self, Args &&args, Block *) {
     if (status != FFI_OK)
         env->raise("LoadError", "There was an error preparing the FFI call data structure: {}", (int)status);
 
-    auto block_env = new Env {};
-    block_env->var_set("cif", 0, true, new VoidPObject { cif, [](auto p) { delete (ffi_cif *)p->void_ptr(); } });
-    block_env->var_set("arg_types", 1, true, arg_types_array);
-    block_env->var_set("return_type", 2, true, return_type);
-    block_env->var_set("ffi_args", 3, true, ffi_args_obj);
-    block_env->var_set("fn", 4, true, new VoidPObject { fn });
+    Env block_env;
+    block_env.var_set("cif", 0, true, new VoidPObject { cif, [](auto p) { delete (ffi_cif *)p->void_ptr(); } });
+    block_env.var_set("arg_types", 1, true, arg_types_array);
+    block_env.var_set("return_type", 2, true, return_type);
+    block_env.var_set("ffi_args", 3, true, ffi_args_obj);
+    block_env.var_set("fn", 4, true, new VoidPObject { fn });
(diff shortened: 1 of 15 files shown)


seven1m:
I could be convinced that we need a second Env constructor that takes a pointer and takes ownership of the passed pointer, but I think it will be confusing which constructor to use and future me will get it wrong.

lib/ffi.cpp
```diff
@@ -329,12 +329,12 @@ Value FFI_Library_attach_function(Env *env, Value self, Args &&args, Block *) {
...
-    block_env->var_set("return_type", 2, true, return_type);
-    block_env->var_set("ffi_args", 3, true, ffi_args_obj);
-    block_env->var_set("fn", 4, true, new VoidPObject { fn });
+    Env block_env;
+    block_env.var_set("cif", 0, true, new VoidPObject { cif, [](auto p) { delete (ffi_cif *)p->void_ptr(); } });
+    block_env.var_set("arg_types", 1, true, arg_types_array);
+    block_env.var_set("return_type", 2, true, return_type);
+    block_env.var_set("ffi_args", 3, true, ffi_args_obj);
+    block_env.var_set("fn", 4, true, new VoidPObject { fn });
     Block *block = new Block { block_env, self, FFI_Library_fn_call_block, 0 };
```
herwinw, line 338: This feels like a candidate for move semantics
```c++
     Block *block = new Block { std::move(block_env), self, FFI_Library_fn_call_block, 0 };
```
Otherwise we still create an additional env in the chain.
Of course, we would have to make the constructor with move semantics first

seven1m: I guess I need you to spell that out for me a bit... what would the constructor look like?

herwinw:
```c++
   // Current constructor (we need to keep this one)
    Block(Env &env, Value self, MethodFnPtr fn, int arity, BlockType type = BlockType::Proc)
        : m_fn { fn }
        , m_arity { arity }
        , m_env { new Env(env) }
        , m_self { self }
        , m_type { type } { }

    // Additional constructor
    Block(Env &&env, Value self, MethodFnPtr fn, int arity, BlockType type = BlockType::Proc) // Difference: &env -> &&env
        : m_fn { fn }
        , m_arity { arity }
        , m_env { new Env(std::move(env)) } // Difference: std::move() added
        , m_self { self }
        , m_type { type } { }
```
And then we'd also need a `Env(Env &&other)` constructor in the Env class.
But in hindsight: since we do need to allocate an Env on the heap, a unique pointer that can be moved would be better.

Maybe just ignore this for now, and I'll have a shot at it this weekend.

seven1m:
Merging this for now, but feel free to make it better @herwinw if you feel inspired. 😄

PR #2484, opened 2025-01-11
Add Block constructor with OwnedPtr<Env>

herwinw:
My take on #2476 

It's not exactly the way I expected it to be in that case. I was planning to use make `Block::m_env` a unique pointer, but since our OwnedPtr implementation does not allow access to the underlying raw pointer, this would require `Block::env()` to return a reference instead of a pointer, and there you'd get turtles all the way down.
Instead, the unique pointer is used in the constructor, where we immediately release it (that returns the raw pointer of the actual object) and save that, so it's only used as a temporary wrapper to signal ownership transfer.

My test script was as follows:
```ruby
# frozen_string_literal: true

GC.disable
to_proc = :to_s.to_proc
puts to_proc.call(1)
```
The GC has been disabled to make the results stable.
On release build, this saw the results of callgrind go down from 20,297,022 to 20,295,904, so it's not much, but it's something.

I did experiment with a `Env(Env &&other)` constructor, but that resulted in more code and less speed gain.

### DIFF
--- a/include/natalie/block.hpp
+++ b/include/natalie/block.hpp
@@ -4,6 +4,7 @@
 #include "natalie/env.hpp"
 #include "natalie/forward.hpp"
 #include "natalie/gc.hpp"
+#include "tm/owned_ptr.hpp"
 
 namespace Natalie {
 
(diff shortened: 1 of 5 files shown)


herwinw:
And I just want to add: the other case was really helpful as a quick reference to find the constructors that could be converted to move semantics, so even if it looks like I threw away most of that code, it saved me a lot of time.

seven1m:
Looks good. I like it.

I don't know why the gc linter is confused; I can look into that later.

herwinw:
I think I've got the GC script fixed now. It gets completely confused when it encounters `OwnedPtr<Env>`, which looks like a bug in ffi-clang (or more likely: clang itself).
Using `TM::OwnedPtr<Env>` resolves the issue :shrug:

seven1m (approved):

Issue #2095, opened 2024-06-14
natalie-lang.org server.rb doesn't work with latest Natalie


seven1m:
Starting in 2ef5771069b8de5f145952612648e0ff27d0bc99 the payload returned from https://natalie-lang.org/stats/stats.json is truncated. Need to investigate...

seven1m:
Also, 9 out of 10 of the child processes see the following error for every socket connection:

```
["Errno::EAGAIN", "Resource temporarily unavailable", ["server.rb:41:in `accept'", "server.rb:41:in `block in block in block in block'", "server.rb:40:in `loop'", "server.rb:40:in `block in block in block'", "server.rb:39:in `fork'", "server.rb:39:in `block in block'", "src/enumerable.rb:624:in `block in collect'", "src/enumerable.rb:624:in `each'", "src/enumerable.rb:624:in `collect'", "server.rb:38:in `<main>'"]]["Errno::EAGAIN", "Resource temporarily unavailable", ["server.rb:41:in `accept'", "server.rb:41:in `block in block in block in block'", "server.rb:40:in `loop'", "server.rb:40:in `block in block in block'", "server.rb:39:in `fork'", "server.rb:39:in `block in block'", "src/enumerable.rb:624:in `block in collect'", "src/enumerable.rb:624:in `each'", "src/enumerable.rb:624:in `collect'", "server.rb:38:in `<main>'"]]
```

It seems that `server.accept` returns one good connection and raises an error for all the other children waiting.

funny-falcon:
It is well known issue with non-blocking accept from many threads/processes: kernel wakes all `epool/poll/select` waiters of socket, but only one succeed.

Nginx works around by using cross-process lock around waiting for accept. Some other servers use blocking listen socket and do accept in separate threads (iirc, Rainbow did it).

seven1m:
Interestingly, Firefox and `curl` were both getting the whole response, but Chrome was not. I suspect it's in how the client reads the data off the socket, causing some combination of a partial write and/or EAGAIN, which we are not handling as we should. Fixed in #2488.

PR #2486, opened 2025-01-11
Handle EAGAIN in Server#accept

seven1m:
#2095

This seems to be what CRuby does, and eliminates the barf of EAGAIN errors from 9 of the 10 children in natalie-lang.org's [server.rb](https://github.com/natalie-lang/natalie-lang.org/blob/master/server.rb).

### DIFF
--- a/lib/socket.cpp
+++ b/lib/socket.cpp
@@ -109,9 +109,17 @@ static Value Server_sysaccept(Env *env, Value self, sockaddr_storage &addr, sock
 
     int fd;
     if (is_blocking) {
-        fd = blocking_accept(env, self->as_io(), reinterpret_cast<sockaddr *>(&addr), &len);
-        if (fd == -1)
-            env->raise_errno();
+        bool retry = false;
+        do {
+            retry = false;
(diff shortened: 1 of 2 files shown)


lib/socket.cpp
```diff
@@ -109,9 +109,17 @@ static Value Server_sysaccept(Env *env, Value self, sockaddr_storage &addr, sock
...
+        do {
+            retry = false;
+            fd = blocking_accept(env, self->as_io(), reinterpret_cast<sockaddr *>(&addr), &len);
+            if (fd == -1) {
+                if (errno == EAGAIN)
+                    retry = true;
+                else
+                    env->raise_errno();
+            }
+        } while (retry);
```
funny-falcon, line 122: Why not
```cpp
for (;;) {
    fd = blocking_accept(env, self->as_io(), reinterpret_cast<sockaddr *>(&addr), &len);
    if (fd != -1)
        break;
    if (errno == EAGAIN)
        continue;
    env->raise_errno();
}
```

PR #2487, opened 2025-01-11
Utilize TM::Vector::concat

herwinw:
The sequel to https://github.com/seven1m/tm/pull/18

`array/plus_spec` went down from 510,627,871 to 510,521,178 operations, `array_concat_spec` from 509,942,054 to 509,827,801 (all callgrind).


herwinw:
And a slight decrease in the operations in boardslam as well (around 0.01%, but still)

seven1m (approved):
💥

PR #2490, opened 2025-01-11
Optimizations in Array#[]=

herwinw:
Use more concat and move instructions.


herwinw:
This looks like there are more use cases for a `Vector::concat_slice` than I originally thought. I might try to add that to the TM lib tomorrow to remove more loops in this code.

seven1m (approved):

PR #2496, opened 2025-01-12
Fix be_an_instance_of behaviour in specs

herwinw:
This one was aliased to be_kind_of, but in this case we do not want subclasses to match.


seven1m (approved):

seven1m:
Haha this looks like a can of worms you opened 🫣

herwinw:
> Haha this looks like a can of worms you opened 🫣

Yes, this is going to take a while before I've got everything fixed.
The good news is that it's mostly just removing a line of code per fix.

herwinw:
I'm surprised by the amount of unrelated bugs I'm fixing now. Take for instance this code:
```ruby
idx = +"el"
x = "hello"[idx]
x << "0"
p idx
```
Natalie did not make a copy of the return value of `"hello"[idx]`, so the output of MRI was "el", but Natalie printed "el0"

seven1m:
I'm on mobile and cannot look... should we change the copy constructors for various classes (String, Array, etc.) to *not* copy the class? Or is there still use for the copy constructors as-is?

herwinw:
> I'm on mobile and cannot look... should we change the copy constructors for various classes (String, Array, etc.) to _not_ copy the class? Or is there still use for the copy constructors as-is?

Not sure, I think they're still in use. We could probably just try it with the constructors removed (or altered to not copy the class object) and see what breaks.

seven1m (approved):

PR #2498, opened 2025-01-12
Be instance of spec alt

herwinw:
Attempt at an alternate solution for  #2496. This MR is just to see what the CI finds.
Not all failures use the updated code path, so a few issues will fail.

### DIFF
--- a/include/natalie/array_object.hpp
+++ b/include/natalie/array_object.hpp
@@ -31,7 +31,7 @@ public:
     ArrayObject(std::initializer_list<Value> list);
 
     ArrayObject(const ArrayObject &other)
-        : Object { other }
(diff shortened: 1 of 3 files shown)


herwinw:
Looks like this breaks the behaviour of clone (and probably dup too)
```
Array#clone
  returns an Array or a subclass instance
    [1, 2] (Array) should be an instance of ArraySpecs::MyArray
    (/natalie/spec/core/array/shared/clone.rb:4:in 'block in block in block')
4 spec(s) passed.
1 spec(s) failed.
0 spec(s) errored.
 FAIL ruby/spec::spec/core/array/clone_spec.rb#test_0001_passes all specs (560.13s)
```

seven1m:
OK thanks for trying

PR #2507, opened 2025-01-16
Cleanups for regexp encoding

herwinw:
This makes it easier to add other encodings.


herwinw:
This started as an attempt to add binary encoding (the `/n` flag), but that one needs some more work

PR #2508, opened 2025-01-17
Import mspec tests to run our spec code

herwinw:
After the recent fixes in  #2496, where we were using the wrong method to run a certain spec, I noticed that the upstream mspec repo contained specs for mspec. The code uses rspec to run the mspec specs (yes, this sentence reads terrible).

This is a very early attempt to import these specs and run our spec runner through it. For now, it just runs a single test, and it needs a wrapper that changes like 90% of the test itself to make it compatible.

I will try some more things, but this might just end up with me abandoning it.


herwinw:
I guess the intention of this change has shifted a bit when I was working on it. Since the internal interface of our specs runner is incompatible with the upstream one, just trying to import everything is a bad idea. Instead, I just started with importing the specs for a single matcher, and refactored the code until we could use the upstream matcher. Then it turned into a change to import the upstream mspec code.
The end goal is to slowly replace our spec runner with the upstream mspec code.

herwinw:
The specs for the matrix gem include a custom matcher. I updated that one in the checkout to match our spec runner, but the nightly tests use a clean specs checkout and those still fail. So I guess I've found some motivation to finish this.

PR #2519, opened 2025-01-23
Move is_integer() to Value

seven1m:
This PR started out by moving `Object::is_integer()` to `Value::is_integer()` as we want to stop hydrating integers. Following from that was a bunch of work to make certain member functions into static functions that can accept a `Value` instead of needing a full `Object`.

This will continue in the next few PRs until I can fully get rid of the `Object::Type::Integer` enum value, as all integers will be only Values.


src/value.cpp
```diff
@@ -189,14 +189,59 @@ Integer &Value::integer() {
...
+            if (val >= (LONG_MIN >> 1) && val <= (LONG_MAX >> 1))
+                return left_shift_with_undefined_behavior(val, 1) | 1;
+        } else {
+            return reinterpret_cast<nat_int_t>(i.bigint_pointer());
+        }
+    }
+
+    if (m_type == Type::Double) {
+        // FIXME: This is scuffed, but gets the tests passing.
+        return 1;
```
seven1m, line 236: I'm ashamed of this, but I plan to fix this up Real Soon Now™.

seven1m:
The spec failure was a timeout in a `threadgroup` spec. I'll increase the timeout for that in the next PR.

PR #2521, opened 2025-01-23
Remove Integer allocations in comparisons in Time

herwinw:
This is just some low hanging fruit. All the other int allocations in this file look to be related to Rational, which will need an overhaul before we can get rid of those.


seven1m:
I so happen to be working on Rational now! 🎉

PR #2524, opened 2025-01-24
Add ObjectType::Collected

herwinw:
The old version marked garbage collected objects with Nil, which can be very confusing when trying to track down a bug.

For a bit of context: I've been working on [define_finalizer](https://ruby-doc.org/3.4.1/ObjectSpace.html#method-c-define_finalizer) and ran into a bug where one of the objects got GC'd, but that took way too long to figure out because inspecting the structure with GDB showed `Nil`.

### DIFF
--- a/include/natalie/object.hpp
+++ b/include/natalie/object.hpp
@@ -92,7 +92,7 @@ public:
     }
 
     virtual ~Object() override {
-        m_type = ObjectType::Nil;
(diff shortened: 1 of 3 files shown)


seven1m (approved):
What a great idea! Cannot tell you how much time I have lost investigating collected objects. 😅

herwinw:
Good to hear that I'm not the only one ;)

PR #2548, opened 2025-01-31
Use the capacity keyword in String.new

herwinw:
This includes the addition of  a function of `TM::String` to grow the capacity manually, which may be useful in other locations as well.

### DIFF
--- a/ext/tm/include/tm/string.hpp
+++ b/ext/tm/include/tm/string.hpp
@@ -1453,6 +1453,29 @@ public:
     }
 
     /**
+     * Grow the capacity (allocated memory) of the string.
+     *
+     * ```
+     * auto str = String { "abc" };
+     * str.set_capacity(100);
+     * assert_eq(100, str.capacity());
+     * assert_eq(3, str.size());
+     * ```
+     *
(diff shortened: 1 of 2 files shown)


ext/tm/include/tm/string.hpp
```diff
@@ -1452,6 +1452,29 @@ class String final {
...
+     *
+     * ```
+     * auto str = String { "abc" };
+     * str.set_capacity(1);
+     * assert_eq(3, str.capacity());
+     * assert_eq(3, str.size());
+     * ```
+     */
+    void set_capacity(size_t new_size) {
+        grow_at_least(new_size);
```
seven1m, line 1475: What if we rename `grow_at_least` to `set_capacity`? I don't think we need to maintain any backward compatibility since no one else is using the library. 😉

seven1m (approved):

ext/tm/include/tm/string.hpp
seven1m: Sorry I didn't see the commit message. Sounds good for now. Carry on!

PR #2555, opened 2025-02-01
Convert internal encoding list from hashmap to vector

herwinw:
The hash keys were a simple list of sequential integers starting at 1. The current Hashmap does not handle integer keys very well, it defaults to pointer lookup, where the return value for non pointers is a hardcoded 0. Effectively, this hashmap was a linked list.


seven1m (approved):

herwinw:
I am trying to rework the hashmap internals to fix these kind of things, but in this case a vector is better even if the hashmap supports integer values.

PR #2579, opened 2025-02-06
Fix timeout in spec for integer left shift

herwinw:
And right is halfway there too, but I'm going to call it a day.

### DIFF
--- a/src/integer_object.cpp
+++ b/src/integer_object.cpp
@@ -458,6 +458,8 @@ Value IntegerObject::bitwise_xor(Env *env, Integer &self, Value arg) {
 }
 
 Value IntegerObject::left_shift(Env *env, Integer &self, Value arg) {
+    if (is_zero(self))
+        return Value::integer(0);
     auto integer = arg.to_int(env);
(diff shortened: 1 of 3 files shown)


seven1m (approved):

src/integer_object.cpp
```diff
@@ -458,23 +458,32 @@ Value IntegerObject::bitwise_xor(Env *env, Integer &self, Value arg) {
 }
 
 Value IntegerObject::left_shift(Env *env, Integer &self, Value arg) {
+    if (is_zero(self))
+        return Value::integer(0);
     auto integer = arg.to_int(env);
     if (is_bignum(integer)) {
-        if (IntegerObject::is_negative(self))
+        if (IntegerObject::is_negative(self) && IntegerObject::is_negative(integer))
```
seven1m, line 465: It doesn't matter in the slightest, but in the future, I hope to convert all the `IntegerObject::is_negative(integer)` type calls to the equivalent and easier-on-the-eyes `integer.is_negative()` where possible.

PR #2583, opened 2025-02-07
TEMP: Freeze result of Symbol#to_s

herwinw:
Just going to use the CI for a bit to find issues like #2582 by making it an error instead of a warning that gets ignored in the CI. This change should **NOT** get merged, but it is an easy way to find the issues that can be fixed.


herwinw:
It's just the specs for chilled strings that fail, and these should fail since we've altered the behaviour that is tested in that spec. This means we don't need any additional fixes.

I really should have done this earlier.

PR #2588, opened 2025-02-09
Implement parser for Kernel.Complex with String argument

herwinw:
This one is now feature complete with `String#to_c`, which is helpful with all the shared specs that should behave the same.

This contains a lot of code duplication, the next step is probably to add an extra argument to specify the failure behaviour (raise an exception or return the parsed part) and delegate `String#to_c` to `Kernel.Complex`.


seven1m (approved):
Wow I had no idea this was a parser lol

PR #2589, opened 2025-02-09
Use libjson-c for JSON handling

herwinw:
After looking into the timeout in https://github.com/natalie-lang/natalie/blob/9c36783dd14ed7c2f9d16087bfe2efd077b51da9/spec/security/cve_2013_4164_spec.rb#L14: It looks like our JSON parser is very inefficient. With the following diff:
```diff
--- a/lib/json.rb
+++ b/lib/json.rb
@@ -149,6 +149,7 @@ module JSON
           num << c
           advance
           while DIGITS.include?(c = current_char)
+            $stderr.puts("#{Process.clock_gettime(Process::CLOCK_MONOTONIC_RAW)}: new digit #{c}, current size = #{num.size}")
             num << c
             advance
           end
```
My output looks like this:
```
17471.477619132: new digit 1, current size = 2
17471.888347512: new digit 1, current size = 3
17472.289083811: new digit 1, current size = 4
17472.6895373: new digit 1, current size = 5
17473.088253879: new digit 1, current size = 6
17473.495533448: new digit 1, current size = 7
```
So parsing a single numeric digit takes about 0.4-0.5 seconds. This case is just to try to see if it will be faster when delegated to libjson-c


seven1m:
We don't have any optimizations around indexing a string, so I imagine this code is a big part of the slowness:

https://github.com/natalie-lang/natalie/blob/master/lib/json.rb#L234-L240

herwinw:
So, I wrote a dumb script to test it with:
```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require 'json'

input = [
  :a,
  'b',
  12345,
  123.45,
  'a long string, because we need to test with lots of data. Have you heard the story about the goat and the cow who met in a street? No? Neither have I',
  {
    a: 1,
    b: 'Some object',
    'with keys' => 'And another long string because why not? Well, I still do not know what happened to the goat and the cow',
  },
]
p JSON.generate(input)
```
Callgrind results:
* Master: 32,267,739
* This branch: 19,413,757

In both cases, just running `bin/natalie -rjson -e ''` is already a low 19M instructions, so it looks like this is roughly a 100x speedup :astonished:

Rakefile
```diff
@@ -498,6 +499,17 @@ file 'build/zlib/libz.a' do
...
+  build_dir = File.expand_path('build/json-c', __dir__)
+  rm_rf build_dir
+  cp_r 'ext/json-c', build_dir
+  sh <<~SH
+    cd #{build_dir} && \
+    cmake . && \
+    make -j 4
+  SH
+end
+
```
seven1m, line 512: Is it possible to build without cmake? I'd rather not add another dependency for building Natalie if we don't have to.

herwinw: I thought we already had that dependency, but the CI showed me otherwise.

The cmake command generates a Makefile, so I'm just going to strip that one and include it as an additional file. If that doesn't work, I might just replace this whole implementation with something like nlohmann (or whatever the name was), a C++ header only JSON library.

seven1m: cool cool

herwinw:
I now works without any additional dependencies.

This is only half of the change, the JSON parsing is still in Ruby. This is left as an ~exercise to the reader~ something for tomorrow

seven1m (approved):

PR #2592, opened 2025-02-10
Add tests for JSON parse of big int

herwinw:
We need to be sure any replacement parser is able to handle values that exceed regular integer bounds.

### DIFF
--- a/test/natalie/json_test.rb
+++ b/test/natalie/json_test.rb
@@ -17,6 +17,8 @@ describe 'JSON' do
       JSON.parse('-3000').should == -3000
       JSON.parse('0').should == 0
       JSON.parse('-0').should == 0
+      JSON.parse((2**64).to_s).should == 2 **64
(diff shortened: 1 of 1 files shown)


herwinw:
I added these tests as additional tests for rewriting JSON.parse with libjson-c. It turns out the parsing of this lib is pretty annoying. By default the parser allows things like `[1,]`, even though they're not valid JSON (because standards are just a suggestion). There is a flag to enable strict parsing mode, but enabling that flag also limits integers to 64 bits and throws a parsing error if it encounters an int that's too big. It does not look like there is a way to enable only half of this behaviour without patching the source code.
It did fix the timeout in the security spec we were hitting though.

There are a few C++ specific JSON parsers (nlohmann, simdjson), but these all look like they're integrated with the STL, so that adds some compilation and conversion overhead.

seven1m (approved):

PR #2602, opened 2025-02-13
Fix parsing of + sign in Rational coercion

herwinw:


seven1m (approved):

herwinw:
And https://github.com/ruby/spec/pull/1244 for the upstream spec

PR #2606, opened 2025-02-15
Turn TM::Hashmap inside out

herwinw:
The old implementation had two major disadvantages:
* The key type and the hash/compare methods could be defined individually. This resulted in issues with integer keys and pointer type, where the hash was always 0 and the hash map was just a linked list (see #2559)
* The lookup methods required a copy of the object to look up. This is okay with a pointer, but inefficient with a string. As a side effect, it was impossible to do a lookup with a stringview or a char pointer, even though this included the exact same information as the string object.

This change extracts the hash and lookup methods into separate typed structs. A hash for an integer type will now fail to compile, until you implement a specific typed struct for integers. The hash/compare methods can now be defined polymorphically, and arguments to the hash methods are being forwarded, which means things like const references to strings or string views now work without the need to create a copy.

This mostly helps the Symbol lookup, which is a Hashmap with TM::String keys.

### DIFF
--- a/ext/tm/include/tm/string_view.hpp
+++ b/ext/tm/include/tm/string_view.hpp
@@ -252,6 +252,20 @@ public:
      * ```
      * auto str = String("foo-bar-baz");
      * auto view = StringView(&str, 4, 3);
+     * String str2 = view;
+     * assert_str_eq("bar", str2);
+     * ```
+     */
+    operator String() const {
+        return to_string();
+    }
+
+    /**
+     * Returns a new String constructed from this view.
+     *
+     * ```
+     * auto str = String("foo-bar-baz");
+     * auto view = StringView(&str, 4, 3);
      * auto str2 = view.clone();
      * assert_str_eq("bar", str2);
      * ```
--- a/src/hash_object.cpp
+++ b/src/hash_object.cpp
@@ -5,7 +5,6 @@
 
 namespace TM {
 
-// this is used by the hashmap library and assumes that obj->env has been set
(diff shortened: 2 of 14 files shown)


ext/tm/include/tm/string_view.hpp
```diff
@@ -246,6 +246,20 @@ class StringView {
...
+     * Returns a new String constructed from this view.
+     *
+     * ```
+     * auto str = String("foo-bar-baz");
+     * auto view = StringView(&str, 4, 3);
+     * String str2 = view;
+     * assert_str_eq("bar", str2);
+     * ```
+     */
+    operator String() const {
```
seven1m, line 259: TIL

seven1m (approved):
That's some mad lad template programming, i.e. my coping mechanism for I could not have programmed this. Well done, friend.

src/hash_object.cpp
```diff
@@ -3,23 +3,27 @@
 #include "tm/recursion_guard.hpp"
 #include "tm/vector.hpp"
 
-namespace Natalie {
+namespace TM {
 
 // this is used by the hashmap library and assumes that obj->env has been set
-size_t HashObject::hash(Key *&key) {
+size_t HashKeyHandler<Natalie::HashKey *>::hash(Natalie::HashKey *key) {
```
seven1m, line 9: I think the comment above this line is inaccurate (and was already inaccurate before this change). You can just remove the comment, if you want, because I think the code is probably clear enough now.

PR #2618, opened 2025-02-23
Refactor String#to_f

herwinw:
This introduces a NumberParser class with a generic framework for parsing numbers, which can be used in all other number coercion methods as well.
Next step is to use this parser in Kernel.Float as well, but it turns out this one behaves slightly different: "0x10" is a valid float according to Kernel.Float (with the value 16.0), but String#to_f does not accept the hex prefix. Ruby, I want to love you, but why? So the next step is probably going to rewrite half of it.


seven1m (approved):
It's so beautiful! 😍

PR #2630, opened 2025-03-01
Implement String#squeeze and String#squeeze!

whtsht:
Related issue #217

### DIFF
--- a/src/string_object.cpp
+++ b/src/string_object.cpp
@@ -3038,6 +3038,57 @@ Value StringObject::split(Env *env, Optional<Value> splitter_arg, Optional<Value
     return ary;
 }
 
+Value StringObject::squeeze(Env *env, Args &&selectors) {
+    auto copy = new StringObject { m_string, m_encoding };
+    copy->squeeze_in_place(env, std::move(selectors));
+    return copy;
+}
+
+Value StringObject::squeeze_in_place(Env *env, Args &&selectors) {
+    assert_not_frozen(env);
+    if (selectors.size() == 0)
+        return squeeze_in_place_without_selectors(env);
+
+    const auto old_len = bytesize();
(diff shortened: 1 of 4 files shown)


src/string_object.cpp
```diff
@@ -3038,6 +3038,57 @@ Value StringObject::split(Env *env, Optional<Value> splitter_arg, Optional<Value
...
+
+    const auto old_len = bytesize();
+    auto handler = character_class_handler(env, std::move(selectors));
+    size_t index = 0;
+    auto last_character = next_char(&index);
+    while (index < m_string.size()) {
+        const auto old_index = index;
+        auto character = next_char(&index);
+
+        if (last_character.to_string() == character.to_string() && handler(character)) {
```
seven1m, line 3060: Can we compare the two StringViews here instead of building a String? (I'm on mobile so I cannot see these types very well.)

whtsht: It seems that StringView comparison (`last_character == character`) doesn't work due to the difference in offset.

https://github.com/natalie-lang/natalie/blob/1819b1c6404276bc03a5debdda36543ae1468811/ext/tm/include/tm/string_view.hpp#L221-L222

I'm not sure if I should fix this since I don't know much about the background.

seven1m: Oh yeah, that looks like a bug to me. @herwin?

herwinw: I'll have a look at it tomorrow, so for now, I think we can merge this with the string conversions, and I can remove them again with the fix in string_view.hpp.

Also: nice find, I would have never looked at this library code if I was just using the views.

seven1m: LOL oops that's what I get for doing github on my phone. 😅

herwinw: For completeness, #2634 includes this fixes

PR #2634, opened 2025-03-02
Fixes to StringView::operator==

herwinw:
Beside the issue found in #2630, there were two other bugs in the code, so I'm surprised we never had other issues with this operation.
Beside those bugs, the order of operation has been shuffled a bit, to make the operation more efficient and remove some duplicate checks.


seven1m (approved):
Nice!

PR #2644, opened 2025-03-05
Fix String#gsub: respects unicode when the pattern collapses

whtsht:


seven1m (approved):

src/string_object.cpp
```diff
@@ -2581,11 +2581,16 @@ Value StringObject::gsub(Env *env, Value find, Optional<Value> replacement_value
         if (match) {
             byte_index = match->end_byte_index(0);
             if (match->is_empty()) {
+                if (byte_index < m_string.size())
+                    out.append(peek_char(byte_index).to_string().c_str());
```
herwinw, line 2585:
```c++
out += peek_char(byte_index)
```
The should work as well, and removes the conversion to String.

whtsht: Thanks for the feedback! I fixed it.

PR #2645, opened 2025-03-05
Fix execution order for ConstantPathWriteNode

herwinw:
```ruby
Foo::Bar = (raise "frop"; 1)
```

Code like this should evaluate the LHS first, so this will fail because `Foo` cannot be found, and never raise

### DIFF
--- a/lib/natalie/compiler/pass1.rb
+++ b/lib/natalie/compiler/pass1.rb
@@ -1140,13 +1140,15 @@ module Natalie
       end
 
       def transform_constant_path_write_node(node, used:)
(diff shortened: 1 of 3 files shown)


seven1m (approved):

herwinw:
Instructions for `Foo::Bar = 1` now look like this:
```
0 push_self
1 const_find Foo
2 push_int 1
3 dup
4 move_rel 2
5 const_set Bar
```
There are tons of similar issues in the specs that all needs fixing by the mov_rel instruction to shuffle the stack. I'm starting to think it would be better to update the `const_set` instruction to expect the values in the flipped order.

PR #2646, opened 2025-03-06
Remove NATFIXME blocks for Hash argument in String#sub

herwinw:
They are passing, so I'm not sure why the block didn't complain about a lack of expected exception raises.


herwinw:
I guess it's similar to #2621, the global variables and block usage can break.

I noticed this one because I looked at `String#sub` to see if it had the same issues that were fixed in #2644, and noticed my local run had more skips than the specs page had failures.

seven1m (approved):

PR #2647, opened 2025-03-06
Improve memory scanning performance a bit

seven1m:
This shows small improvement when running a program with lots of GC runs, including the self-hosted compiler.

```
→ hyperfine --warmup 10 'bin/nat_master -c h1 examples/hello.rb' 'bin/nat -c h2 examples/hello.rb'

Benchmark 1: bin/nat_master -c h1 examples/hello.rb
  Time (mean ± σ):     927.4 ms ±  32.2 ms    [User: 720.6 ms, System: 156.1 ms]
  Range (min … max):   881.7 ms … 991.4 ms    10 runs

Benchmark 2: bin/nat -c h2 examples/hello.rb
  Time (mean ± σ):     878.5 ms ±   5.2 ms    [User: 723.1 ms, System: 154.5 ms]
  Range (min … max):   870.5 ms … 884.7 ms    10 runs

Summary
  bin/nat -c h2 examples/hello.rb ran
    1.06 ± 0.04 times faster than bin/nat_master -c h1 examples/hello.rb
```

### DIFF
--- a/src/gc.cpp
+++ b/src/gc.cpp
@@ -260,13 +260,13 @@ NO_SANITIZE_ADDRESS void Heap::scan_memory(Cell::Visitor &visitor, void *start, 
         Cell *potential_cell = *reinterpret_cast<Cell **>(ptr); // NOLINT
         if (!potential_cell)
             continue;
+        fn(potential_cell); // this must be here for ASan to check if it's on the fake stack
         if ((reinterpret_cast<uintptr_t>(potential_cell) & 0b111) != 0b000)
             continue;
         if (reinterpret_cast<uintptr_t>(potential_cell) < m_lowest_pointer_address || reinterpret_cast<uintptr_t>(potential_cell) > m_highest_pointer_address)
             continue;
         if (is_a_heap_cell_in_use(potential_cell))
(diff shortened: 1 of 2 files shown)


src/gc.cpp
```diff
@@ -243,6 +243,8 @@ NO_SANITIZE_ADDRESS void Heap::scan_memory(Cell::Visitor &visitor, void *start,
         Cell *potential_cell = *reinterpret_cast<Cell **>(ptr); // NOLINT
         if (!potential_cell)
             continue;
+        if ((reinterpret_cast<uintptr_t>(potential_cell) & 0b11) != 0b00)
```
herwinw, line 249: We're 8 bit aligned (or rather: the OS is 8 bit aligned), so I would guess we could change this to `& 0b111`.

seven1m: Yeah I had that before and the `asan` test failed with an unexpected object collection, so I'm trying this to see what's up. Just experimenting. :D

seven1m: ...and it's still failing, so I guess there's some other issue. Probably the range check. 🤔

PR #2655, opened 2025-03-17
Update BeEmptyExpectation

herwinw:
Use the `#empty?` method instead of checking on length.


herwinw:
I found this while trying to run the matrix gem through natalie, and this was the first breaking spec.

PR #2656, opened 2025-03-18
Import Matrix gem

herwinw:
This includes a fix for the call to `initialize_copy` when calling `Kernel#dup`, which may make some other code simpler. Two failing tests have been disabled, this is probably an issue with our enumerator, and should be fixed in other code.

### DIFF
--- a/src/kernel_module.cpp
+++ b/src/kernel_module.cpp
@@ -926,8 +926,11 @@ Value KernelModule::dup(Env *env, Value self) {
     case Object::Type::Hash:
     case Object::Type::String:
         return dup_better(env, self);
-    default:
-        return self->duplicate(env);
+    default: {
+        auto result = self->duplicate(env);
+        result.send(env, "initialize_copy"_s, { self });
+        return result;
+    }
     }
 }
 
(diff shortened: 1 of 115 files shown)


src/kernel_module.cpp
```diff
@@ -926,8 +926,11 @@ Value KernelModule::dup(Env *env, Value self) {
     case Object::Type::Hash:
     case Object::Type::String:
         return dup_better(env, self);
-    default:
-        return self->duplicate(env);
+    default: {
+        auto result = self->duplicate(env);
+        result.send(env, "initialize_copy"_s, { self });
+        return result;
+    }
```
seven1m, line 933: We might be able to call `dup_better()` for plain Objects, which calls `initialize_dup`, which in turn calls `initialize_copy`. No biggie for this PR. I can investigate further in another PR.

```ruby
class Foo
  def initialize_copy(*)
    puts caller
    super
  end
end

f = Foo.new
f.dup

# wat.rb:9:in `initialize_dup'
# wat.rb:9:in `dup'
# wat.rb:9:in `<main>'
```

PR #2657, opened 2025-03-19
Add lock for build directory

seven1m:
...so multiple processes can utilize the same build directory without stepping on each other's work.

This also adds the `--build-quietly` option and enables the build directory for our test runs.


seven1m:
Full CI time is now down from 40 minutes to about 25 minutes, so that's nice.

PR #2672, opened 2025-04-03
Update specs after mass update upstream

herwinw:
This excludes the changes for stringscanner, this library has a lot of changes that are version related and are just too much to update now, this can be treated as a separate project.


seven1m (approved):

herwinw:
The change was based on an older checkout of natalie/master, it looks like a rebase did the trick.

PR #2675, opened 2025-04-07
Add Version constant to StringScanner

herwinw:
The value is defined as the maximal value in the current upstream specs (Ruby 3.4.3).
This should resolve the crashes in the nightly spec runner and turn them into actual failed specs which we can fix.


herwinw:
I merged this one without waiting for the CI to time out again: that failure is unrelated to this change, and I would prefer to have this one included in the next nightly CI run

PR #2688, opened 2025-04-11
Implement StringScanner#charpos

herwinw:
By aliasing it to #pos. This smells fishy, either it is an alias and it should use the shared spec for #pos, or it behaves diffently but the specs don't cover those differences.


seven1m (approved):

herwinw:
https://github.com/ruby/spec/issues/1256

Upstream case to fix these specs, it looks like our implementation is incorrect, but the current specs don't trigger the errors.

PR #2696, opened 2025-04-19
Make StringScanner#skip_until spec compliant

herwinw:


herwinw:
And that was the last of the "convert string to regex" low hanging fruit of the stringscanner specs.

seven1m (approved):

PR #2699, opened 2025-04-23
Avoid creating an array from Args every time

seven1m:
Doing `Args::pop()` and `Args::shift()` allows us to avoid creating an ArrayObject for many cases.


herwinw:
I see you found the issue that I wanted to tackle today.

seven1m:
Yeah I have ideas for making keyword args not require a real HashObject, but I needed to change how we work with the args array first.

seven1m:
I have another branch where I'm revisiting this, but I need to fix a GC bug first. Closing this for now...

PR #2703, opened 2025-04-27
fix compile issue on x86 MacOS

Stevo-S:
Fetch the thread context RSP register differently on x86_64 Apple devices.
This fixes a build issue where compilation fails with a

```
error: member reference type 'struct __darwin_mcontext64 *' is a pointer; did you mean to use '->'?
```

On such platforms, `ctx->uc_mcontext` is a pointer itself.
Thus the members of the struct being pointed to need to be accessed via the arrow operator.
Also, `gregs` is not a member of the struct so the registers are accessed via the `__ss` member.

### DIFF
--- a/src/natalie.cpp
+++ b/src/natalie.cpp
@@ -960,7 +960,7 @@ void gc_signal_handler(int signal, siginfo_t *, void *ucontext) {
         memcpy(ctx, ucontext, sizeof(ucontext_t));
         thread->set_suspend_status(ThreadObject::SuspendStatus::Suspended);
 
-#if defined(__x86_64__)
+#if defined(__x86_64__) && !defined(__APPLE__)
         thread->set_end_of_stack(reinterpret_cast<void *>(ctx->uc_mcontext.gregs[REG_RSP]));
 #endif
 


seven1m (commented):
Thanks for taking a stab at this! Actually we don't use the `gc_signal_handler` on Apple, so I think we could change:

```
#if defined(__x86_64__)
```

...to:

```
#if defined(__x86_64__) && !defined(__APPLE__)
```

Can you give that a try on your machine?

Stevo-S:
Thanks for taking a look at this.
(I love what you're doing with the Natalie project, by the way!)

Let me make the change and give it a try on my machine.

Stevo-S:
Made the change. 
I'm able to get it to build successfully with the suggested change.
Safe to say it works!

seven1m:
Perfect! Let's rebase this into a single commit and I'll be happy to merge. Thanks.

seven1m (approved):

Stevo-S:
> Perfect! Let's rebase this into a single commit and I'll be happy to merge.

Done.

PR #2708, opened 2025-05-01
Import specs for Time#strftime

herwinw:
Since part of the specs are shared with the Date class, we might have to fix them simultaneously.


herwinw:
I was looking at fixing the date output of YAML, but I got sidetracked with the formatting functions (read: I had to copy-paste half of the implementation of `TimeObject::inspect` to get it to work).

By the looks of it, Ruby supports a bunch of additional modifiers that we need to implement manually instead of forwarding things to `strftime(3)`.

seven1m (approved):

PR #2709, opened 2025-05-01
Rename @enums ivar to @ffi_enums

herwinw:
This keeps the name consistent with mri-ffi.


herwinw:
I'm trying to get the internals more similar to the upstream ffi gem, this makes extending things a bit easier. This is just a simple rename of some internal values, which should not have any influence on the calling code.

seven1m (approved):

PR #2711, opened 2025-05-04
Support Enum class in FFI

herwinw:


seven1m (approved):
Cool!

herwinw:
Now it's just the FFI::Struct and we should be able to run ffi-mysql

PR #2716, opened 2025-05-07
Refresh specs after upstream updates

herwinw:
And include a tiny error message fix for read-only globals, this was easier than adding NATFIXME blocks around all those failures.


seven1m (approved):
That failing test is a GC bug in the self-hosted compiler. I'm now on week 3 of looking into it. 🤪 You can ignore it.

herwinw:
I had pressed retry before I read that :)

PR #2726, opened 2025-05-11
Move ClassObject* argument before the method call arguments

herwinw:
That was far less of a chore than I expected it to be


herwinw:
Accidentally found an issue in the spec of `Time#-`, I added an upstream MR on https://github.com/ruby/spec/pull/1262 and included that fix in this change.

seven1m (approved):

PR #2729, opened 2025-05-12
Show verbose compile time warnings

herwinw:


herwinw:
For now I'm skipping this compile time warning in the generated bytecode. It doesn't really fit the regular structure, so we would probably need something more like this:
```
define_block
push_string "whatever Prism gave us"
end define_block
push_argc 0
send :warn with block once
```
This feels far out of scope

herwinw:
On second thought: the bytecode VM should just print those when compiling to bytecode, not include it in the bytecode

seven1m (approved):
This is certainly fine.

I do wonder (in agreement with your last comment), if we should fully embrace the AOT nature and show these at compile time. Maybe we could alter the specs to change `eval` to `ruby_exe` with a note about Natalie being special.

(I'm also ok with kicking that can down the road and deciding later.)

PR #2730, opened 2025-05-12
Fix case with no target expression with splatted values

herwinw:


herwinw:
I guess the CI is broken

seven1m (approved):

Issue #2735, opened 2025-05-17
REPL incomplete code handling


Stevo-S:
The REPL currently doesn't seem to allow for code to be input over multiple lines.
It evaluates every line as soon as "Enter" is hit.

```sh
❯ bin/natalie
nat> def some_method
syntax error: expected a delimiter to close the parameters
unexpected end-of-input, assuming it is closing the parent top level context
expected an `end` to close the `def` statement
```

I spend a lot of time testing out stuff on REPLs and this would be really useful to have.

If there isn't already some way to achieve this (like with a key-binding I don't know about for example),
I could send a PR with a very rudimentary implementation of it.
Enough to allow for something like:

```sh
❯ bin/natalie
nat> def some_method
nat* puts 'some_method called!'
nat* end
:some_method
nat> some_method
some_method called!
nil
nat>
```

seven1m:
Yeah, that would be nice! If you wanna take a stab at it, please do.

herwinw:
It looks like Prism has a few special exception types for these kind of cases. I tested it with the following two lines (separately):
```ruby
def foo(abc)
def foo(abc) = a +
```
This gave me the types `:unexpected_token_close_context` and `:expect_expression_after_operator`, so I guess we should be able to catch these as a "need more input"

Stevo-S:
Thanks @seven1m and @herwinw.

Will definitely take a stab at it.

While we're talking about the REPL, would you be open to switching from `linenoise-rb` to `reline` down the line?
`reline` is pure Ruby which I think would make it easier to give the REPL more quality of life features.

On the other hand, `linenoise` prides itself on being "zero-config" so I can see preferring it over `reline` which aims to be compatible
with GNU Readline.

PR #2736, opened 2025-05-17
allow repl to accept more lines if code incomplete

Stevo-S:
Catch "unexpected end-of-input" parser errors and await more input.

These errors are likely due to incomplete code such as but not limited to:
- open brackets that are yet to be closed
- module, class, method, block, loop definitions that don't yet have a corresponding `end`
- operators where an operand is specified on a subsequent line

When these errors are caught, we change the prompt from 'nat> ' to 'nat* '
to signal to the user that further input is need before we can evaluate their code.

We keep track of all the lines input by the user until the parser no longer returns an
"unexpected end-of-input" error.
At that point we either proceed with the compilation or show the user the error if there is one.

### DIFF
--- a/lib/natalie/repl.rb
+++ b/lib/natalie/repl.rb
@@ -136,8 +136,16 @@ GC.disable
 
 LibNat.init
 
+prompt = 'nat> '
+lines = []
+
+reset_lines = lambda do
+  lines = []
+  prompt = 'nat> '
+end
+
 loop do
-  line = Linenoise.readline('nat> ')
+  line = Linenoise.readline(prompt)
   break if line.nil?
 
   next if line.strip.empty?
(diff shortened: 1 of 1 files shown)


Stevo-S:
Fix for #2735

Stevo-S:
This is pretty basic in my opinion but I feel like it's a good enough start.

It lacks things like automatic indentation for subsequent lines and multiline editing (you can still only edit the current line).
Syntax highlighting may need a little work for some constructs (e.g. string literals) when they span multiple lines.

But we should be able to take care of this over time, I feel.

I've tested it against a bunch of inputs and it seems to work as expected for anything I can throw at it.

Even something as contrived as the following

```ruby
nat> module M
nat*   class C
nat*     def meth
nat*       puts "
nat*             1 + 2 = #{
nat*                       1 +
nat*                           2
nat*                       }
nat*             "
nat*       end
nat*     end
nat*   end
:meth
nat> M::C.new.meth

1 + 2 = 3
nil
nat>
```

It still handles other syntax errors like we would want it to:

```ruby
nat> class;
nat* C
nat* end
syntax error: unexpected constant path after `class`; class/module name must be CONSTANT
unexpected constant path after `class`; class/module name must be CONSTANT
unexpected constant path after `class`; class/module name must be CONSTANT
```
```ruby
nat> 1 +
nat* = 2
syntax error: unexpected '='; expected an expression after the operator
unexpected write target
```

Open to suggestions of any more tricky expressions that we can test with.

seven1m:
Very nice, good job.

PR #2741, opened 2025-05-21
repl automatically includes pp for pretty-printing Ruby objects

Stevo-S:


Stevo-S:
This is weird.
I'm seeing odd test failures.
First on linux (clang, group3) and then (after pushing an empty commit to rerun the tests) on linux (gcc, group1).

Not sure why this is happening but I'd find it odd if things got flaky just from adding `require 'pp'` to `repl.rb`.
🤔

seven1m:
That particular failure is just a flaky test. Something to do with sockets.

Stevo-S:
I see. Will ignore then.
Thanks!

PR #2748, opened 2025-05-27
Additional unit test for Data#deconstruct_keys

herwinw:
This tests the change made in #2747 to ensure we hit this path in the tests.


herwinw:
https://github.com/ruby/spec/pull/1264

Created better specs for upstream, going to close this item

PR #2750, opened 2025-05-27
repl handles multi-line strings

Stevo-S:


seven1m:
This is cool, that formatting task is failing because it cannot commit to your branch. Would you be willing to run `rake format` on this and amend the commit?

PR #2755, opened 2025-06-03
Use connect_timeout argument in TCPSocket connect

herwinw:


herwinw:
So, this now adds two passing specs, and makes one fail on MacOS, so it's still progress (and we can now start with the additional timeout specs)
I want to leave the MacOS issue as is for now, I don't have the hardware to debug that.

PR #2758, opened 2025-06-12
fix tcp sock initialize econnrefused scenario on macos

Stevo-S:
Fix `TCPSocket#initialize` on MacOS when no server is listening
on the given address.

`poll(2)` can indicate that a socket is writable following
`connect(2)` returning an `EINPROGRESS` result.

We need to check with `getsockopt(2)` again to be sure the socket
is truly writable.

### DIFF
--- a/lib/socket.cpp
+++ b/lib/socket.cpp
@@ -1129,6 +1129,16 @@ Value Socket_connect(Env *env, Value self, Args &&args, Block *block) {
                 auto error = Object::_new(env, TimeoutError, { StringObject::create("Connect timed out!") }, nullptr).as_exception();
                 env->raise_exception(error);
             } else {
+                // poll() can return a positive value even if the socket is not writable.
+                // We need to check the socket options to see if it really is writable.
+                int option_value = 0;
+                socklen_t option_len = sizeof(option_value);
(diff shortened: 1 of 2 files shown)


Stevo-S:
From the linux `connect(2)` [man page](https://man7.org/linux/man-pages/man2/connect.2.html):
> EINPROGRESS
              The socket is nonblocking and the connection cannot be
              completed immediately.  (UNIX domain sockets failed with
              EAGAIN instead.)  It is possible to [select(2)](https://man7.org/linux/man-pages/man2/select.2.html) or [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html)
              for completion by selecting the socket for writing.  After
              [select(2)](https://man7.org/linux/man-pages/man2/select.2.html) indicates writability, use [getsockopt(2)](https://man7.org/linux/man-pages/man2/getsockopt.2.html) to read
              the SO_ERROR option at level SOL_SOCKET to determine
              whether connect() completed successfully (SO_ERROR is zero)
              or unsuccessfully (SO_ERROR is one of the usual error codes
              listed here, explaining the reason for the failure).

Looks like the advice on using `getsockopt(2)` after `poll(2)` indicates success applies to MacOS as well.

PR #2761, opened 2025-06-16
Fix Addrinfo#initialize pfamily setting for single argument

herwinw:
It looks like our old logic is inside out: we copied the afamily value, but instead the callers should provide the pfamily value explicitly.
Draft request, this adds a lot of new offenses that should be fixed.

Short test with MRI for `Addrinfo.getaddrinfo` indicates that it does not explicitly call `Addrinfo#initialize`, so we might want to rewrite some logic.


seven1m (approved):
🎉

PR #2767, opened 2025-06-28
fix wrong spawn error on macos

Stevo-S:
Fix an issue where the wrong error was being raise on MacOS
when Process.spawn was called on a file without execute permissions
or a directory instead of a file.


seven1m (approved):

Stevo-S:
> Maybe (as a future work) we could add an optional argument to raise_errno,
> I kind of dislike the assignment to errno

Agreed. Definitely feels a little weird to have to do it.

Stevo-S:
I guess this is good to merge @seven1m?

PR #2771, opened 2025-07-03
Remove cast from EncodingObject::encode

herwinw:
This was casting the const property away. Instead, just use the numeric value and preserve the constness.


herwinw:
Tyring to get it to compile with `-Wold-style-cast`, but we have a long way to go.

PR #2772, opened 2025-07-03
Use perfect forwarding for StringObject::create

herwinw:
This removes all the specialized implementations of `StringObject::create` with a generic one that just forwards the arguments to the actual private constructors. We can do the same thing with all other classes that have these create factory methods.

Small caveat: perfect forwarding is not perfect. The callers that use a stack allocated array don't work, these have been converted into calls using `TM::String::create_and_take_ownership` which removes a memcpy too.

### DIFF
--- a/include/natalie/string_object.hpp
+++ b/include/natalie/string_object.hpp
@@ -39,89 +39,10 @@ public:
         Invalid,
     };
 
-    static StringObject *create(ClassObject *klass) {
+    template <typename... Args>
+    static StringObject *create(Args &&...args) {
         std::lock_guard<std::recursive_mutex> lock(g_gc_recursive_mutex);
-        return new StringObject(klass);
-    }
-
-    static StringObject *create() {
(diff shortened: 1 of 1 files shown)


seven1m (commented):
Well ain't that fancy!

Does the template programming slow down compilation at all? I do worry that using template programming seems to be slower (at compile time), at least with gcc.

rolandpj1968:
> Well ain't that fancy!
> 
> Does the template programming slow down compilation at all? I do worry that using template programming seems to be slower (at compile time), at least with gcc.

This is an interesting question when running at -O3. Legend has it that -O2 is a better balance of compilation time vs perf.

But no inlining? Dunno what clang and gcc actually do nowadays.

I'm actually surprised that member fields can be forward-referenced in header files. My C++ foo is a bit rusty.

In any case we could definitely factor some of the instruction->c++ generation in order to cut parsing at least.

seven1m (approved):

PR #2775, opened 2025-07-08
Generic solution for append+snprintf in TM::String

herwinw:
Just use the m_str buffer in place. This reduces some code duplication, and the new code uses the existing buffer in place, which removes a copy operation as well.

The following script:
```c++
TM::String str {};
str.append(static_cast<int>(1));
str.append(static_cast<size_t>(2));
str.append(static_cast<ssize_t>(3));
str.append(static_cast<double>(4.5));
str.print();
```
roughly reduced 4K operations in callgrind.

Meanwhile, compile down went down: the above script compiled 100 times in a row with G++ (14.2.0) and -O3 takes roughly 14.6 seconds on my machine when using master, and 13.5 seconds with this patch.

### DIFF
--- a/ext/tm/include/tm/string.hpp
+++ b/ext/tm/include/tm/string.hpp
@@ -932,10 +932,7 @@ public:
      * ```
      */
     void append(const size_t i) {
-        const int length = snprintf(NULL, 0, "%zu", i);
-        char buf[length + 1];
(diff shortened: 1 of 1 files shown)


seven1m (approved):
Cool! 😎

PR #2781, opened 2025-07-10
Precompute the length of a string when calling the _s operator

TheGrizzlyDev:
so that there isn't any need to later compute the string's length.


TheGrizzlyDev:
> Looks great, just need the formatting fixed. Thanks!

Thanks! The failure in the formatting job does not look related to formatting though. Any clues what's up with it? Or else I can find some time to debug it later

TheGrizzlyDev:
thanks @seven1m , the autoformatter really wasn't having it ahhahah

PR #2798, opened 2025-07-25
Fix memory leak in socket.cpp

seven1m:


lib/socket.cpp
```diff
@@ -533,9 +533,11 @@ Value BasicSocket_getsockopt(Env *env, Value self, Args &&args, Block *block) {
 
     socklen_t len = 1024;
     auto buf = new char[len];
+    Defer delete_buf([&buf]() { delete[] buf; });
     auto result = getsockopt(self.as_io()->fileno(), level, optname, buf, &len);
     if (result == -1)
         env->raise_errno();
+    delete_buf.disable();
     String str { std::move(buf), len };
```
herwinw, line 541: An alternative solution could be:
```c++
auto result = getsockopt(self.as_io()->fileno(), level, optname, buf, &len);
String str { std::move(buf), len };
if (result == -1)
    env->raise_errno();
```
That way we transfer the ownership before the error statement and we don't need the temporary Defer-object that must be disabled again.

The other ones depend on a length from a read operation, so those would be something like:
```c++
const auto length = read(...)
String str {std::move(buf), bufsize };
if (length < 0) env->raise(....)
str.resize(length);
```

But I feel it would be better to have a read ownership object around that `buf` instead of just using a plain `new[]`

seven1m: I got a read buffer overflow from ASan trying this. I will try again later.

PR #2799, opened 2025-07-26
Add TM::MemBuffer class

herwinw:
This is an early draft for a generic solution for the issues shown in #2798. I still need to add some member functions for things like accessing and resizing.

The central mechanic is probably this cryptic looking method:
```c++
operator T *() &&;
```
The trailing `&&` makes this work only if the object is used as an rvalue. This should make it possible to create a String like this:
```c++
TM::MemBuffer<char> buf { 5 };
// read into the membuffer, at this moment it's still missing an accessor to the memory
TM::String str { std::move(buf), 5 };
```

### DIFF
--- a/ext/tm/include/tm/mem_buffer.hpp
+++ b/ext/tm/include/tm/mem_buffer.hpp
@@ -1,0 +1,99 @@
+#pragma once
+
+#include <cstddef>
+
+namespace TM {
+
+template <class T>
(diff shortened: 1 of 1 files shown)


herwinw:
Nope, that interface will not work in this way, this will use the constructor with `(char*, size_t)` and still make a copy.

PR #2804, opened 2025-07-27
Don't execute block in TCPServer#initialize

herwinw:
Even though it did print the generic "does not take a block. use open instead" error, the block was still executed. Even worse, any exception raised in the block is caught and ignored, so the code in the spec with `{ raise }` did actually raise the exception in the block, but we never saw that.


herwinw:
Fixing bugs by throwing code away, that's still my favourite way of fixing bugs :sweat_smile:

seven1m (approved):

PR #2805, opened 2025-07-28
Add IPv6 support to TCPSocket.gethostbyname

herwinw:
By removing the call to gethostbyname(3) and forwarding it to Addrinfo.getaddrinfo instead. This is what MRI does too, shown by the exception you get when calling this method with an unknown hostname.

The function gethostbyname(3) has been removed from POSIX.1-2008, the fact that this still worked is very much by accident.

### DIFF
--- a/lib/socket.cpp
+++ b/lib/socket.cpp
@@ -1743,37 +1743,6 @@ Value TCPSocket_initialize(Env *env, Value self, Args &&args, Block *block) {
     return self;
 }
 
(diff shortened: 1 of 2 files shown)


herwinw:
I just tried running the socket specs with the NATFIXME line in `spec/library/socket/fixtures/classes.rb` enabled again, this was the first one that generated issues. There are probably going to come more changes like this.

PR #2812, opened 2025-08-04
Import specs for Regexp#source

herwinw:


herwinw:
(This was just an import of a missing spec, with one NATFIXME marker added because that one fails (for reasons outside of the method under test), I merged this one without a review, since a review doesn't add anything in this case).

PR #2826, opened 2025-08-24
fix strftime with %z  on MacOS

Stevo-S:
Fix strftime on MacOS when the %z conversion specification is included in the format string.

On MacOS, the `tm_gmtoff` member of the `time` struct is ignored by `strftime(3)` which uses the curent set timezone of the machine regardless of the value.

To fix this issue, on MacOS, the `build_string` member function is updated to convert the  value of the `tm_gmtoff` member of the `time` to a timezone representation and update the `TZ` environment variable accordingly.

The original `TZ` value is restored once `strftime(3)` has been called.

A lock is employed to make the process of reading/updating the environment variable thread safe.

### DIFF
--- a/spec/shared/time/strftime_for_time.rb
+++ b/spec/shared/time/strftime_for_time.rb
@@ -121,27 +121,19 @@ describe :strftime_time, shared: true do
     end
 
     it "formats a local time with positive UTC offset as '+HHMM'" do
-      NATFIXME 'fails on MacOS', condition: RUBY_PLATFORM.include?('darwin'), exception: SpecFailedException do
-        @new_time_in_zone["CET", 1, 2005].strftime("%z").should == "+0100"
-      end
+      @new_time_in_zone["CET", 1, 2005].strftime("%z").should == "+0100"
     end
 
     it "formats a local time with negative UTC offset as '-HHMM'" do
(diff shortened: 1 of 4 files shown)


spec/shared/time/strftime_for_time.rb
```diff
@@ -121,27 +121,19 @@
...
-        @new_time_in_zone["PST", -8, 2005].strftime("%z").should == "-0800"
-      end
+      @new_time_in_zone["PST", -8, 2005].strftime("%z").should == "-0800"
     end
 
     it "formats a time with fixed positive offset as '+HHMM'" do
-      NATFIXME 'fails on MacOS', condition: RUBY_PLATFORM.include?('darwin'), exception: SpecFailedException do
-        @new_time_with_offset[2012, 1, 1, 0, 0, 0, 3660].strftime("%z").should == "+0101"
-      end
+      @new_time_with_offset[2012, 1, 1, 0, 0, 0, 3660].strftime("%z").should == "+0101"
```
Stevo-S, line 132: I'm not 100% sure that this test was not already fixed by a previous change and just
needed to not be skipped.

Stevo-S: Brilliant!

I can rest assured that the change captures, exactly, all the fixed issues.

Thanks for the clarification.

PR #2834, opened 2025-11-29
Fix build+CI for MacOS

herwinw:


herwinw:
Let's see if this fixes the build issues on macos

herwinw:
It fixes the build, but now we're having issues with Bundler.

seven1m (approved):

PR #2949, opened 2026-04-11
String#split

kddnewton:
Fully implement String#split. This is a shockingly complex function, which has four distinct modes, depending on its parameters. I tried to keep the implementation pretty close to MRI to make it easier to review/change in the future. All String#split specs now pass.


seven1m (approved):
CI failure looks related, but code looks great 👍

PR #2952, opened 2026-04-11
String#unicode_normalize and String#unicode_normalized?

kddnewton:
Copied straight from MRI, since it's already implemented in Ruby.


seven1m (approved):
Nice!

seven1m:
We'll want to ignore that file for auto formatting

PR #2954, opened 2026-04-14
Make recvfrom block regardless of socket blocking

kddnewton:
**Let me say first that I am not totally confident about this code.**

I think this is correct, based on my cursory understanding of https://github.com/ruby/ruby/blob/ee90e8e380190468a87ab5df3597dafd4f5def53/ext/socket/init.c#L202-L219. But I am super not comfortable with this stuff in general so I'm a bit guessing. I keep hitting this Errno::EAGAIN error being raised from this one spec in CI, so I took a shot at fixing it.

Below is the actual commit message:

When recvfrom(2) returns EAGAIN, there is currently no retry mechanism, meaning if a child calls recvfrom and there is no content yet we just end up raising Errno::EAGAIN. This happens in CI a lot with recvfrom_spec.

Instead, this commit introduces a loop that retries when EAGAIN is set to select(2) until content is available, effectively making this blocking. Note it does this only if the first call to recvfrom(2) sets an error, so the fastpath is still the same.

### DIFF
--- a/lib/socket.cpp
+++ b/lib/socket.cpp
@@ -575,6 +575,27 @@ static ssize_t blocking_recv(Env *env, IoObject *io, char *buf, size_t len, int 
     return ::recv(io->fileno(), buf, len, flags);
 }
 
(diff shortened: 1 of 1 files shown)


seven1m (approved):
lgtm 🙈

(I don't know either)

PR #2955, opened 2026-04-14
Fix flaky back references spec

kddnewton:
**Not entirely confident about this one.** I think it's correct? I ran that spec a bunch of times locally on my mac and it failed a bunch, and then with the fix it passed. But not super sure.

Clear calling_env of the block to ThreadObject::initialize so $~ and other per-thread state are not leaked from the parent thread.


seven1m (approved):
Nice! 👏

PR #2956, opened 2026-04-15
Object#send(:initialize) should not return self

kddnewton:
send(:initialize) should return whatever the initialize method returned, not self necessarily. Class.new is the place where we need the override to explicitly ignore the return value of initialize.

There are some builtins that explicitly return self in their initialize (namely the enumerator chain object). That I updated in order to maintain the right behavior.

This has an added benefit of making every send just a little faster to have one fewer comparison.


seven1m (approved):
Ha, nice.

PR #2957, opened 2026-04-15
Encoding::Converter

kddnewton:
Okay, so this is a fairly sizable PR, sorry about that. I tried to keep it relatively reviewable (definitely go commit by commit). Feel no rush.

This PR introduces Encoding::Converter, the object that MRI uses internally to do all of its string transcoding. In particular, it's the object that is used to back String#encode and IO transcoding. Each commit in the stack introduces more functionality for it, then at the end we make the big switch moving stuff from String to instead create a temporary Encoding::Converter that does the transcoding.

I made a slight deviation from MRI, which I think is actually fine: MRI maintains a graph of transcoders and does a BFS to walk from one encoding to another (because they have transcoder shortcuts between various Japanese encodings and such). Natalie instead uses Unicode as a hub and everything else as spokes. This ends up working out well because the encoding objects already had the required infrastructure for it.

This work unblocks Natalie being able to handle a lot of the IO specs, it adds all of the Encoding::Converter specs, and generally gets us closer to parity with MRI in terms of all of the various complexities around encoding.

Again, sorry about the size of the PR.


seven1m (approved):
Awesome! 😎

PR #2960, opened 2026-04-16
Locale-agnostic ctype

kddnewton:
There are a bunch of places in MRI that uses these ctype helpers that are locale-agnostic because in some locals things like \xFF are printable. Here we centralize all of that into our own include/natalie/ctype.hpp to match that behavior.


seven1m (approved):
Clean.

PR #2966, opened 2026-04-19
Speed up JSON lexer

kddnewton:
Instead of using string slices (which will allocate), instead use the running index and only slice when necessary. This speeds up the lexer enough to make it reasonable to guard against the timing out CVE spec.


seven1m (approved):
Nice! 🏎️

PR #2967, opened 2026-04-20
IO#set_encoding work

kddnewton:
Explicitly use FMODE_* flags for IOs, matching MRI's enum (minus a few that we don't support yet). This makes it a little easier to handle things like readable/writeable/sync checks.

This also works on what happens when you call external_encoding, which goes through a whole conditional chain to see what to return.

### DIFF
--- a/include/natalie/io_object.hpp
+++ b/include/natalie/io_object.hpp
@@ -29,8 +29,8 @@ public:
         return new IoObject(type, klass);
     }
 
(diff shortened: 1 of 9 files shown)


kddnewton:
Marking as draft as I'm no longer sold on this approach. I think I'm going to mirror the FMODE_* flags from MRI instead, it seems like they might be more necessary than I originally thought.

seven1m (approved):

PR #2974, opened 2026-04-20
TCPServer#gets read before select

kddnewton:
read before select in gets to ensure under O_NONBLOCK we do not hang forever in select(2) waiting for content.


seven1m (approved):
cool

PR #2983, opened 2026-04-28
Thread.handle_interrupt and async exceptions

kddnewton:
Adds Thread.handle_interrupt / Thread.pending_interrupt? and replaces m_exception with a queue of pending exceptions to raise.

Lots of comments in this commit because it is somewhat tricky to describe/implement. Basically a cross-thread exception can be deferred with thread APIs to raise either when the next blocking thing would happen (usually an IO or sleep) or they can be hidden entirely.

### DIFF
--- a/include/natalie/thread_object.hpp
+++ b/include/natalie/thread_object.hpp
@@ -93,8 +93,11 @@ public:
     Value status(Env *env);
     String status();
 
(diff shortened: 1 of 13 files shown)


seven1m (approved):
Wow! TIL

PR #2984, opened 2026-05-01
Syslog, IO#reopen, output_to_fd fixes

kddnewton:


src/io_object.cpp
```diff
@@ -1367,6 +1367,89 @@ Value IoObject::readline(Env *env, Optional<Value> sep, Optional<Value> limit, O
...
+                if (::dup2(other_fileno, m_fileno) < 0)
+                    env->raise_errno();
+                if (m_fileno > STDERR_FILENO)
+                    set_close_on_exec(env, Value::True());
+            }
+
+            // MRI seeks both IOs to other's logical position to keep reads continuing.
+            // natalie tracks read state in a per-instance read buffer instead, so
+            // transfer it across (covers e.g. "reads from current position" spec).
+            m_read_buffer = other_io->m_read_buffer;
```
seven1m, line 1446: We don't have to fix this now, but may want to copy MRI's behavior here sooner than later because we could get some subtle and hard-to-find bugs.

(Full disclosure: I haven't looked at the IO and read buffer situation in a long time, so this was the result of me talking with ChatGPT to try to understand the difference.)

```ruby
# reopen_buffer_repro.rb

path = File.expand_path('reopen_buffer_repro.txt', __dir__)

File.open(path, 'w') do |file|
  file.write("abcdef\n")
end

other = File.open(path, 'r')
io = File.open(path, 'r')

begin
  puts "engine=#{RUBY_ENGINE}"
  puts "other.getc=#{other.getc.inspect}"

  io.reopen(other)

  puts "io.getc=#{io.getc.inspect}"
  puts "other.getc=#{other.getc.inspect}"
  puts "io.getc=#{io.getc.inspect}"
  puts "other.getc=#{other.getc.inspect}"
ensure
  io.close unless io.closed?
  other.close unless other.closed?
  File.delete(path) if File.exist?(path)
end
```

```sh
→ ruby reopen_buffer_repro.rb
engine=ruby
other.getc="a"
io.getc="b"
other.getc=nil
io.getc="c"
other.getc=nil

→ bin/natalie reopen_buffer_repro.rb
engine=natalie
other.getc="a"
io.getc="b"
other.getc="b"
io.getc="c"
other.getc="c"
```

seven1m (approved):
Moves the needle forward, but we should put a pin in that buffer copy issue to fix later.

src/io_object.cpp
kddnewton: Oof I didn't understand it was that different. Woof. I'll take a look at it after I merge this.

PR #2985, opened 2026-05-01
IO#reopen read-buffer semantics

kddnewton:
Seek the shared fd to the source's user-visible position and clear both per-IO read buffers, instead of copying the source buffer to self (which caused both IOs to read the same data after reopen).

Largely AI assisted on this one, but I _think_ it makes sense?


seven1m (approved):
Cool

PR #2989, opened 2026-05-04
Run ensure when return exits a begin/ensure

kddnewton:
This one is a bit complex. Basically TryInstructions gains a `has_ensure` field in order to determine how it should generate code in the catch block. If it's true, it will re-raise exceptions after the body has been run.


seven1m (approved):
Dang, that fixes a nice big bug. Well done.

PR #2990, opened 2026-05-04
Vendor net/http

kddnewton:
A direct port of net/http and its related specs.


seven1m (approved):
Awesome! We should discard that formatting commit and ignore these files for formatting before we merge.

PR #2997, opened 2026-05-08
IClass

kddnewton:
This is a bit of a bigger change, and is a matter of taste, so we can definitely _not_ do this if preferred.

Basically, the way MRI does module inclusion is that it inserts something called an iclass (I'm assuming inserted class?) into the ancestor chain above the current singleton. prepend puts it below. Method lookup/ancestry lookup is then a straight walk through the chain. Natalie has a separate list of included modules, which means method lookup/ancestry lookup needs to look in two places. Not a bad thing necessarily, but it does make the logic (IMO) a little more complex.

This PR migrates over to the iclass approach that MRI uses. As a bonus, it also implements `singleton_methods`, which was actually the impetus. That's now a straight walk through an ancestor chain.

### DIFF
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -20,6 +20,7 @@
 #include "natalie/binding_object.hpp"
 #include "natalie/block.hpp"
 #include "natalie/class_object.hpp"
+#include "natalie/iclass_object.hpp"
 #include "natalie/complex_object.hpp"
 #include "natalie/constant.hpp"
 #include "natalie/dir_object.hpp"
--- a/include/natalie.hpp
+++ b/include/natalie.hpp
@@ -20,10 +20,10 @@
 #include "natalie/binding_object.hpp"
 #include "natalie/block.hpp"
 #include "natalie/class_object.hpp"
-#include "natalie/iclass_object.hpp"
 #include "natalie/complex_object.hpp"
 #include "natalie/constant.hpp"
 #include "natalie/dir_object.hpp"
+#include "natalie/iclass_object.hpp"
 
 #include "natalie/encoding/ascii_8bit_encoding_object.hpp"
 #include "natalie/encoding/big5_encoding_object.hpp"
--- a/include/natalie/class_object.hpp
+++ b/include/natalie/class_object.hpp
@@ -63,6 +63,13 @@ public:
     bool is_singleton() const { return m_is_singleton; }
     void set_is_singleton(bool is_singleton) { m_is_singleton = is_singleton; }
 
+    // Set on the first prepend into this class. Points at the origin iclass that
+    // holds the class's own methods so prepended modules' `super` finds them.
+    ClassObject *origin() const { return m_origin; }
+    void set_origin(ClassObject *origin) { m_origin = origin; }
+
+    virtual void visit_children(Visitor &) const override;
+
     virtual String backtrace_name() const override final;
 
     virtual TM::String dbg_inspect(int indent = 0) const override {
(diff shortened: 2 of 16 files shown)


seven1m (commented):
This seems to make things conceptually much more complicated than our current approach. Do you think using the IClass approach is critical for gaining the needed MRI compatibility or could we stick with our current approach and get the same results?

kddnewton:
Yeah it's a good, question honestly I'm not sure! Let me do a write up of the tradeoffs so we can properly evaluate this. On vacation for this week, but I will be back at it after that. I will let you know!

kddnewton:
Okay so I'm back from vacation and finally getting back to this. Here's what I think about it.

The picture before was that each ModuleObject carried two fields: `m_superclass` and `m_included_modules`. This was simple enough on the module object side of things, but somewhat complex for the consumers for that. What I mean is, in order to get the full ancestor list you had to recurse through both sides (walk one level of m_included_modules, recurse into each entry's own m_included_modules, eventually advance to m_superclass, repeat).

The picture after is that each ModuleObject carries a single `m_superclass` field, which is the next node in a single linear chain. This chain includes both classes and modules, with modules represented as iclass objects. The `m_origin` field on the ModuleObject serves as a marker to indicate if it was prepended into another module (i.e., if it's a hollow wrapper). The `m_iclasses` field is used for backlinks to support modifying a module after it has already been included into something.

More concretely, here are a couple of consumers and how they are impacted:

**ancestors**

Before:

```
do {
  push self // if m_included_modules empty
  for m in m_included_modules: push m
  klass = klass.m_superclass
} while (klass)
```

A bug here because it didn't recurse into m_included_modules (in the old model it needs to do both containers). After:

```
for (p = this; p; p = p.m_superclass) {
  if (p.is_origin_hollow()) continue
  push p.defined_class() // unwrap iclass
}
```

This is correct because in the new model it's a single linear list.

**find_method**

Before:

```
for m in m_included_modules {
  if m == this {
    search m.m_methods // reads own table
  } else {
    recurse m.find_method(...) // walks m's flat list AND m's m_superclass
  }
  recurse m_superclass.find_method(...) // walks parent's flat list AND parent's super
}
```

This is correct, but it means that we end up visiting some modules twice (because we go up the chain with find_method and then again with m_superclass). In practice this was masked by caching, but it was redundant. After:

```
for (p = this; p; p = p.m_superclass) {
  if (p.is_origin_hollow()) continue
  if found in p.methods_table(): return ...
}
```

**is_subclass_of**, **included_modules**, **does_include_module**, **constants**, **cvar_get_maybe**, **methods**, **find_constant_in_modules**:

Each one makes a choice of walk one level or recurse both axes. Some of them recursed, some didn't. Some had bugs, others didn't but could end up visiting stuff twice. For example: `find_method` recursed but `ancestors` didn't, so `obj.respond_to?(:foo)` could return true when foo was in a transitively-included module, but `obj.class.ancestors.include?(M)` returned false for the same M that has `foo` on it. The two halves of the API disagreed about what was in the chain.

**singleton_methods**

For implementing the actual feature this is meant to, I could have built another walk through the double container, and probably mitigated any issue with caching, but I think the iclass model is just cleaner.

All that being said, I think all of these things could be mitigated/fixed in the old model with enough care. So I am happy to go with that instead if you'd prefer.

seven1m (approved):
Thank you for that thorough explanation. I went from "this feels more complicated and our old model was fine" to understanding our old model wasn't really "fine" -- I didn't realize how broken it was. Very good, thank you for working on this and taking the time to explain it to me! +1

PR #2999, opened 2026-06-23
Match upstream mspec matchers

kddnewton:
.should should return a proxy if no arguments are given to it. That proxy can then be used to make assertions.

When this is implemented properly, it means a bunch more assertions are now firing, which means we need to NATFIXME a bunch of stuff because otherwise it just silently fails.


seven1m:
Oh, I see: because `Matcher` was a regular Object, the `method_missing` wouldn't fire for stuff like `eql?`. Whoops!

seven1m (approved):

kddnewton:
Yeah, so something like `frozen?` would call the regular predicate method and then not actually assert anything.

PR #3000, opened 2026-08-06
OrderedHashmap

kddnewton:
So there are quite a few places in Ruby where insertion order for hashes is depended on, unfortunately. This includes Module#methods, Module#constants, Module#class_variables, etc. So we need an insertion order container. One exists in HashObject, but that's Ruby level and it got complicated when I tried to use it.

This introduces a new TM primitive: OrderedHashmap. It's loosely based on Hashmap and also loosely based on the CRuby hashmap implementation. It keeps the keys in a contiguous table so iteration just walks the list.

We _might_ want to consolidate these into one thing, depending on benchmarks/memory usage. It looks good to me from the benchmarks I ran, but I figured it would be better to get this merged as its own thing first. Plus then we can just replace the hashes that back user-facing functionality that require insertion order.

### DIFF
--- a/ext/tm/include/tm/ordered_hashmap.hpp
+++ b/ext/tm/include/tm/ordered_hashmap.hpp
@@ -1,0 +1,594 @@
+#pragma once
+
+#include "tm/hashmap.hpp"
+
+namespace TM {
+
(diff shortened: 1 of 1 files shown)


seven1m (approved):
Cool

